PR #3, opened 2025-10-23
Documenting the code via doxygen

wmiler:
This is just the config file for running _doxygen_ against the tree. 
Stores the generated html in ./docs/html/

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1158,7 +1158,7 @@ FILTER_SOURCE_PATTERNS =
 # (index.html). This can be useful if you have a project on for instance GitHub
 # and want to reuse the introduction page also for the doxygen output.
 
-USE_MDFILE_AS_MAINPAGE =
+USE_MDFILE_AS_MAINPAGE = README.md
 
 # The Fortran standard specifies that for fixed formatted Fortran code all
 # characters from position 72 are to be considered as comment. A common


aknrdureegaesr:
I mildly **object** to this being merged: I expect the folder where the doxygen stuff is generated to be added to `.gitignore`.

Not a second objection, but I would very much like to see a short how-to added to the project. I suggest starting a new file `CONTRIBUTING` or, possibly better yet, `CONTRIBUTING.md` that has a couple of lines on this or three. Lest every developer who's never used doxygen needs to go to the web side, one developer at a time.

Chris-AC9KH:
> (Yes, I did notice the breach of process and understand my objection comes to late to be addressed before the merge. I don't know what to think about it. If you people do not discuss my proposal and also ignore it, why the heck did I bother writing it up?)

This was a simple configuration file that doesn't change the functionality of the code, and that I didn't see any reason to leave it as a open PR. Modifications to it can be easily made with subsequent PR.

Chris-AC9KH:
As far as that goes, there's other things yet that should probably be added to .gitignore. Xcode can put a new artifact I discovered into the source tree (depending on how Xcode is set up) for app entitlements.

Chris-AC9KH:
> I suggest starting a new file `CONTRIBUTING` or, possibly better yet, `CONTRIBUTING.md` that has a couple of lines on this or three. Lest every developer who's never used doxygen needs to go to the web side, one developer at a time.

@aknrdureegaesr there is a contributor's guide in the docs (at least in JS8Call-improved):

https://github.com/Chris-AC9KH/js8call/blob/dev/2.4/docs/CONTRIBUTOR_GUIDE.md

If it needs changing for here (it likely does), then feel free to make any changes to it. It was included in the documentation update I did on JS8Call official, but never merged (yet).

I see zero reason to leave simple PR's that don't change the functionality of the codebase, but they add necessary development features. These can just be merged without going thru any formal review process. They are easy to reverse, or change, with subsequent PR's without affecting the build-ability or functionality of the code. It helps to keep the PR list cleaned up, and if something was missed just fix it with another PR and move on. I've got one coming up with a new tools directory with linux build scripts that is the same thing. I wouldn't expect wasting everybody's time going thru a formal review process for such a thing. If two developers agree on it, merge it.

PR #5, opened 2025-10-24
Much improved heartbeat and CQ loop implementation, various minor cleanups.

aknrdureegaesr:
Thus far, the CQ loop and the HB loop code would trigger the PTT up to 1 second too late. So the transmitted signals might loose up to one second of air time. Half that on average.

- In SLOW, one second of air time less results in 0.5 s of symbol time less, nominally 1.6 symbols.
- In NORMAL, typically used for CQ and HB on a loop, nominally 3.1 symbols.
- In FAST, nominally 8 symbols (this takes into account that FAST has only 200 ms of initial quiet air time).

Also, I have seen it with injected logging that whatever TX delay I might have configured was ignored by the previous code when running HB or CQ loops. This bug has also been fixed.

This PR also contains a good measure of additional comments, name improvements, and some signal and slot cleanup.

I have tested this only somewhat, hope to do more of that in the upcoming days. I hope I did well, but there may be subtle bugs in here somewhere.

I'm not sure what our release plans are.

One option would be to release 2.4.0 from commit 02070ccfc3 (which is where that git tag currently sits and what has been tested extensively by @Chris-AC9KH  and possibly other people). Then this work could become part of a different release 2.4.1. This would offer a fallback route from 2.4.1 to 2.4.0, should these changes turn out to be buggy.

### DIFF
--- a/Modulator.cpp
+++ b/Modulator.cpp
@@ -31,12 +31,12 @@ Modulator::start(double        const frequency,
 {
   Q_ASSERT (stream);
 
-  if (m_state != State::Idle) {
+  if (isIdle()) {
       qCDebug(modulator_js8)
           << "Modulator does not find itself in state idle, but"
-          << (m_state == State::Active ? "Active" :
-              m_state == State::Synchronizing ? "Synchronizing" :
-              m_state == State::Idle ? "Idle" : "??What??")
+          << (m_state.load() == State::Active ? "Active" :
+              m_state.load() == State::Synchronizing ? "Synchronizing" :
+              m_state.load() == State::Idle ? "Idle" : "??What??")
           << "so calling stop()";
       stop();
   }
@@ -95,7 +95,7 @@ Modulator::start(double        const frequency,
 
   initialize(QIODevice::ReadOnly, channel);
 
-  m_state = 0 < m_silentFrames ? State::Synchronizing : State::Active;
+  m_state.store(0 < m_silentFrames ? State::Synchronizing : State::Active);
   m_stream = stream;
 
   if (m_stream)
@@ -131,7 +131,7 @@ Modulator::close()
     else              m_stream->stop();
   }
 
-  m_state = State::Idle;
+  m_state.store(State::Idle);
   AudioDevice::close();
 }
 
@@ -149,7 +149,7 @@ Modulator::readData(char * const data,
   qint16       *       samples         = reinterpret_cast<qint16 *>(data);
   qint16 const * const samplesEnd      = samples + maxFrames * (bytesPerFrame() / sizeof(qint16));
 
-  switch (m_state)
+  switch (m_state.load())
   {
     case State::Synchronizing:
     {
@@ -167,7 +167,7 @@ Modulator::readData(char * const data,
 
         if (!m_silentFrames)
         {
-          m_state = State::Active;
+          m_state.store(State::Active);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


Chris-AC9KH:
@aknrdureegaesr how do you want to handle this? There is already a 2.4 tag at that commit, it just hasn't been branched (yet). I intended to create the 2.4 branch from that commit this evening yet to store a "snapshot" in history where the codebase has been thoroughly thrashed and is as bug-free as we can make it.

So do you want this to be a WIP and label it as such? Or would you rather merge it to master as-is and fix any bugs later? I'd like to have @bazineta take a look at this. He did the port to C++ and is most familiar with this part of the code. I'm not thoroughly convinced your fix achieves the desired result since in my testing I haven't been able to duplicate any timing issues, or noticed any failed decodes on the receiving end. An HB or CQ is always transmitted in one frame here no matter what speed mode is used, and replies always come on the very next frame.

bazineta:
To add more context to observation (3), for small integral-sized values, a pattern like the following is simple and efficient; an excerpt from some of my own stuff (the `time_point` used here is just what my stuff happens to use; any integral value is in general fine as a std::atomic<> now):

Class member, private:

`std::atomic<std::chrono::system_clock::time_point> _last;`

Accessor and mutator, public:
```
[[nodiscard]] auto
last() const noexcept
{
            // Readers only need the timestamp value for freshness
            // checks/persistence. Writer uses release-store; readers
            // use relaxed-load to avoid fences.
            
            return _last.load(std::memory_order_relaxed);
}
void
setLast(std::chrono::system_clock::time_point const last) noexcept
{
            // Publish the updated timestamp for other threads promptly;
            // no pairing acquire is required by current readers.
            
            _last.store(ltd, std::memory_order_release);
}
```
This has the desired effect of the protective singleton without all the ceremony; it might be a good pattern here. While I've excerpted from what is a class in my implementation, this works equally well at namespace scope when the mission is just to have a single thread-safe global value.

aknrdureegaesr:
This does not yet even work. Can no longer type in messages. Need to fix that. So I decidedly **object** to this being merged as it presently sits.

aknrdureegaesr:
My own objection has been addressed:

> This does not yet even work.

For all I know, it does now. I'll try a few QSOs tomorrow.

bazineta:
> Later edit: `volatile` is not what I thought it was. I need to change the code to make it thread-safe.

Yes....the meaning of the term has changed since 2011, and now, it means 'data race here' ;).

Using `std::atomic<>` is probably your short path given the use case.

aknrdureegaesr:
So, thanks to @bazineta for the review comments! I learned a thing or two in the process!

In the end, I decided to use plain `store` and `load` with the standard arguments. I considered paying the price to document why we would accept the (low) risk of the modulator writing a new state and the main window reading an older one, and/or having subsequent readers of the code wonder about it. My judgement is that this price is not worth paying for a (wild guess) few hundred nanoseconds runtime improvement every 100 ms or so.

Chris-AC9KH:
It's been looked at plenty. It builds, seems to run without issues (on MacOS 26). I'd say if @bazineta has no real issues with it, go ahead and merge it. While it won't create a conflict anyway, I'd like to bump master to 2.4.1-dev as soon as this gets merged.

Chris-AC9KH:
Merged. Thanks!

PR #6, opened 2025-10-25
Updates to .gitignore

wmiler:
Left out of PR #3


Chris-AC9KH:
@wmiler since you are adding to .gitignore, would you mind making one more change to it that will save me from fixing it?

In the Xcode patch please change the upper and lower case JS8Call/js8call.entitlements to a wildcard `*.entitlements` I found Xcode can add entitlements files in the source tree for the other targets in a JS8Call build, depending on how Xcode is set up to do the build.

Issue #1, opened 2025-10-21
Things That Need To Be Done


Chris-AC9KH:
@bazineta @aknrdureegaesr @rruchte @wmiler @Joe-K0OG @orioncrocker @jfrancis42 I think I got licensing, copyrights, yadda yadda covered.

JS8Call-improved is NOT a f()rk of the JS8Call project. It was created from a bare branch with the current dev/2.4 code so it does not have a parent, while retaining all commit history, credits and copyrights. It was impossible to f()rk it because the current JS8Call code is too far behind.

The hope is that we can have a more up-to-date master branch which may, in effect, become upstream for the JS8Call project. Any work done here on development branches can be PR'd to the JS8Call project (when it catches up) by adding your personal JS8Call repo as a remote and push your JS8Call-improved development branch to your repo for PR on the official project.

For instance, I added my own JS8Call repo as downstream remote on this project:

`chris@Mac-mini-M1 src % git remote -v`
`downstream	https://github.com/Chris-AC9KH/js8call.git (fetch)`
`downstream	https://github.com/Chris-AC9KH/js8call.git (push)`
`origin	https://github.com/Chris-AC9KH/JS8Call-improved.git (fetch)`
`origin	https://github.com/Chris-AC9KH/JS8Call-improved.git (push)`

We can attempt to keep JS8Call up-to-date with PR's, but I don't know what else to do. We can't have developers submitting PR's on out-of-date code. It's just not right.

I'd like the policy to be, if it builds and passes review from a couple other developers, merge it into master so people aren't working with out-of-date code. Master is NOT a release branch, it's a development branch. And sometimes dev branches are buggy as a warm August night. But when the team decides it's time for a release, then branch it and concentrate on squashing bugs so the master development branch can continue to do what it does.

So there's a lot of stuff in the works; we need a post-build system for linux. We got new CMake configs, but that's in limbo right now so it's hard to proceed. We need a new versioning system (which was mentioned but not acknowledged). We need a way to get the latest code into the hands of more testers/end-users for debugging. I created this as a place where things can happen on a more timely basis, and still be able to feed the results to the JS8Call project. Maybe it will fly, maybe it won't. But I thought it was worth a try since it seemed to work for the wsjt-x-improved guys, which had the same problem and they've now got some really cool software.

Chris-AC9KH:
> I mean, our hope is that Jordan will eventually merge this, right?

[...]

PR #10, opened 2025-10-25
Mark issues and PRs as stale

wmiler:
Partially implements [@aknrdureegaesr comments](https://github.com/Chris-AC9KH/JS8Call-improved/issues/1#issuecomment-3436081436) in #1 

I'm open to changes in the timing of things. As this PR stands, issues are marked "stale" after 15 days, and closed 45 days later if no activity. PRs are ONLY marked stale after 30 days, they will never be closed automagically.


aknrdureegaesr:
In other projects, I have seen my issues not being reacted upon, then they turned stale, then the were closed. I usually was really happy when my issues were closed without reaction.

No, just kidding. What I actually think: Autoclosing heaps insult on harm. To do that on someone who tries to be nice and help us out with reporting stuff is plain rude. I think auto-closing issues is toxic behavior towards the community.

I think any issue should be closed when it has been addressed. That part may happen automatically, e.g., if a commit that has a comment "fixes # soandso" reaches the main branch.

Alternatively an issue should be closed if it will not be addressed (in some long-term foreseeable future). In my opinion, that should never happen automatically. It should be done by a responsible person who explains the rationale for that action.

Why do you think autoclosing is beneficial? For the time being, I **object** to this PR.

PR #14, opened 2025-10-27
Wmiler ci linux 2

wmiler:


.github/workflows/ci-linux.yml
```diff
@@ -0,0 +1,231 @@
...
+        key: ${{ runner.os }}-${{ steps.which-os.outputs.thisos }}-${{ runner.arch }}-boost
+        path: /home/runner/work/JS8Call-improved/JS8Call-improved/local/boost
+        
+## QT to change between 6.8.3 and 6.9.3 just change the 8 to a 9 everywhere here and at the bottom
+    - name: Cache Check QT 6.9.3
+      id: cache-qt
+      uses: actions/cache@v4
+      with:
+        path: /home/runner/work/JS8Call-improved/JS8Call-improved/local/qt693-install
+        key: ${{ runner.os }}-${{ steps.which-os.outputs.thisos }}-${{ runner.arch }}-qt693
```
aknrdureegaesr: I like it that you embed the version number 693 into the cache key for qt.

Would that not be beneficial to also do for the other libraries (hamlib, fftw, ...)? We'll use different versions of those some time in the future.

.github/workflows/ci-linux.yml
```diff
@@ -0,0 +1,231 @@
...
+      with:
+        path: /home/runner/work/JS8Call-improved/JS8Call-improved/local/libusb
+        key: ${{ runner.os }}-${{ steps.which-os.outputs.thisos }}-${{ runner.arch }}-libusb
+
+    - name: Install libusb 1.0.29
+      if: steps.cache-usb.outputs.cache-hit != 'true'
+      id: install-usb
+      run: |
+        mkdir -p tmp
+        curl --location --output ./tmp/v1.0.29.tar.gz https://github.com/libusb/libusb/archive/refs/tags/v1.0.29.tar.gz
```
aknrdureegaesr: You download to `./tmp` here, and to plain `.` in other, similar cases below.

It doesn't matter much, but I think I like the `./tmp` better. I recommend to consistently use the same target directory for all `curl` downloads.

.github/workflows/ci-linux.yml
```diff
@@ -0,0 +1,231 @@
...
+      with:
+        path: /home/runner/work/JS8Call-improved/JS8Call-improved/local/qt693-install
+        key: ${{ runner.os }}-${{ steps.which-os.outputs.thisos }}-${{ runner.arch }}-qt693
+
+    # Takes about 1 hour and then will be cached
+    - name: Install QT 6.9.3
+      if: steps.cache-qt.outputs.cache-hit != 'true'
+      id: install-qt693
+      run: |
+        git clone https://github.com/qt/qt5.git qt6
```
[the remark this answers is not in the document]
wmiler: Eh, it's not an issue for the runners, and doesn't use that much space either.

.github/workflows/ci-linux.yml
wmiler: Oops, missed that one. That was a workflow debug thingie.

Issue #18, opened 2025-11-05
UI inconsistencies within panes


wmiler:
Version: 2.4.0-Improved
OS: Win10

Description:
One would expect to be able to sort the Call or Band Activity panes by column header selection. This doesn't work. The pane(s) can only be sorted view the menu item View->Sort Call/Band Activity by... 

Expected:
If the column headers allow you to push them, them should be selectable for sorting.

Resolution:
Either hook the column headers up to allow the sort, or disable being able to push them.

Chris-AC9KH:
From my standpoint, I probably set these the way I wanted long ago and never looked at it again. But I think I would have to agree with Wyatt that I would expect to be able to sort on header click. However, the right click context menu is also not "bad" IMO. It allows the user to make two settings in one menu, and is not a bad design.

The mouseover on both tables is just plain annoying (to me). Sometimes if there's something like a MSG being transmitted I might like to mouseover on that to "copy the mail". That darn mouseover box pops up and follows my mouse pointer around and I don't particularly care for that.

rruchte:
@Chris-AC9KH You mean that "Received callsigns are displayed with time since last heard, SNR, and grid locator (if reported)" tooltip? I dislike that too, I don't think it adds any value, I can yank that out while I'm in there.

I'm split on the click-to-sort headers. On one hand, it's pretty standard behavior, on the other, I've never changed the default sort, so I'm not sure it's worth the effort. The more I think about it, the more it bothers me though.

rruchte:
The header sorting does not play well with the custom sorting that is implemented in the app currently, which is probably why it's not enabled. I've disabled the header interactions and tooltips in PR #23

Chris-AC9KH:
> I'm split on the click-to-sort headers. On one hand, it's pretty standard behavior, on the other, I've never changed the default sort, so I'm not sure it's worth the effort. The more I think about it, the more it bothers me though.

Rob, I like your PR as-is. I think this is something that most people set once and forget it. I'd like to have @wmiler take a look at it since he raised the issue. Wyatt, if you like it I'd say to just merge it. Getting rid of the annoying mouseover popups, I think, is a really good idea. It makes the UI cleaner and less "busy" when there's already alot going on in the first place.

PR #23, opened 2025-11-08
Disable confusing interactions in the band activity and callsign tables

rruchte:
Addresses #18 

Disables mouseover and click behavior on table headers in band activity and callsign tables
Disables "Received band activity is displayed" tooltip in band activity table
Disables "Received callsigns are displayed" tooltip in callsign table

The interaction in the table headers seemed to indicate that the headers could be used to sort on click, but this is not the case. To prevent confusion, these interactions have been disabled.

The tooltips in the tables are unnecessary, they simply describe what the tables contain. Users have reported annoyance at their constant presence.


wmiler:
Looks good, thanks Rob!

Issue #16, opened 2025-11-02
Switching bands leaves the split frequency on the previous band.


aknrdureegaesr:
I'm working on the 40 m band. I've been using a fairly low audio frequency, maybe 1200 Hz or so.

I use Settings / Radio / Rig Options / Split Operations: Rig.

I use the JS8Call-improved frequency UI (under the big QRG display in the upper left corner) to change to the 20 m band.

Expected and seen: The Frequency changes to 14.078 kHz, both on the screen and on the rig's display. But the split frequency is still at 7.077.

I hit the "tune" button.

Expected: The rig sends a test signal in the 20m band.

Actually seen: The rig sends a test signal back in the 40 m band.

Workaround: Click a high audio frequency, like 2200 Hz, and hit tune again.

rruchte:
Hmmm. The bug I fixed prevented the TX frequency from updating unless a transmission was occurring, and then it would incorrectly set the TX frequency to the RX freq. This looks like something different that has been uncovered by fixing that logic bug. I run in split mode and have not seen this happen. I'm away from my radio for a few days, if someone else want to look at this, have at it. The fact that it does not occur consistently seems like a clue.

rruchte:
@aknrdureegaesr The way the split mode frequencies are set using two nearly simultaneous calls that both result in a transceiver update has always seemed problematic to me. There could be a potential race condition (maybe even in the radio) that's causing this now that the code is actually running. I made [a commit in my repo](https://github.com/JS8Call-improved/JS8Call-improved/commit/159dee067f41d3ee315c0ac919e2d4e4c8d93d24) that sets the split mode RX and TX frequencies in one transceiver update. If you can, give this a try and see if it solves your problem.

rruchte:
@aknrdureegaesr Have you have a chance to test my changes?

aknrdureegaesr:
I added a bit of logging. Here is what the logging says:

```
mainwindow.js8: 25-12-16T00:17:43.001Z: For incoming audio of  1594 setting tx frequency to 28078000 and transmit_frequency to 1594
mainwindow.js8: 25-12-16T00:17:43.001Z: Setting PTT to on
mainwindow.js8: 25-12-16T00:17:43.104Z: MainWindow::handle_transceiver_update: Transceiver::TransceiverState(online: yes Frequency {28078000Hz, 28078000Hz} Transceiver::DIG_U; SPLIT: on; PTT: on)
[...]

PR #26, opened 2025-11-09
Fixes logic bug that prevented the split mode TX frequency from being set on frequency change

rruchte:
Addresses #16 

Fixes logic bug in MainWindow::setRig that prevented the split mode TX frequency from being set correctly when the frequency is changed.


Chris-AC9KH (approved):
This LGTM. Requested a review for merge by @aknrdureegaesr

aknrdureegaesr (commented):
👍

PR #27, opened 2025-11-10
Fix for deprecated setTimeSpec in Qt 6.9

Chris-AC9KH:
This will fix the compiler complaints re: deprecation of setTimeSpec in Qt 6.9 and later. Addresses issue https://github.com/Chris-AC9KH/JS8Call-improved/issues/25


wmiler (approved):
I don't see any issues.

Issue #9, opened 2025-10-25
Reporting to PSK reporter: Several bugs.


aknrdureegaesr:
* In their [documentation](https://pskreporter.info/pskdev.html), PSK reporter asks the same call should not be reported twice in the same hour unless there's been a significant change. Our code will report the same call once every 5 minutes. Or at least that is what it tries to do.
* What it actually does: There is a trivial bug causing the present code to check whether the cache entry is younger than 300 seconds into the epoch. If so, it is considered old enough to be ignored.
* Also, the present code does not remember the band of the observation. So if several people change bands together, they will not report each other before the respective cache entries expire.
* The PSK reporter documentation suggests to add randomness to the sending interval. This is not done.
* I have seen it happen that PSK reporter did not report, in my name, stations that I've heard. Trying to find the place in the code that caused this resulted in the other findings.

## Mildly related attempt at humor

Part of this reminds me of a coveted industry joke:

> There are four hard things in IT:
>
> - Coming up with good names.
> - Cache invalidation.
> - Counting.

(I've seen all of these problems in the present code.)

aknrdureegaesr:
Can someone explain to me what the perceived advantages of this "pointer to impl" thing are that I keep seeing?
 
In my opinion, the compiler doesn't care much, but it somewhat hides the implementation needlessly from us developers, making our life harder.

aknrdureegaesr:
Also, all stations heard above 49 MHz are always reported to PSK reporter, bypassing the "slow down repeated reports" functionality. This exception was first introduced by @bazineta in Git commit 1188110a6 dated "Wed Sep 4 14:53:21 2024 -0700", but there is no comment explaining why signals above 49 MHz are to be reported again and again. The git commit's log message does not explain this, neither does anything else in that git commit I've found.

Alan? Do you have any insight to share why you did it that way?

(I'm for removing this weird special rule.)

bazineta:
The pimpl idiom is extremely common in C++.

https://en.cppreference.com/w/cpp/language/pimpl.html

There are arguments for and against; as template metaprogramming has become increasingly common and compilation times in general are no longer measured in lengthy coffee breaks, I think the against begins to win out at the moment in code of this nature. Codebases with many millions of lines in the 90's, we used this religiously, but at present it's a bit overkill in most contexts. We do, however, use this technique without fail in library design for ABI stability, so even now, it's a handy thing in the right context.

[...]

PR #33, opened 2025-11-12
PSK Reporter Updates

rruchte:
Updates comment in callsign cache routine to indicate band-awareness

### DIFF
--- a/PSKReporter.cpp
+++ b/PSKReporter.cpp
@@ -46,7 +46,7 @@ namespace
 
   constexpr auto             HOST               = "report.pskreporter.info"_L1;
   constexpr quint16          PORT               = 4739; // 14739 for test
-  constexpr int              MIN_SEND_INTERVAL  = 120;  // in seconds
+  constexpr int              MIN_SEND_INTERVAL  = 600;  // in seconds
   constexpr int              JITTER_MAX  		= 5;    // in seconds
   constexpr int              FLUSH_INTERVAL     = 125;  // in send intervals
   constexpr qsizetype        MAX_STRING_LENGTH  = 254;  // PSK reporter spec
@@ -693,7 +693,8 @@ PSKReporter::addRemoteStation(QString           const & call,
                                QString          const & grid,
                                Radio::Frequency const   freq,
                                QString          const & mode,
-                               int              const   snr)
+                               int              const   snr,
+							   QDateTime 		const & utcTimestamp)
 {
   m_->check_connection();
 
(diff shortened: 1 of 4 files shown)


aknrdureegaesr:
Not sure how you'd want me to review this. Do you want to simply improve on what we have and leave some problems to be dealt with later, or do you want to fix all known problems?

If the later, there is more to be done.

## Check packet full

To quote https://pskreporter.info/pskdev.html :

> The datagrams should be sent at a rate of no more than one every five minutes (unless the packet becomes full).

I have not looked into this in detail, but this "unless the packet becomes full" promises to become tricky: We need to test whether we have enough data to fill the packet to the brim and send it. This test is most conveniently done the moment the new information comes in from the decoder. Yet, we would do a bad job if we sent right then and there. That would make sending more probable at the times when signals complete in JS8. Once we have a more sizable user crowd than we do now, this would result in load spikes at PSKReporter.

I suggest to use a random wait time that, in its maximum, is the same than the current mode's period, once enough data to fill a packet has become available.

## Precision timer

I have found in the Qt documentation: When you do **not** use precision timers, Qt tries to run several timers' triggers at the same time. This could potentially concentrate network sending triggers to such times as when the protocol has activities. So, even though this may look like overkill (and probably is, though I'm not sure), I would feel better if we used a Qt precision timer here.

## Put network stuff out of the main thread

I have not looked into this in detail, but I find it quite suspicious that `addRemoteStation` doesn't take part in Qt's signal+slot mechanism. I think it should.

And I think network stuff belongs into a thread of its own. If a user switches TCP/IP on and connection attempt fails with a timeout, there is danger that this will freeze the UI.

## It is "all five minutes"

... so the `min_send_interval` should be increased from 120 to 600.

## Send last available info, not first

If several transmissions come in for the same station on the same band while we're waiting for our 5 minutes to elapse before sending again, I would argue the sender wants to know about the last transmission, not the first; but our present code sends the first, not the last.

## Avoid lifecycle issues

I'm a bit concerned about the line

```
      m_->spots_.enqueue({call, grid, snr, freq, mode, DriftingDateTime::currentDateTimeUtc()});
      m_->calls_.insert(cache_key, std::time(nullptr));
```

The objects behind `call` and `cache_key` will go out of scope and disappear. If I have the rules correctly, `cache_key` disappears when the function is complete, while `call` is an object controlled by the caller that may disappear whenever the caller decides it no longer needs it. So I think we need to copy before pushing stuff into our long-term data structures, and we need to delete those copied objects when we remove them from those data structures. I do not see any of that here. But maybe my C++ is rusty and I'm all wrong?

Finally, we are in a Qt project here, so I would slightly prefer to we standardized on `QDateTime` instead of using `std::time`. But that may be more a matter of taste and doesn't really matter much.

rruchte:
> I have seen (and partly repaired) the code that does the ALL.TXT writing. It writes the timestamp that catches the beginning of the reception of the received data.

I saw that after some debugging and a close examination of ALL.TXT output.

Before we just close this out with the simple comment change, do we want to try to roll a handful of improvements in in the interest of keeping the changes with this discussion for historical purposes?

For instance:
> If several transmissions come in for the same station on the same band while we're waiting for our 5 minutes to elapse before sending again, I would argue the sender wants to know about the last transmission, not the first; but our present code sends the first, not the last.

This would simply be a matter of updating the frequency and timestamp in the cached spot if one is found, instead of a noop. Since some of these issue #9 changes are related to this cache, it makes sense to go ahead and do that in this context.

> For ALL.TXT writing and other purposes, we already have code that calculates the start of a transmission. That timestamp should be handed to this functionality here and used when sending data to PSK reporter.

This is an issue that goes all the way up to `MainWindow::processDecodeEvent`, where all of the timestamp values in the detail structs are set to the current time with `DriftingDateTime::currentDateTimeUtc()`, not using the start of receipt time.

Example
```cpp
          //ActivityDetail d = {};
          d.isLowConfidence = decodedtext.isLowConfidence();
          d.isCompound = decodedtext.isCompound();
          d.isDirected = decodedtext.isDirectedMessage();
          d.bits = decodedtext.bits();
          d.dial = freq;
          d.offset = offset;
          d.text = decodedtext.message();
          d.utcTimestamp = DriftingDateTime::currentDateTimeUtc();
          d.snr = decodedtext.snr();
          d.isBuffered = false;
          d.submode = decodedtext.submode();
          d.tdrift = m_wideGraph->shouldAutoSyncSubmode(d.submode) ? DriftingDateTime::drift()/1000.0 : decodedtext.dt();
```
I've made changes locally to pass the timestamp from the `CallDetail` through to `PSKReporter::addRemoteStation` so we're not generating new timestamps there, but if we want to change to use the start of receipt time as the canonical time for all events instead of setting the current time when we _process_ the event, that's a bigger task that I think deserves it's own issue. From my first glance though it looks like we can just use `decodedtext.time()` for that. People with more time spent in that area of the app can fact check me on that.

After my change to pass the `CallDetail` timestamp, that change would flow through to `PSKReporter::addRemoteStation`. There may be (probably are) similar cases elsewhere.

rruchte:
I have pushed a set of changes that address the requests that I believe are in scope of the original issue, new issues have been created that can serve as context for broader changes.

-Increases MIN_SEND_INTERVAL to 5 minutes
-Changes PSKReporter::addRemoteStation to use timestamp passed as an argument instead of generating a new timestamp
-Changes spot handling in PSKReporter::addRemoteStation to keep queued spots updated with the newest info, rather keeping the oldest
-PSKReporter::addRemoteStation refactored to remove multiple calls to std::time(nullptr), increase clarity, and enhance Fahrvergnügen

I ran a build with these changes with enhanced logging during a fairly active net tonight, it behaved the way it is supposed to.

Issue #4, opened 2025-10-24
Documenting the project


wmiler:
To better document the project as we go along, I've pushed to main a Doxyfile for the [doxygen](https://www.doxygen.nl/) generator, which is cross platform. Right out of the box, it generates some interesting docs. Those docs are placed in ./docs/html/

As part of a much longer series, I'm planning to start documenting the code itself, as well as the API with @jfrancis42 I expect as people touch the code, they should document as they go 😀

Basic instructions can be found in the [doxygen manual](https://www.doxygen.nl/manual/docblocks.html).
I'd recommend the /** some description */ method.

I don't expect massive quantities of docs, just enough that somebody can figure out what's going on.

Thanks!

wmiler:
@Chris-AC9KH Yeah, I will go ahead and close it here. I can catch things thru the PRs.

PR #41, opened 2025-11-14
Create workflow to build docs

wmiler:
This will build the source code docs as it stands now, and publish it to GitHub Pages.

This workflow is triggered only if we have labeled the PR with "documentation", a Release has been published or a manual trigger.

@Chris-AC9KH There's some stuff you may need to turn on, or you can wait on this until we go to Org.
This is setup for "gh-pages", don't merge until you set it :)

> You must configure your repository to deploy from the branch you push to. To do this, go to your repository settings, click on Pages, and choose Deploy from a Branch from the Source dropdown. From there select the branch you supplied to the action. In most cases this will be gh-pages as that's the default.

Part of #4

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,23 +1,25 @@
 name: Build and Deploy Docs
 # Trigger only if we have labeled the PR, a Release has been published or a manual trigger
 on: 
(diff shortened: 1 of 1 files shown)


wmiler:
That's fine, and was expected. It's here for when that happens. :)

PR #51, opened 2025-11-19
Add .clang-format configuration file

wmiler:
This is just for **testing!!!** Uses **clang-format-18**.

Please give it a whirl, and let me know if this totally breaks your coding style. One note, and something I didn't even know was in the c++ standards was the use of apostrophes in large numbers (from c++14).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -182,7 +182,9 @@ PointerAlignment: Left
 PPIndentWidth:   -1
 QualifierAlignment: Leave
 ReferenceAlignment: Pointer
-ReflowComments:  IndentOnly
+# IdentOnly is available only in 19+
+#ReflowComments:  IndentOnly
+ReflowComments:  false
 RemoveBracesLLVM: false
 RemoveParentheses: Leave
 RemoveSemicolon: false


aknrdureegaesr:
Excellent! I was considering doing this, but you were first.

Can you please simply run this and open a PR with all the changes to the source this does, so that we can discuss coding style? (Not intending to merge that PR as long as other PRs are still open, maybe.)

aknrdureegaesr:
This is not about "totally breaks your coding style". But I like the bare, unconfigured clang style better.

* With its smaller indentation level and the placement of curlies, code becomes more concise.
* There is a school of thought that, to minimize readers' mental load, the most important stuff should come first in a line. Your style puts "," first in a line when a long list needs to be broken up. I think this looks weird.
* I think it valuable to have a predictable order of include statements. People seldomly actually read include statements, but a predictable order here sometimes reduces merge conflicts.

aknrdureegaesr:
I took a closer look.

Please add 

```
# See https://clang.llvm.org/docs/ClangFormatStyleOptions.html
```

at the top of the file, so people know where to find the documentation.

I think the funny comma thing is repaired by

```
BreakConstructorInitializers: AfterColon
```

The include sorting is

```
IncludeBlocks:   Regroup
```

together with

```
SortIncludes: true
```

I'm also a bit afraid the clang-format would ruin our documentation comments unless we say

```
ReflowComments:  false
```

In the past, we had a mixture of different line endings, partly in the same file. To not have that happen again:

```
LineEnding: LF
```

wmiler:
Ok, give that a whirl. Turns out there is a fix for the ReflowComments, but only if you have clang 19+, which I don't.
I think there was one other thing I did not change from @aknrdureegaesr on your request (it may have been the commented EmptyLines at end of file). Also, thanks for finding the DecimalMinDigits! Much better!

wmiler:
This can merge whenever.

PR #60, opened 2025-11-22
Automatic changes produced by: clang-format-18 --Werror -i *.h *.hpp *.cpp

aknrdureegaesr:
Another opportunity to check whether we like the code changes our shiny new `.clang-format` does.

If we do, we can merge that whenever only pull requests by me are open.


aknrdureegaesr:
Ooops! And we have a compilation error under Windows.

> error: 'MessageBoxW' is not a class, namespace, or enumeration

where the code never mentions `MessageBoxW`, it is all `MessageBox` without trailing `W`.

wmiler:
Just as a general info comment. 
The clang-format **is not**, and will not be enforced until we get everybody happy with it. That will require a change to the CI files.

Chris-AC9KH:
Frankly, I haven't had much luck with it in the past. It forces a particular coding style that some developers are not going to adhere to anyway. You have to remember clang was developed by Apple, using the LLVM back end for Objective-C. It was initially a proprietary project that they ended up open-sourcing and licensing under the Apache license. Today, just because of the open sourcing there is many variations of clang - Apple clang is not compatible with clang on linux, etc.. Linux tries to modify it to be compatible with GCC. Since Apple switched to Swift it is modified for Swift. And Swift is basically Objective-C. Obj-C uses manual memory management, where Swift uses ARC. Swift is just about 3x faster than Objective-C, and over 8x faster than Python. While Swift is fully interoperable with Obj-C, the Apple clang compiler has still been optimized for Swift.

So it will be an interesting experiment here to see how clang-format works but it's likely going to change every file in the codebase simply because it will try to force the Objective-C coding style onto C++ code.

Chris-AC9KH:
So MacOS Xcode has clang-format built-in. But it's a lot safer than what your'e doing here.

Screenshot 2025-11-24 at 21 40 25

This is all native to MacOS and Xcode so we'll see how this turns out after you get it working. Xcode runs the code in real-time to make sure nothing breaks.

Andreas, after the HB/CQ timer update Xcode pointed out that you left a dead variable in the code in the Modulator. Do you want me to just comment that variable out and leave it there, or remove it?

wmiler:
> > Through our re-sorting of `#includes`, this is now probably done too late, the damage already done. Didn't fully investigate.
> 
> Where it broke was in main.cpp by resorting the includes. I ran Apple clang-format on it with Xcode and fixed it. You're using a different variety of clang than I got, since Apple clang is proprietary and Apple doesn't release the source code for it. You're likely using linux clang. Apple uses the LLVM backend and style, you're likely using WebKit. Apple Clang is optimized for Swift/Objective-C, you're likely using the "horse of all colors" clang.
> 
I believe you are correct that the example I took the format from is WebKit. My run thru is using the Linux clang-format version 18, which is standard with Ubuntu 24.04. I'm not a huge fan of clang, it seems to cause more problems than it's worth. 🤷🏻‍♂️ 

The only goal here was to standardize the code format, and in its current incarnation, it's not working correctly.

Chris-AC9KH:
Well, LLVM/clang is the industry standard. Just about every Fortune 500 company has switched to it because it produces better code than MSVC and is more flexible than gcc. And since it's licensed under Apache 2.0 the big companies can build their own proprietary versions of it without releasing the source code. Microsoft was forced to go to it for arm64. Apple, of course, developed it with LLVM and released the version you're using under open source. It is the standard for building apps on mobile platforms, including Android.

So MacOS 26/Xcode 26 is using a variant of the open source clang 17. I ran it in Xcode against master, checked out this PR branch and ran diff to see what it's doing. Apple clang is not re-sorting the includes and I think that's what's breaking your experiment here. The other advantage with Xcode is that you can select a block of code in the IDE, run clang-format and it doesn't reformat the whole file. I rather like the Objective-C/LLVM formatting style because it's what I'm used to and it makes the code more readable. With C++ it doesn't seem any two coders have the same coding style and the arguments can go on ad infintum as to who's got the "correct" or "best" style. Clang enforces a standard but it is based on Objective-C, not C++. Objective-C comes from proprietary Unix systems like MacOS, it was the evolution of Dennis Ritchie's invention of C when he and Ken Thompson invented Unix at Bell Labs.

So while I believe clang-format could be used here to clean up the code clutter from different coding styles, IMO it would have to be done selectively with a tool like Xcode to prevent breaking things.

aknrdureegaesr:
> I would assume it could be but I wouldn't do it (at least on Mac).

The whole point of automated code formatting tool is that every contributor uses the same tool.

The current state of our code is a disgrace. To improve at least those parts that can be helped with by an automated code formatting tool, let us agree on such a tool, use it, and **require every future commit to be formatted with that tool as well**, or, to be precise, formatted compatibly with that tool's code formatting.

The "doesn't work under Windows" with the free version is a non-issue, as the `#undefine` hack of our predecessors can be improved on easily by renaming that one class.

Chris-AC9KH:
So I came up with an idea on how to do an easy comparo between the Xcode formatting and this one. Just delete that PR that I created from the experiment one, then try to create another PR that would merge it with this branch. There's no way it will merge but the web-based comparison tool will show what's different.

Chris-AC9KH:
Also, what I would do here is delete the PR and branch. Rework your .clang-format file, create a new branch and run it on the new branch to see what it does. Technically, you could start over with reset HEAD --hard to fix this one, but it won't be a very clean commit history, and I think this branch is already behind master anyway.

Chris-AC9KH:
> @Chris-AC9KH I'm ok with any fixes that works for the team, and doesn't break the code. I understand that all the platforms have their own insanity when it comes to this. It might be a good test just to switch the IncludesSort back to Preserve from ReGroup, as I recall, that was honoring the sort order regex.

Yeah, there's going to be other differences like function() formatting too if you use WebKit vs industry standard LLVM, and I'm guessing other developers who code much bigger projects are using LLVM standards.

The header sorting thing is easy enough to fix. There was some custom sorting because somebody didn't want to include windows.h for a Windows API call. So either change the name of it, or define it and include windows.h. The ifdef would be a lot cleaner than reworking the code to change the name.

Again, I'd get it so it works, then request to have Allan look at it.

jsherer:
I recommend not doing this:
image

It makes the history harder to reason about (git blame) and is usually no additional value / just cosmetic. 

Instead, I'd normally set up a lint step and format new and updated code as it's touched. A simple git pre-commit hook can go a long way.

PR #64, opened 2025-11-23
Update js8call.desktop

wmiler:
Update desktop file for linux with correct exec name. Any reason to change the icon name? If not, feel free to push.

### DIFF
--- a/js8call.desktop
+++ b/js8call.desktop
@@ -2,7 +2,7 @@
 Version=1.0
 Name=js8call
 Comment=Amateur Radio Weak Signal Operating
-Exec=js8call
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
This is necessary and it is not necessary to change the icon name. This is essentially in the linux build scripts, which create their own desktop file and populate it during install. When I originally wrote those scripts I did not want to use the ready-made desktop file as the scripts install the binary and icon in different places than AppImage will use. I didn't figure it would be a good idea to modify the existing one, as the scripts were considered a temporary solution.

Chris-AC9KH:
Merged. Thanks!

PR #66, opened 2025-11-23
Fix Software Update feature

Chris-AC9KH:
This feature hasn't worked since js8call v2.2.0. It is now working again for JS8Call-improved.

Screenshot 2025-11-23 at 17 31 07

Some caveats
- does NOT work with development software, the feature is disabled. To test you have change Versions.cmake to produce a production build from this branch (Edit: I guess it does work with development software. But in practice this will never be the case.)
- it requires a text file called `version.txt` to be present in the latest production release downloads, populated with one line - the current semantic version number for that release.
- the current URL points to the github releases. However, I believe this can be changed later to point to the website once the website is up to speed. @mgochoa57 can comment on that, as I believe it is intended to link to the newest release from the website, which is a "cleaner" URL than the github one.
- I experimented with putting version.txt in the docs folder in the source tree. It did not work as the software cannot parse the file from there.


Chris-AC9KH:
Additional comment. This is an example where void `MainWindow::checkVersion(bool const alertOnUpToDate)` could be pulled out of mainwindow eventually and called as an include. That monster needs to have something done with it. But when it happens I prefer the surgical approach instead of the shotgun approach. The shotgun approach is guaranteed to break it.

Chris-AC9KH:
> Suggest you change line 1461 to wrap the web address with HTML tags so that it can be a selectable blue-underlined URL in the pop-up window, thusly (notice I had to escape the " marks with a \ to make it work):

Good suggestion, Joe. I will play with that a bit and test it on MacOS as well.

Chris-AC9KH:
So on Debian 13 (Gnome) if formats differently than on MacOS. I need to get rid of the line break tag and that will "fix it" on linux, won't make any difference on MacOS.

PR #68, opened 2025-11-25
Pr/clang format experiment

Chris-AC9KH:
Experiment to see if code will build on all platforms after being reformatted with Xcode clang-format. Done on MacOS 26.1 Tahoe with Xcode 26


Chris-AC9KH:
Further explanation of clang-format, since this experiment came from [this one](https://github.com/JS8Call-improved/JS8Call-improved/pull/60) where the build fails on Windows.

This is Apple's implementation of C++ formatting. It came from Objective-C, which came from C, which came from Dennis Ritchie when he and Ken Thompson invented Unix at Bell Labs. So Objective-C was pretty entrenched in the Unix world and Apple used it in early versions of MacOS. Then Chris Lattner (Vikram Adve and Chris Lattner were the original inventors of LLVM) at Apple invented Swift in 2010 as a replacement for Objective-C and Apple released it in 2014.

Apple's clang compiler is proprietary - they don't release the source code for it. Apple developed clang with LLVM, with some input from Microsoft, Google and some others. Previous to MacOS 10.6 they used a heavily modified version of gcc, the switch to LLVM was made in 2005 and then clang came in 2009. Apple originally was going to keep clang proprietary, but they released it to the open source community under the Apache 2 license. But the open source versions of clang are not Apple clang. Today, LLVM/clang is pretty much the industry standard but companies that use it (Google, Microsoft, Intel, IBM et al) all have their own proprietary versions of it because the license permits that.

Since many MacOS software projects use C++, Apple made Xcode to be compatible. But their clang-format does some things different than many programmers' C++ coding style, such as placement of curly braces for functions or statements, indenting and code spacing. Apple's claim is that it follows the LLVM standard and makes the code easier to read. But some C++ programmers (and no two seem to have the same coding style) may not like it because it doesn't fit their style, as its roots came from Objective-C. And there is significant difference in include header sorting between C++ and Objective-C, which is where I think the experiment in the other PR failed.

So this was just an experiment to use a different version of clang-format than the open source version to see if it would work. I never intended it for merge, just as a comparison experiment. Unlike the open source clang-format it is not a command line tool on MacOS - it is built into Xcode.

PR #71, opened 2025-11-27
Rename MessageBox to JS8MessageBox as a safer fix to Microsoft's #define MessageBox madness.

aknrdureegaesr:
When compiling for MS Windows, the usual suspect MS compilers `#define` a macro `MessageBox`.

We previously tried to stay sane in this madness by `#undef MessageBox` , but this turned out to be an unstable fix in view of updates that our intended code format utility does.  For this specific reason and for general stability, it is better to evade the problem by renaming our `MessageBox` class to something else. This systematic renaming is done in this PR.

### DIFF
--- a/JS8MessageBox.hpp
+++ b/JS8MessageBox.hpp
@@ -3,21 +3,19 @@
 
 #include <QMessageBox>
 
-// get rid of the nasty MS define
-#ifdef MessageBox
-#undef MessageBox
-#endif
-
-//
-// MessageBox - wrap the Qt QMessageBox class to give a more platform
-// 							neutral and functional interface
-//
-class MessageBox
+/**
+ * JS8MessageBox - wrap the Qt QMessageBox class to give a more platform
+ *							neutral and functional interface.
+ *
+ * Microsoft chose to #define MessageBox.
+ * We dogde the resulting problems by calling our MessageBox JS8MessageBox.
+ */
+class JS8MessageBox
   : public QMessageBox
 {
 public:
-  explicit MessageBox (QWidget * parent = nullptr);
-  explicit MessageBox (Icon, QString const& text, StandardButtons = NoButton
+  explicit JS8MessageBox (QWidget * parent = nullptr);
+  explicit JS8MessageBox (Icon, QString const& text, StandardButtons = NoButton
                        , QWidget * parent = nullptr
                        , Qt::WindowFlags = Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint);
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -251,7 +251,7 @@ target_sources(
   main.cpp
   mainwindow.cpp
   Message.cpp
-  MessageBox.cpp
+  JS8MessageBox.cpp
   MessageClient.cpp
   MessageError.cpp
   messagereplydialog.cpp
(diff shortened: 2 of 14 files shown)


Chris-AC9KH:
Somebody used a Windows API call that's been around longer than any of these ham radio programs. It's not Microsoft's "madness", it's a programmer's failure to include windows.h when you make a Windows API call. That being said, I see nothing wrong with this fix either.

JS8MessageBox.hpp
```diff
@@ -3,21 +3,19 @@
...
-//
-// MessageBox - wrap the Qt QMessageBox class to give a more platform
-// 							neutral and functional interface
-//
-class MessageBox
+/**
+ * JS8MessageBox - wrap the Qt QMessageBox class to give a more platform
+ *							neutral and functional interface.
+ *
+ * Microsoft in its "wisdom" chose to #define MessageBox.
```
Chris-AC9KH: Get rid of the Microsoft comment out of the code. We don't need here since that API call has been around longer than this program has.

aknrdureegaesr: In other words, that comment reflects my current understanding of the problem we ran into. But I'm willing to learn from you.

Chris-AC9KH: Thank you. FYI, the Microsoft MessageBox class has been in the Windows API, and in the .NET framework, since Windows 3.0 - that's 35 years. As a programmer, you're expected to know that and if you're going to compile this code on Windows and are going to use it, implement it correctly. Microsoft did nothing badly, a programmer who wrote this code did. And used a "hack" to get around it.

Yes, we all know you don't like Microsoft. Check it at the door when you come in because there's millions of people who use Windows every day. In our current 2.4.0 release stats there's 422 downloads for the Windows build vs only 31 for linux. Windows is our primary user base, Mac is number 2, linux is dead last. Keep that in mind when you decide to put offensive remarks in the code that now appear in the code documentation, without even knowing what you're talking about where that class in the Windows API is concerned. Any time you're dealing with cross-platform libraries like Qt, implement it correctly if you're going to use it, because the Qt macros in Windows Qt do.

aknrdureegaesr: I am arguing based on the suggestion you made:

> How about a programmer using it without including the proper headers?

If that is the thinking behind it, I would say that even 35 years ago, there were better and there were worse ways. Forcing a name down a person's throat who has never called upon the appropriate include file I would count among the worse ways. And just because many people used it does not make it any better.

But your suggestion might have led me into a wrong direction. It would actually make more sense if that `#define` precisely lives in some appropriate Microsoft header file. Then our problem might be that Qt did not manage to write their header files in a fashion that prevents the MS header file and its definitions leaking into the Qt includes we do. Incidentally, I see that this would be an excellent use case of the `pimpl` patter that was discussed here early. If that is what we ran into here, the snite-ness against MS is indeed misplaced here.

So thanks for your review.

aknrdureegaesr: As I'm not sure, and feel it's not worth it to go through the trouble finding out, MS deserves the benefit of the doubt. It was lack of careful thinking that made me think it was their fault.

I love reviews. You always learn.

aknrdureegaesr:
> mainwindow is probably the biggest kludge in the entire codebase and for a Qt application should've never integrated the UI with the back end.

Yes. Actually, for *any* GUI application, be it Qt or whatever, having application code mixed up with the UI code is a big bold red flag.

> That class is so poorly written hardly nobody can touch it without breaking something.

We have to accept our fate that we have to be diligent and careful *and* will still break things nevertheless. (Part of that fate is also that we will be heroes, though probably rather unsung ones.)

But hopefully, with each touch, the mess will become less messy and more manageable. The beginning is the hardest.

But even a long journey can be completed one step at a time...

PR #74, opened 2025-11-29
CI update in preparation for Release builds

wmiler:
This is Part 1 of 2  to get the CI system to be able to do Release builds on-demand.

Supported at this point is:
Linux Ubuntu 22.04/Debian AppImage
Windows Installer
MacOS dmg generation

The MacOS build is most likely not fully functional, based on conversations with Chris. Consider this a _Work In Progress_.

**Note**: On Standard CI builds, a fair amount of extra fluff is generated in the artifact files, this is for **DEBUGGING** purposes, and is meant to be there. Release versions will be just the applicable installer/AppImage/dmg file.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -54,14 +54,6 @@ jobs:
         echo "thisos=$ImageOS" >> $GITHUB_OUTPUT
       shell: bash
 
-    - name: Set our version tag for AppImage
-      if: ${{ inputs.release-ver }}
-      run: export VERSION=${{ inputs.release-ver }}
-      
-    - name: Set our version tag for AppImage
-      if: ${{ !inputs.release-ver }}
-      run: export VERSION=-devel-
-      
 # Move all of our cache checks here, so we know if we need to free bunches of disk space
     - name: Cache Check libusb 1.0.29
       id: cache-usb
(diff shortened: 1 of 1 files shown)


Joe-K0OG:
I can report that the AppImage works fine on Fedora as well as Debian/Ubuntu/Mint.  Good work @wmiler Wyatt!

73,
-Joe-
K0OG

wmiler:
Thanks Joe!

.github/workflows/ci-linux.yml
```diff
@@ -139,23 +175,23 @@ jobs:
...
       if: steps.cache-ffmpeg.outputs.cache-hit != 'true'
       id: install-ffmpeg
       run: |
-        git clone --depth=1 --branch n7.1.1 https://git.ffmpeg.org/ffmpeg.git /tmp/ffmpeg
+        git clone --depth=1 --branch n7.1.3 https://git.ffmpeg.org/ffmpeg.git /tmp/ffmpeg
         mkdir -p /tmp/ffmpeg/build
         cd /tmp/ffmpeg/build
         ../configure --prefix=${GITHUB_WORKSPACE}/local/ffmpeg --disable-shared --enable-static --disable-doc \
-          --enable-network --enable-gpl --enable-nonfree --enable-libass --enable-libfreetype --enable-libmp3lame \
+          --enable-network --enable-gpl --enable-libass --enable-libfreetype --enable-libmp3lame \
```
wmiler: This is the ffmpeg configure line that needs changing for license issues. @Chris-AC9KH

Chris-AC9KH:
Looks like the linker is failing at libavformat for some reason. That doesn't make sense. The Qt libs for linux were built with that command you changed it to. I know this branch will build on my Debian 13 arm64 system, so this has to be something unique to either the runner, or does your Qt package already have dynamically linked FFmpeg in it?

Joe-K0OG:
> Thanks Joe!

Today I also tested and operated on-air a bit with Ubuntu 22 and Ubuntu 24 x86_64 systems, all working fine with the AppImage from your repo.

73,
-Joe-
K0OG

wmiler:
It failed for both x86 and arm64, I've got some runs going on my repo to try and narrow it down, but it definitely is ffmpeg that's breaking it. The QT is building from source, nothing fancy going on there, just have the prefixs, no other configure options.

wmiler:
Ok, so this was a two-fold issue.
1) Switching to ffmpeg shared-only cause the AppImage build to fail
2) AppImage for whatever reason, is unable to ascertain which shared library name to use due to failing symlink logic

The fix, is simply to copy the ffmpeg libraries to the AppDir/usr/lib/ with the correct name before building the AppImage.
Then everything is happy. Update to ci-linux.yml coming shortly, if it passes here, we should finally be able to push this.

PR #82, opened 2025-12-04
RX Decoder optimizations for improved low SNR decodes

punk-kaos:
Some decoder optimizations to increase the sensitivity of the decoder. 
In THEORY these changes should net a +3-+6db sensitivity increase... 
With these changes I'm seeing consistent 'normal' mode decodes down to about -28db.

These changes are fully compatible with standard JS8Call's decoder. Nothing on the TX path has been touched. 

I've been running these changes for a few days now and haven't noticed anything breaking so I'm REASONABLY confident that they're good to go. Hopefully they're useful!

### DIFF
--- a/kalman.hpp
+++ b/kalman.hpp
@@ -5,6 +5,13 @@
 
 namespace js8
 {
+    /**
+     * @brief Lightweight PLL/Kalman-style tracker for residual frequency offset.
+     *
+     * Initialized with coarse estimates and sample rate; apply() rotates samples
+     * by the tracked offset, update() nudges the estimate using pilot residuals.
+     * Used inside the JS8 decode loop per candidate frame.
+     */
     class FrequencyTracker
     {
     public:
@@ -45,6 +52,14 @@ namespace js8
     {
     public:
 
+        /**
+         * @brief Tracks residual timing (sample) offset between the symbol clock and the signal.
+         *
+         * Initialized with bounds and step limits; update() ingests early/late energy
+         * errors from pilots to refine sample alignment. Used per candidate in the
+         * JS8 decode loop alongside FrequencyTracker.
+         */
+
         void reset(double initial_samples,
                    double alpha           = 0.15,
                    double max_step        = 0.35,
@@ -72,4 +87,3 @@ namespace js8
         int    m_updates     = 0;
     };
 } // namespace js8
-
--- a/ldpc_feedback.hpp
+++ b/ldpc_feedback.hpp
@@ -15,6 +15,14 @@ Q_DECLARE_LOGGING_CATEGORY(decoder_js8);
 
 namespace js8
 {
+    /**
+     * @brief LDPC erasure threshold config and feedback refinement helpers.
+     *
+     * Inline env readers expose thresholds/pass limits; the templated
+     * refineLlrsWithLdpcFeedback shrinks/boosts LLRs using the decoded
+     * codeword to retry LDPC. Used inside the JS8 decode loop between
+     * LDPC passes.
+     */
     constexpr float LLR_ERASURE_THRESHOLD_DEFAULT      = 0.25f;
     constexpr float LLR_FEEDBACK_CONFIDENT_MIN         = 3.0f;
     constexpr float LLR_FEEDBACK_UNCERTAIN_MAX         = 1.0f;
--- a/soft_combiner.hpp
+++ b/soft_combiner.hpp
@@ -19,9 +19,14 @@ Q_DECLARE_LOGGING_CATEGORY(decoder_js8);
 
 namespace js8
 {
-    // Helper for caching and combining repeated LLR frames. Template on
-    // the LLR vector length so the caller can bind it to the decoder's
-    // bit count without duplicating that constant here.
+    /**
+     * @brief Cache and combine repeated LLR frames for the same decode candidate.
+     *
+     * Uses a coarse freq/dt bin and a small LLR signature as the key; repeated
+     * receptions accumulate LLRs to improve decode probability without changing
+     * over-the-air behavior. Templated on the LLR length so the caller binds it
+     * to the decoder's bit count.
+     */
     template <std::size_t N>
     class SoftCombiner
     {
--- a/whitening_processor.hpp
+++ b/whitening_processor.hpp
@@ -16,6 +16,13 @@ Q_DECLARE_LOGGING_CATEGORY(decoder_js8);
 
 namespace js8
 {
+    /**
+     * @brief Compute per-tone/symbol noise medians and whiten LLRs for a JS8 frame.
+     *
+     * Given symbol magnitudes (sans Costas) and winners, produces normalized LLR0/LLR1,
+     * optionally applying noise-based whitening and erasure. Fully templated on matrix
+     * dimensions, so it stays header-only; used inside the JS8 decoder per candidate.
+     */
     template<int NROWS, int ND, int N>
     class WhiteningProcessor
     {


Chris-AC9KH:
@punk-kaos would you like me to label this WIP and then you can notify when ready for testing?

Joe-K0OG:
> @aknrdureegaesr @jsherer this actually looks good to me, at least the code looks sound. Requesting your review on it. I'm going to pull this one, build and test it on MacOS. @Joe-K0OG if you would be interested in testing this on linux as well, to see what you think.

I tested this with my weak-signal testbed, which Allan and I used for testing the code when we were working on 2.3.0.  I find no difference in performance between 2.5.0 (official) and the @punk-kaos source.  Both detect the same weak signal down to -26dB reliably, with an occasional -27dB decode.  Neither beats the other with weaker signals, and both appear to report the same dB measurement with very-weak signals.  I did not do any testing with stronger signals (above around -25dB), but was testing for differences in detection threshold.

My testing does not look for differences in performance when there are other signals present, but only when there is clean band noise.

My assessment is that this code variant does not improve the weak-signal detection/decode threshold.

73,
-Joe-
K0OG

Chris-AC9KH:
Thanks for the review, Joe. I'm building for MacOS with SDK 26 as we speak. I intend to test the difference with multiple signals where adjacent signals, even though valid, are passband "noise" for the decoder. Did you turn on the decode attempts display in the waterfall for your test?

Chris-AC9KH:
OK, I have this running. Testing with MacOS 26 Tahoe with my IC-7300. Being I'm testing on-air I'll probably want to run this for a couple days or so to evaluate the results. Being we're using a variant of Kalman filtering the measurements over time, including statistical noise, should produce estimates of extremely weak signals that tend to be more accurate. But it's going to take some operating hours to determine how much perceived difference there is.

I like the basic design of the filter.

jsherer:
I have this built and running. 

There's not an immediately noticeable impact (positive or negative). 

@punk-kaos did you happen to have a test harness for identifying / verifying performance?

Chris-AC9KH:
**Intermediate review:**
I've been running it for about 14 hours with no immediate noticeable difference. However, the real test hasn't come yet, so don't write this off.

Late tonight is our sked on 160 where we see if we can hook up with stations in Sweden and Russia that we exchange MSG's with. Conditions don't always allow it, but as we get further in the shorter days in the northern hemisphere it gets better on 160 and there's no such thing as a strong signal in the noise on MF running only 4 watts.

Keep in mind there's a lot of things that can affect this - the base noise in your area, the quality of your receiver and how it's set, antenna system, etc..

I'm quite familiar with Kalman filters and the design here is sound. We use them extensively in aircraft autopilot code. They are used not only in navigation systems, they are used in software defined radios for signal processing. You have to realize when testing that you not going to see any difference within the normal limits of the decoder. In theory, the design of the filter should provide some gain while attempting to filter the "noise". It's a linear quadradtic estimation, however the variants used in aircraft systems is an EKF (Extended Kalman Filter) for nonlinear systems.

So to actually see what the filter can do, it needs to be run in real world "noise" to the limit of your setup to see if it can separate signal from noise.

However, I have a question for @punk-kaos 

This kind of relates to a long-term goal we have for the JS8 code; is it possible to code the filter as a new class, call it kalman.cpp or whatever you want? Call that class from the decoder. The reasoning here is that we want to break down classes in mainwindow eventually, and sort the tree out in folders so the source tree begins to make more sense. In my experience with Kalman filters, I've never seen one that can't be "tweaked" for better performance. If that filter was in a separate class we can tweak on it all we want without affecting the basic decoder class.

As long as we're doing this we could make a new folder called `Decoder` and put the decoder, filter and headers in it and fix up the include in mainwindow for the new path to the Decoder class, we got a start towards cleaning up that source tree. Hate to pile a bunch of extra work on @punk-kaos not being a member of the team (but anybody who can code up a Kalman filter deserves to be).

Just a thought and question.

Chris-AC9KH:
> We'll see how ambitious I get. QT will compile for Android, but like you said I got stuck on trying to split the UI from the back end in a way that works for Android. I'll noodle on it a bit and see if its something I want to tackle. I was hoping originally that I could split off the frontend and have the backend be a relatively easy port, but when I realized the frontend and the backend are so tightly tied together I realized it was a lot bigger task than I expected it to be... Probably why nobody else has done it yet :P

Yes, it definitely is a hurdle. I'm looking at a port to iPadOS and I got the same problem. I can design the UI and lay it out with Swift in one afternoon. Now, I have to hook the UI to the back end :-)

It would require a complete port to Objective-C for the back end under a difference license. Apple's TOS for the App Store are not compatible with GPL due to the fact they don't distribute the app to everybody - only Mac or iOS users can get it there. The GPL claims that's a "restriction" on people's "rights" so Apple just flat rejects GPL licensed software hosted on the App Store.

Thinking about it, if some guys (and gals) would want to try an Android port a completely separate repository for it would better than a branch. Then your commit history, PR's, etc. is separate from the desktop version. Could make that happen too.

Chris-AC9KH:
Last night I only saw one instance where this may have done something. I sent a group SNR? query to see who was on and five stations responded. One of them decoded at -31. I sent the standard group QUERY MSGS and got MSG ID's from two of them, the extremely station was one of them. Still at -31. I queried the MSG ID on that one first, and now he came up to -24. But I got a 100% successful decode on the MSG with no dropped frames and it took about 5 minutes to transmit it. To me that's pretty impressive because at -24 it can sometimes take a couple tries on a long MSG.

My basic settings are to adjust audio drive with the radio with antennas disconnected so I'm down near 5dB. With antennas connected I adjust RF gain to show 25-30 dB on a clean bandpass with no signals on it. On 160 I run AGC on Fast with 0.1 second decay time (IC-7300).

Antenna system used was my 127' vertical for transmit, 920ft Beverage for receive, the antennas are switched with a relay downstream of the radio, the Beverage is pointed E/NE.

Running it with my FlexRadio today, but 30m is dead at present. Although about an hour ago I did decode a MSG that was sent on the same offset I'm using that came from a station in Canada. SNR says he was -25 and was using "F" mode, it was a pretty long MSG and no dropped frames.

Overall, I think I'm seeing better decode reliability with this with signals in the mid -20's range. The Flex does have a better receiver than the IC-7300 and there is no hardware interface or physical connection between the computer and the radio with the Flex. Which usually makes it an overall better system with less "built in" noise. The IC-7300 is probably a better test bed and it is the radio I use for digital modes most of the time. The Flex, due to its superior audio, is normally used for SSB phone.

So I'm continuing to run it and test it. Not disappointed and it definitely doesn't do anything bad.

jsherer:
Just curious, are any folks measuring the impact of features like this on older hardware, say like a Pi trying to multi decode?

Chris-AC9KH:
@Joe-K0OG yep, those -30 to -60 ones are going to be outside the usable range of the SNR calculations and that is duplicatable with your revised test setup to insert some extra "noise" into the bandpass. Even the low 30's with the stock software is outside the range of the calcs. SNR is a sampling of power in the signal fundamental relative to the RMS sum of the energy of the in-band noise components, excluding harmonics. When it shows anything less than -28 you can ignore that. It was a very weak signal and it attempted to calculate it.

I'm not seeing outside the usual on 30m today. Very limited activity on JS8, the dB meter is dead on 30 and not even moving. The FT8 guys are going strong a few KC up the waterfall (on my Flex). But I'll be going to 80 at 6:30 before we eat supper and that's where the more interesting things happen :-)

wmiler:
@punk-kaos I'm going to make a request that you document your code before any merge. See #4 for instructions. Thanks!
CI is building now as well.

punk-kaos:
Ok, classes include Doxygen hooks for basic descriptions etc.

Chris-AC9KH:
@punk-kaos BTW - if you're still eyeing that Android port, I created a new Android-port repository for it yesterday. There's some other guys interested in it too, and announcement made on groups.io, so you might have collaborators with time. All it will take is for somebody to get a rough start, and I think they'll come.

So if you get a slow day looking for something to do, let me know and that repo is yours. We'll create an Android Development team for it.

punk-kaos:
> @punk-kaos BTW - if you're still eyeing that Android port, I created a new Android-port repository for it yesterday. There's some other guys interested in it too, and announcement made on groups.io, so you might have collaborators with time. All it will take is for somebody to get a rough start, and I think they'll come.
> 
> So if you get a slow day looking for something to do, let me know and that repo is yours. We'll create an Android Development team for it.

I MIGHT have noticed that already, and I've been giving it the eyeball. Dunno yet if I'll tackle it but it sure would be nice to have a good Android port. I'll keep you posted :)

PR #86, opened 2025-12-06
Change name to JS8Call

Chris-AC9KH:
The one commit changes the program name AND the binary name to JS8Call instead of JS8Call-improved. Along with changing related build instructions and necessary support files. It does leave reference in certain places that the software now comes from the JS8Call-improved group.

Screenshot 2025-12-05 at 19 13 14

Screenshot 2025-12-05 at 19 14 04

The second commit is an attempt to fix the workflows for this. You'd better take a look at that one, @wmiler and see what I broke there :-)

### DIFF
--- a/.github/workflows/scripts/ci-windows-installer.iss
+++ b/.github/workflows/scripts/ci-windows-installer.iss
@@ -17,7 +17,7 @@ AppUpdatesURL={#MyAppURL}
 LicenseFile=D:\a\JS8Call-improved\JS8Call-improved\build\JS8Call\LICENSE
 DefaultDirName={autopf}\{#MyAppName}
 UsePreviousAppDir=no
-UninstallDisplayIcon={app}\{#MyAppExeName}
+; UninstallDisplayIcon={app}\{#MyAppExeName}
 ; "ArchitecturesAllowed=x64compatible" specifies that Setup cannot run
 ; on anything but x64 and Windows 11 on Arm.
 ArchitecturesAllowed=x64compatible
@@ -34,6 +34,7 @@ OutputDir=D:\a\JS8Call-improved\JS8Call-improved\build\JS8Call
 ; This can be changed from the ci-windows.yml for release builds
 OutputBaseFilename=JS8Call-installer
 SetupIconFile=D:\a\JS8Call-improved\JS8Call-improved\icons\windows-icons\js8call.ico
+UninstallDisplayIcon=D:\a\JS8Call-improved\JS8Call-improved\icons\windows-icons\js8call.ico
 SolidCompression=yes
 ; WizardStyle introduced in 6.6.0, Github has 6.5.4
 ;WizardStyle=modern dynamic


Chris-AC9KH:
Oh yeah. The CI build is broken on Windows. You might know :-)

Joe-K0OG:
I built the pr/name-change branch on Linux and Windows, and find no functional problems with it.  It's good to have "JS8Call" back again ;).

On Windows, with my Inno installer script, modified to account for the name change, the install/uninstall all works well.  Of course, we will need to test the modified script for the CI Windows build to make sure it works as expected.

73,
-Joe-
K0OG

Joe-K0OG:
I suggest that we set Hamlib 4.6.5 as our version for 2.5.0 release, since 4.6.5 is the latest stable release.

73,
-Joe-
K0OG

wmiler:
> I suggest that we set Hamlib 4.6.5 as our version for 2.5.0 release, since 4.6.5 is the latest stable release.
> 
> 73, -Joe- K0OG

This shouldn't be an issue. Do we have an eta for our next release?

Joe-K0OG:
> I would not recommend changing the Boost version. That is like changing glibc in linux.

I couldn't tell which version you used in the CI builds.  If you are at 1.88, just keep it there, like Chris said.  It ain't broke, so don't fix it ;)...

Chris-AC9KH:
@Joe-K0OG @jsherer I changed the About dialog per Joe's recommendations. How does this look?

Screenshot 2025-12-06 at 11 08 49

wmiler:
@Joe-K0OG Ok, fixed. The installer is included in the artifact zip. Thanks for pointing it out. :)

Joe-K0OG:
I posted this over in Discussions, and meant to post it here.  Sorry about that...

=====

I think the OutputBaseFilename does make a difference in the resulting file name. The way you have it, the output is:
D:\a\JS8Call-improved\JS8Call-improved\build\JS8Call-improved\JS8Call-installer-devel_X64.exe.

I have mine as:
OutputBaseFilename={#MyAppName}_{#MyAppVersion}

...and this results in an output file name of: JS8Call_2.5.0.exe. Isn't that what we want?

73,
-Joe-
K0OG

wmiler:
> > Attestations are turned on now when this PR gets pushed. This provides build provenance for the artifacts. Signing and what not will be provided in it's own separate PR. See [GitHub Attestations](https://docs.github.com/en/actions/concepts/security/artifact-attestations) for more info.
> 
> Do we want to run attestations on CI builds? Or only on CD?

Ideally only on CD, but I'm ok letting a few thru now just to make sure it's all working correctly. I want to check it thru to the merge, and I'll put in the block for CIs in a follow up PR.

Joe-K0OG:
I just ran a Windows CI build again, and it worked fine, but the resulting installer still does not show the JS8Call icon.  This is no biggie.  Looking at the Inno installer script, it looks like it is correctly pointing to the js8call.ico file, so I don't know why it's not picking up the icon.  This isn't worth holding anything up though.

Other than that tiny nit-pick, I think it's ready to go.  My tests on Linux, Win10 and Win11, as well as the Linux install scripts in the source tools directory all seem to be working well.

73,
-Joe-
K0OG

Chris-AC9KH:
> as well as the Linux install scripts in the source tools directory all seem to be working well.

I changed those linux scripts too, using a wildcard for the uninstaller, in the event somebody still has an installation named JS8Call-improved. Tested that on Debian 13 and it uninstalled my current 2.4.0 -improved install just fine.

Since they likely won't be included in the Assets of a 2.5.0 release I changed them to pull master. I've found them to be extremely useful for a quick test build on linux. And they're also useful for a end-user to do a quick test build for debugging, like I did with Tom K1TWH, vs trying to get an AppImage to him. He said his system is not real high-end and it only takes ~15 min from scratch to build and install a test build with the script.

OTOH, we could include them in the release Assets too, as we still might run into some odd systems (like arm64 Raspberry Pi) where the script will work, but an AppImage won't.

Joe-K0OG:
@wmiler Wyatt, that "maybe fix for icon issue" didn't seem to change anything as the installer still shows a generic icon when pulled out of the .zip file.  That's ok as far as I'm concerned - it's a super-minor issue, and maybe can be looked at later.  It's certainly not worth holding anything up.

73,
-Joe-
K0OG

PR #101, opened 2025-12-16
Split bandaid.

### LINKED ISSUE #94: JS8Call uses rig's split mode in a chaotic fashion.
**Describe the bug**

When configuring JS8Call to use rig split, it apparently will *always* use split mode, and which split frequency it uses depends not only on the audio frequency of the present transmission, but also on the audio frequency of the previous transmission.

**To Reproduce**

1. Use a rig that allows CAT. (I used one controlled by `rigctld` and had JS8Call talk to that.)
2. Configure JS8Call to use rig split mode (see screenshot)
3. Go 28078 kHz on the rig.
4. Select audio frequency offset between 2600 and 2700 Hz. Hit tune. The rig goes to 28079 kHz (So the actual audio used hopefully is between 1600 and 1700 Hz.)
5. Now, go to some frequency offset between 1500 and 1600 Hz and hit tune. The rig splits to 28079 kHz again. If the math is done correctly by JS8Call, the audio is now between 500 and 600 Hz. This defeats a purpose of split, namely, to keep the audio actually used above 1500 Hz so its harmonics are above 3000 and above 4500 Hz.
6. Now go to some audio frequency offset between 1100 and 1200 Hz. Hit "tune". The rig jumps to 28077.5 kHz, so the actual audio used is between 1600 and 1700 Hz.l
7. No, return to the 1500-1600 area. Hit tune. The rig again jumps to 28077.5 kHz. So this time, the audio actually used is between 2000 and 2100 Hz.

Image

**Expected behavior**

* The audio frequency actually used should always sit in the same 500 Hz slot. (Open for debate, I would use 1700-2200 Hz.)
* Which split frequency is used to transmit a certain desired radio frequency should only depend on that desired radio frequency, not on history of previously transmitted frequencies.
* When the audio frequency offset as requested by the user is already in the 500 Hz slot the program wants to use anyway, split functionality should not be used.

**Desktop (please complete the following information):**

 - Debian Trixie
 - I used a commit that had @rruchte's commit introduced at https://github.com/JS8Call-improved/JS8Call-improved/issues/16#issuecomment-3592122884 on top of the present main branch HEAD, namely 39d9a8ee07e550a9426938a6645424c0
 - IC-705
 - `rigctld` from Debian's `libhamlib-utils` at Version: 4.6.2-1+b1

**Additional context**

In some way, this is a follow-up of #16 .

aknrdureegaesr:
The underlying bug below #94 and probably #16 is a hamlib bug.

This works around that hamlib bug. So it fixes #94.

Also some documentation changes I did on the way to finding out what the problem was.

I also changed the sign of `m_XIT` in `mainwindow.cpp` so that it is positive in the common case of an audio frequency between 500 and 1500 Hz, instead of, as it used to be, negative. This should make the code
somewhat more comprehensible.

### DIFF
--- a/HamlibTransceiver.cpp
+++ b/HamlibTransceiver.cpp
@@ -746,6 +746,17 @@ void HamlibTransceiver::do_frequency (Frequency f, MODE m, bool no_ignore)
     }
 }
 
+/** Work around hamlib bug [#1966](https://github.com/Hamlib/Hamlib/issues/1966). */
+void HamlibTransceiver::hamlib_bug_bandaid(TransceiverState const&s) {
+    if (s.frequency() == s.tx_frequency() && s.split() && s.ptt()) {
+        // Change the frequency ever so slightly without telling anybody.
+        // Will not matter during the upcoming transmit and will be corrected
+        // with the next update when the receive comes in.
+        TRACE_CAT ("HamlibTransceiver", "Fiddling with rx frequency to work around a hamlib bug.");
+        error_check (rig_set_freq (rig_.data (), RIG_VFO_CURR, s.frequency() + 1), tr ("fiddling frequency as a hamlib bug workaround."));
+    }
+}
+
 void HamlibTransceiver::do_tx_frequency (Frequency tx, MODE mode, bool no_ignore)
 {
   TRACE_CAT ("HamlibTransceiver", tx << "reversed:" << reversed_);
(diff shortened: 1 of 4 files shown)


Chris-AC9KH:
@aknrdureegaesr not necessary to do two PR's. If happy with it, just merge it. The Hamlib one is plainly revert-able IF Hamlib ever fixes the issue and even then it likely won't hurt anything.

Chris-AC9KH:
I'll add that I don't have high hopes that Hamlib will fix this any time soon. Since Mike passed away there these sort of things just don't get looked at on a timely basis and they will probably get into 4.7 before it is ever addressed. Since we don't plan on switching away from 4.6.5 any time soon, this is the best that we have to fix it.

PR #105, opened 2025-12-18
Add JS8Call User Guide to Help menu

Chris-AC9KH:
This adds the JS8Call User Guide as a Help menu item. It will open your web browser and display the User Guide, fetched from where it's currently hosted on Google Docs.

@jsherer the source code (I suspect it's .docx?) for this document could be placed in the docs folder in the source tree so developers can make improvements to it as necessary, unless you want to maintain it yourself.

@mgochoa57 this could also be exported as PDF from the source code and hosted on the website as a PDF displayable in a web browser. If is desirable to add this to website, there is a copy of it in PDF format in the current Release assets.
https://github.com/JS8Call-improved/JS8Call-improved/releases/tag/release%2F2.4.0

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -2596,12 +2596,9 @@ void MainWindow::on_actionCheck_for_Updates_triggered()
     checkVersion(true);
 }
 
-// Chris WIP
-
 void MainWindow::on_actionUser_Guide_triggered()
 {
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
@jsherer markdown would be perfect. Then the URL can be changed to the github repo and it will display nicely in a web browser. If you want to add it to the docs directory just push a commit to this branch and then I'll update the URL to it when you get it done.

PR #112, opened 2025-12-31
General cleanup of unused and legacy files in source tree

Chris-AC9KH:
We are gathered here today in the presence of fellow developers to witness the un-joining of stuff we don't need from the source tree.

And maybe even begin cleaning the place up a little bit. Sort classes out into folders. Break mainwindow down into manageable classes. Make the source tree pretty again.

If anybody objects to removal of these legacy Fortran timer, WSJT-x/FT8, CMake, or otherwise dead or unused files from the source tree, speak now or forever hold your peace.


rruchte (approved):
LGTM

Chris-AC9KH:
I added a second commit to remove some old Fortran module files, a couple .bat files, and some Qt Creator artifacts that got left in the source tree at one point. I'll squash those two commits to make it a little cleaner.

There is two python scripts in the tree that probably don't need to be there and could be moved to the tools directory. I'll wait for some feedback from @jsherer to see where he wants to put those. But that can be done as part of starting to sort the classes out into folders.

We need to get the source tree so it's more understandable. There's some of the header files that have a .h extension, and others that are .hpp. Those should be standardized to use one or the other. I prefer just .h since that is used for both C and C++ and it's just a matter of personal preference as to which you want to use. In this case we don't have any C code so either could be used.

A general cleanup and organizing of the tree is going to touch just about every file. But it needs to be done because at present it's like a bunch of animals got loose and started scattering files all over the place. I realize that most of this came from WSJT-x, but still. Do it one step at a time and get everything sorted out. Then we can tackle that mainwindow monster before it gets bigger and grows two heads.

wmiler:
We need to check the files in docker/ and see if they still work.

wmiler (approved):

Chris-AC9KH:
Yeah, I don't know what's different with docker, but after the code was ported to 6.9 Joe reported it won't build anymore with Qt6.4. Qt6.4 is no longer supported by the Qt Group. Seems to me they had a 6.5.3 LTS release, they skipped over 6.6, had a 6.7 LTS, skipped over 6.8 and 6.9 is their latest LTS.

This is not a problem at all with .AppImage, which is the only thing we can officially support on linux. I use the linux build scripts to build a test version on linux, takes less than 10 minutes and doesn't rely on distro-packaged Qt. The official Qt6.9.3 libs for linux have FFmpeg audio, wayland support, and scalable vector graphics so Qt elements display correctly on the more modern linux desktops.

PR #114, opened 2026-01-01
Code cleanup - move files to JS8_Mode and JS8_Transceiver class groups

Chris-AC9KH:
Step 2 of code cleanup.
- This moves loose files in the source tree to JS_Mode and JS8_Transceiver class groups
- The convention is that class groups should be prefixed with `JS8_` to organize them in the directory structure
- Optional renaming of classes to match their class group can be done later
- This also moves the dual-use of `.hpp` and `.h` header file extensions to standardize on `.h` but only affects the moved the files so far

I figured this is an invasive-enough change for step 2.


punk-kaos (approved):
Looks pretty straightforward... Sorta iffy on the idea of leaving part of the files hpp and part h but I guess its not really any worse than it was before. And probably safer to do in smaller chunks that one huge forklift move.

Chris-AC9KH:
I'm working on step 3 right now, which is going to be a lot bigger than step 2. Step 3 gets rid of the rest of the .hpp's and the rest of the "loose" files in the source tree.

Yes, there was a hodge-podge of .h and .hpp files. I've always preferred .h even for C++ because you tend to look at the end of the file (the extension) to pick out the headers from the source files, and it's more immediately clear if .h is used.

I see no reason to separate source files into a `src` folder, and headers into an `include` folder for this project. It's not that big. Moving headers to an include folder involves more paths. If a function is added, it's declared in the header anyway, and doing it this way keeps the two together.

wmiler (approved):
+1 for switching to just plain .h

Chris-AC9KH:
@wmiler I did not move those .desktop files, did not want to break your workflow. They can be moved to JS8_Include when you get time up update your workflow path(s).

wmiler:
> @wmiler I did not move those .desktop files, did not want to break your workflow. They can be moved to JS8_Include when you get time up update your workflow path(s).

What's message_aggregator.desktop? That's not used in any of the workflows to my knowledge.
Looks like it was last touched 9 years ago, is this something from the original creation @jsherer ?

Chris-AC9KH:
I'm not sure what message_aggregator.desktop is. It doesn't appear to be used anywhere. So if it is a dead file, or unless somebody knows where it came from, it can be removed.

Chris-AC9KH:
Xcode commit history shows  message_aggregator.desktop appeared in the source tree back in 2018 with Jordan's initial commit and hasn't been changed since. So it probably came from wsjt-x and it's a dead file.

PR #118, opened 2026-01-04
Mainwindow Breakdown

Chris-AC9KH:
So everybody knows mainwindow.cpp is a monster. It is all one huge MainWindow class. Which in itself is not a problem. You can create a whole bunch of new classes and break it down that way. But the design muddy's this up. It is a monolithic design, the engine that powers everything is in mainwindow.cpp and it also powers the UI.

So what I'm doing here is moving the includes in the source to the class header. Those should've never been there in the first place, except for a couple that are necessary. Developers see a long list of includes in the source file, oh - let's just bolt on another one (or two, or three).

So then I moved three member functions of the MainWindow class to separate files placed in JS8_Mainwindow, all of these use the same class header so they are just source files for the function implementation. So then you have CMake tell clang to compile the object for these function source files, and the linker knows how to link that object with a one-line function() call in mainwindow.cpp.

So this got rid of 550 lines of code out of mainwindow and put it into separate files. So now if you want to work on the software updater, you can find the code for it in the file that's the same name as the function, without groping around in mainwindow.cpp to find it.

This all works fine on MacOS 26 with clang 17. Submitting the PR here to see how the CI build system handles it.


Joe-K0OG (approved):
@Chris-AC9KH Chris, excellent work! I built https://github.com/Chris-AC9KH/JS8Call-improved.git the pr/mainwindow branch on Linux Mint 22.2 and Windows 11 (JTSDK64-Tools + Inno Installer), and all appears to work perfectly on both platforms.

One thing I noticed is this: Before, I would often need to press SHIFT-F4 twice to get the waterfall controls panel to display or close. Now it works on a single SHIFT-F4 press. Something "got" fixed I guess. I never reported it before because it was such a minor issue I forgot about it. This improvement was apparent on both Windows and Linux.

73,
-Joe-
K0OG

wmiler (approved):

wmiler:
I'll give it a go here in a few minutes. Are keyboard shortcuts actually mention in the docs anywhere? 🙀 I didn't even know about shift-F4 🤦‍♂️

Joe-K0OG:
> Are keyboard shortcuts actually mention in the docs anywhere? 🙀 I didn't even know about shift-F4 🤦‍♂️

I don't think they are mentioned in the documentation, but they are not hidden in any way.  When you click on any of the menu items, the keyboard shortcuts for several functions are shown.  Click on View and you will see that "Show Waterfall Controls" is associated with "Shift+F4".

73,
-Joe-
K0OG

wmiler:
I'm going to withdraw the documentation tag for this PR, because it has broken with all the files getting moved around. I'll have a related docs PR when I get it all fixed up.

Question: Would it be beneficial from the code documentation side of things, to have a previous version available? Ie 2.5.0 and master.

Chris-AC9KH:
I'll merge this to prevent broken PR's. I don't really think a code documentation run from earlier versions is going to do much good? Developers will want to look at the current version to get an idea of the class structure and execution tree. 2.5.0 and master are very similar except master will have mainwindow partially broken down.

I wanted to tackle this over the New Year when PR activity was at a minimum. There's still some things that can be done to break down mainwindow further. But let's do it one member function at a time as work gets done to the affected member function that's still in mainwindow.cpp.

This was a big one stripping out over 4,000 lines of code from mainwindow. But I wouldn't want to do it on a regular basis so as to prevent disruption of normal development in the future.

PR #124, opened 2026-01-08
Add documentation to JS8_Logbook files

wmiler:
Added a bit more to the initial docs, mostly to check how it looks. No C++ code changed.


wmiler:
The review request is specifically for regarding what's including in the top block. I set the `@version` and the `@author` and the `@date` which should be considered the "modified date". Not sure if any of those should be included.

```
/**
 * @file adif.cpp
 * @brief Implementation of the ADIF class for handling ADIF log files.
 * @version 2.5.0
 * @author JS8Call Development Team
 * @date 2026-01-08
 */
```
 
It looks pretty good.

LogBookClassReference

PR #128, opened 2026-01-10
User Manual as markdown file, other documentation changes

wmiler:
The big item here is Jordan's User Manual as a markdown file. It most likely will need further layout out refinement.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -25,7 +25,15 @@ jobs:
       - name: Get doxygen and graphviz
         run: |
           sudo apt-get update
-          sudo apt-get install -y doxygen graphviz
+          sudo apt-get install -y doxygen graphviz flex bison make cmake g++
+          wget https://www.doxygen.nl/files/doxygen-1.16.0.src.tar.gz
+          tar zxf doxygen-1.16.0.src.tar.gz
+          cd doxygen-1.16.0
+          mkdir build
+          cd build
+          cmake -G "Unix Makefiles" ..
+          make
+          sudo make install
 
       - name: Build docs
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
Also @aknrdureegaesr mentioned something that he wanted to do with that User Guide for GitHub pages. I believe he mentioned that in a group email the other day.

In any case the .docx needs work to bring it up to 2.5.0. Then export it to PDF for the website download. I was hoping @jsherer might have time to do that now that is easily available for editing. I did modify it a little bit in a couple spots, in particular the one section of it that explains directed commands. But I haven't had time to do a complete update to ver 2.5.0 yet and include all the new stuff for group messaging, etc..

Chris-AC9KH:
So how do we get it converted to PDF for a downloadable version?

Joe-K0OG:
> But for whatever reason Mac LibreOffice has that font, linux LibreOffice does not have it and it tries to replace it with Liberation serif, which doesn't work and none of the pages line up.

@Chris-AC9KH My Linux Mint 22.2 system does have the Microsoft Sans Serif font, and it's accessible in LibreOffice.  I'm fairly certain I installed it at some time in the distant past.  It can be downloaded from several places, for instance:  https://font.download/font/microsoft-sans-serif

73,
-Joe-
K0OG

Chris-AC9KH:
Yeah, it must be something you have to install extra on the system. It looks like we're going to go with markdown instead for editing the document. And figure out how to convert it to PDF for downloads instead of exporting .docx from Office.

Chris-AC9KH:
@wmiler no need to waste too much time on PDF. That can be done easily by loading the web page into a web browser and exporting it as PDF. I did it two ways - one with the Table of Contents, then edited the Table of Contents out. I honestly don't think the ToC is needed for the PDF version.

Issue #103, opened 2025-12-17
Dark mode needs a cleanup regarding the text entry fields.


aknrdureegaesr:
**Describe the bug**

Some friend without Github account tells me: With darkmode configured on his Win 11 machine, 2.4.0 shows white text on white background.

<img width="551" height="217" alt="Image" src="https://github.com/user-attachments/assets/b2b163bf-cb82-4f11-bb0a-a9c8484140b5" />

Tried to verify myself on Debian Linux Trixie. Got to see black writing on white background in the text fields. Should be the other way around in dark mode.

**To Reproduce**

1. Configure dark mode
2. Start JS8Call
3. Check text entry input fields.

**Expected behavior**

The usual dark mode.

**Screenshots**

See above.

**Desktop (please complete the following information):**
  
 - Windows 11
 - 2.4.0

Chris-AC9KH:
Cleaning up some of these dead issues after the 2.5.0 release that will never be resolved since it is not an issue in our code. This one is a platform-specific graphics rendering issue that we can do nothing about in our code. This would need to be addressed by your platform vendor.

Chris-AC9KH:
@rruchte ok, then it will be a platform-specific patch. JS8Call does not have a built-in selection for light or dark mode that I know about. It simply uses your system theming. So I would suspect something like this would require platform-specific code, and then is it the same between Gnome and KDE, or other desktop variants on linux? You can get the same thing to happen by trying to apply themes on linux with applications like LibreOffice.

rruchte:
@rubberduck203 My understanding is that a long time ago that was an acceptable way to do things, but no longer. I think we can be a lot smarter about when the style gets applied, and do it in a way that respects the theme.

rruchte:
Should I add warning icons if the fields are empty?

<img width="807" height="286" alt="Image" src="https://github.com/user-attachments/assets/19025a3b-144a-43f8-b8c8-e0dd5f32775d" />

Chris-AC9KH:
Fixed by https://github.com/JS8Call-improved/JS8Call-improved/pull/133

PR #133, opened 2026-01-13
Addresses issue #103, dark mode cleanup

rruchte:
- Removes explicit background color specifications
- Adds alert icons to the callsign and grid, entry fields when they are empty
- Removes custom styles from saved messages list

### DIFF
--- a/JS8_UI/Configuration.cpp
+++ b/JS8_UI/Configuration.cpp
@@ -393,6 +393,9 @@ class Configuration::impl final : public QDialog {
     void load_audio_devices(QAudioDevice::Mode, QComboBox *, QAudioDevice *);
     void update_audio_channels(QComboBox const *, QComboBox const *, int, bool);
     void load_network_interfaces(CheckableItemComboBox *, QStringList current);
+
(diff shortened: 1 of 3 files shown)


Chris-AC9KH:
@wmiler I created a release/2.5.1 branch with this fix cherry-picked into it. And have a draft release for it. I'll get MacOS builds going right away and get those sent to Apple. As soon as the run completes on that branch let me know where the Windows and Linux builds are and I'll get those uploaded. If the MacOS ones take longer than I expect it is no big deal and I'll put them there later. Apple's notarization team SHOULD be fully staffed after the holidays so I expect it to go pretty fast.

Chris-AC9KH:
If I actually knew how to sign it from Windows. Which I don't. I fuddled my thru making a certificate and key but not sure how to use it :-)

The release/2.5.0 build is signed. I uploaded that to the virus total and it said it was signed by the JS8Call Developers.

I just checked the arm64 build running in the VM on the Mac with Debian 13. That works fine.

Screenshot 2026-01-13 at 21 28 51

Chris-AC9KH:
The installer works, Windows Defender had no issues with it. The program runs fine. So I'll publish that patch release. I just put minimal release notes in for it, explaining the reason for the patch release. But it will create a release/2.5.1 tag on publish.

PR #134, opened 2026-01-14
Inbound APRS message relay into JS8 mesh.

punk-kaos:
Adds support for inbound relay from APRS. Listens for message packets in APRS for callsigns in our heard list. If heard it relays them to the host under the existing @APRSIS format. 

Example: KI7WKZ: @APRSIS MSG TO: N0CALL THIS IS A MESSAGE FROM APRS! DE APRSCALL-6.

This will be placed in the users message inbox as coming from APRS. 

This allows for bidirectional APRS messaging across HF Via JS8 to augment our current unidirectional outbound APRS features. 

USE CASES: I hike and camp a lot in remote areas. I often relay APRS messages via NVIS JS8 message out of a deep canyon where VHF wouldn't reach to SMS to reach home, this would allow me to get return messages or to contact WXBOT for weather reports for example.


Chris-AC9KH:
Kelly, this looks good. As we work on mainwindow.cpp let's try to continue to move affected MainWindow class member functions out of that file if we can. I would say anything reasonable, especially over 50 lines like processTxQueue() could be stripped out of that file. I didn't get 'em all over New Year's. I just stopped at my goal of stripping at least 4,000 lines of code out of that file. At that point we decided that was invasive enough for one sitting. Then as we work on stuff in it, continue that process more in a one-step-at-a-time approach :-)

punk-kaos:
> Kelly, this looks good. As we work on mainwindow.cpp let's try to continue to move affected MainWindow class member functions out of that file if we can. I would say anything reasonable, especially over 50 lines like processTxQueue() could be stripped out of that file. I didn't get 'em all over New Year's. I just stopped at my goal of stripping at least 4,000 lines of code out of that file. At that point we decided that was invasive enough for one sitting. Then as we work on stuff in it, continue that process more in a one-step-at-a-time approach :-)

Noted... I'll confess I actually wrote this pre-refactor and just massaged it into the new format after the fact. I probably should have paid a LITTLE more attention to the changes in mainwindow.

punk-kaos:
> @punk-kaos excellent! I may end up stealing your work there! :-)

Beauty of collaboration, its there for the taking ;)

punk-kaos:
> Please add some doxy blocks to the functions in APRSISClient.cpp that you added. Feel to fixup any docs that are already there. Most of the blocks I added were for testing, so prolly aren't the best, lol!
> 
> Otherwise, looks fine. Maybe also a doxy block about how to actually use it.

Whoops, that was on my to-do! I'll get that done!

punk-kaos:
Ok, split out functions better and added some DOXYGEN docs.

punk-kaos:
> @punk-kaos other than a couple minor changes I'm requesting, this looks good!

Nothing too hard there, I'll get on it.

PR #136, opened 2026-01-14
Windows signing

wmiler:


wmiler:
@Chris-AC9KH Chris, this should now work correctly for regular CI builds, and as a manually run "Release" build, and can be run from a branch. **Please note** When running a "release" build, **only** JS8Call Admins can approve it to run (You, me, or Jordan currently), if you need me to change that to a different group, just let me know.

Github will only put out zipped artifacts, so still stuck with that.

The "release" signing key is the same as the CI key for now, but is easy to change when we get that far.

Chris-AC9KH:
@wmiler sounds good. The zipped artifacts aren't any big deal. I just unpack 'em and upload the AppImages and .exe installer. Works fine.

Issue #138, opened 2026-01-15
Stablilize code formatting


aknrdureegaesr:
Some time ago, we agreed to let `clang-format-18` format our C++ code.

By now, not all code is still formatted this way.

So, the next step in this endeavor:

* Re-format our code one more time.
* Add a check to our automated builds, so code that is not formatted in that way will be flagged at PR level.
* Add a stance to our `CONTRIBUTING.md` file that explains how to format contributions.

Chris-AC9KH:
I don't actually recall "agreeing" to any such thing. It was more like an idea that was abandoned and the agreement was that it would be looked at on a per-PR basis.

We're not even close to stripping down mainwindow yet. Over New Year's when nobody was around I just stopped at my goal of stripping at least 4,000 lines of code out of that file. But while doing it I examined the code with a fine tooth comb and fixed improper declaration of member overrides, removed unused variables, etc.. The code is now compiled differently with class member functions compiled as separate objects, linked by the linker, which fixed some outstanding issues with UI functionality due to flawed execution order in mainwindow.

So while that one was big, it actually fixed things. This doesn't fix anything.

Chris-AC9KH:
I'll add that you can put **recommendations** in the developer guide, but leave it at that. Not all developers use the same tools, and if you try to push that it's going to go over the proverbial lead balloon.

Chris-AC9KH:
It's kind of a difficult situation. I can generate any number of different clang-format configurations with Xcode. Here's an example of a configuration file using the LLVM standard (on the left) vs the WebKit kit one that was used in the PR. The LLVM one generates more readable code, and that's why it's used as a standard by almost all the Fortune 500 companies.

<img width="1554" height="1027" alt="Image" src="https://github.com/user-attachments/assets/2a563bc6-3aa5-4ee1-a1af-00bb60b9ed02" />

But I'll bet you can run this on linux (the community version released by Apple) vs running it on MacOS or Windows (the proprietary versions) with Visual Studio, and you won't get the same results. As you can see in the above there is considerable differences between LLVM and the WebKit one robbed off the internet.

Chris-AC9KH:
Wow. I generated a WebKit config file (left) vs the one that's in the PR (right). The one in the PR has been messed with big time. It's not a standard at all. It's somebody's idea of a new coding style that has never been seen before. No wonder it don't work!

[...]

PR #139, opened 2026-01-15
Restore the .clang-format file and explain how to use it.

aknrdureegaesr:
With 4a7a955f4f5a1a53, the `.clang-format` file was removed, I presume, accidentally.

As part of the effort for #138 , this restores that file and also explains how to use it.


aknrdureegaesr:
As I found out the hard way and @Chris-AC9KH also mentioned in https://github.com/JS8Call-improved/JS8Call-improved/pull/141#issuecomment-3757751005 :

> clang-format will sometimes reformat files fed to it even if the file hasn't been edited

So the recommendation to run `clang-format-18` on the entire code base other than check-only mode needs to be removed.

We should do this only once, to establish a baseline that's clean, as far as coding style is concerned.

aknrdureegaesr:
> Add .clang-format to .gitignore so linux developers can copy it to the root of the file tree and then call it with:
clang-format -i -style=file JS8_xxx/filename.cpp (or whatever).

Would the existence of that file in the root directory hurt Windows or Mac people?

aknrdureegaesr:
> But you can try it.

Ok, in the process. First result: The Debian Trixie `clang-format-18` doesn't like that file. To see what the options are, my PC presently compiles `clang-format` from source, their branch `origin/main`. (I like that branch name.)

Chris-AC9KH:
> Would the existence of that file in the root directory hurt Windows or Mac people?

The file doesn't need to be in the root of the source tree or you're gonna format every file in the system with it. I'm trying to figure out why you think you need a .clang-format file in the first place? If using clang-format at the MacOS command line you use:

`clang-format -i --style="{BasedOnStyle: LLVM, IndentWidth: 4}" <filename.cpp>`

That command line command is the same one Xcode uses in the IDE. You don't use any .clang-format file. If I want the indenting to be 2 space instead of 4, I just click on that on the Xcode configuration like so:

Screenshot 2026-01-19 at 19 46 05

So if you want clang-format to use 2 space indenting at the command line you use:

`clang-format -i --style="{BasedOnStyle: LLVM, IndentWidth: 2}" <filename.cpp>`

If I'm working with any of the MacOS WebKit code then Xcode detects that automatically and switches to WebKit. But at the command line I would use:

`clang-format -i --style="{BasedOnStyle: WebKit, IndentWidth: 2}" <filename.cpp>`

I think you get the idea. But you don't need any sort of .clang-format file unless you are defining some custom version of one of the code formatting standards. So like I said, put it in the tools directory as a reference or tool if somebody doesn't know how to use clang-format. But otherwise I been using it for 19 years and have never used a .clang-format file yet.

Chris-AC9KH:
@aknrdureegaesr  I also fail to understand why you are using clang-format 18. clang-format 21 is in Xcode 26.0.1 and that will be updated to clang-format 23 at the next Xcode point release. clang-format 18 is seriously out-of-date and does not support some of the newer C++ standards that have come out since it was released.

Chris-AC9KH:
I used four-space indent when I did the recent code overhaul.

But anyway, that goes to show that clang-format is not the same between the linux versions and what MacOS and Windows uses. You can run clang-format --dump-config with your desired style and generate a .clang-format file, which is what I did with clang-format 21 in Xcode 26. If that file doesn't work with linux, then linux clang and MacOS/Window clang are not compatible. Which makes sense because the clang compiler won't build JS8Call on linux, but it does on both MacOS and Windows.

wmiler:
Closing this PR as wontfix.

PR #141, opened 2026-01-15
Source format via clang-format-18

aknrdureegaesr:
This also has the .clang-format file restored, as does https://github.com/JS8Call-improved/JS8Call-improved/pull/139 , and it contains all the changes that `clang-format-18` does to our code base.

There is nothing else in this PR. So it can be ditched and recreated easily.


Chris-AC9KH:
@wmiler what do you make of this? This type of thing was expressly voted down before because it just reformats code to a different format that is non-LLVM standard, touches way too many files for commit history sake, and doesn't provide any functional improvement.

From a functional standpoint, I only removed around 4,000 lines of code from mainwindow.cpp. That one is not done yet. But at the time it was done over New Year's it was deemed to be a good time to disrupt general development and PR's in order to get it started.

But the UI still needs to be separated from the back end "engine" of JS8Call. That is going to be best accomplished by designing a new UI with Qt Creator. Then hook up the UI to the back end and strip the current UI code out of the JS8 engine.

This is going to be necessary for me to create an iOS app for JS8. The iOS app won't use the Qt UI - it will be coded with SwiftUI. The back end engine will be ported to Objective-C, not using any of the current source code, and licensed under a proprietary less-restrictive license than GPLv3 so it can be distributed on the App Store.

So is this something we want do (again), or continue with the original plan of stripping down mainwindow.cpp one step at a time to re-design JS8Call as a non-monolithic application?

Chris-AC9KH:
You're just using WebKit instead of the LLVM standard. Which, I don't know why anybody would want to use WebKit for C++. It is used for HTML/CSS in Safari and the Apple WebKit Core classes. Either LLVM or Microsoft is used everywhere else.

We'll see what Wyatt thinks of it.

Chris-AC9KH:
https://github.com/JS8Call-improved/JS8Call-improved/pull/134 is not a final PR. We're using that one for one more step towards stripping more out of mainwindow.cpp. After the initial strip-down, we agreed that any further work on mainwindow would continue breaking it down one step at a time under it's finally under control.

Chris-AC9KH:
The key word there is "enforce". We are doing no such thing, i.e we are not going to become the Code Formatting Police. Nor are we going to keep re-formatting code, applied to the entire codebase, just because somebody came up with something different. It has already been done except for the checker files which I didn't do and there's only like 6 of them.

It says right in the commit comments that the codebase was reformatted with LLVM during the refactor of the tree:
https://github.com/JS8Call-improved/JS8Call-improved/commit/356bdc0f503342efbbcc1996d10e709f15b50c2f

So now we're going to do it (again) with WebKit? Ok. Whatever. This is a small, open-source volunteer project. It is not Apple's Safari/WebKit Core development team, which has about 120 people on it.

wmiler:
Yeah. I just ran it locally with the above (swift), and it breaks the doxy blocks. So I'm going to have to go with a **NO** at this point.

Chris-AC9KH:
Actually, there's no sense to running your clang-format experiment yet. Since I got roped into putting doxy hooks in all the source files, I found there has been some doxy hooks put in that broke some code formatting. I didn't have #include sorting turned on in Xcode the last time I formatted the source. So now since I have to check every file anyway, I turned on #include sorting.

I don't like #include sorting because sometimes it breaks things.

aknrdureegaesr:
Ok, let's step back and think about this some more.

I want this team to unite behind a common code style, document that code style, port our entire code base to that code style, and to have something in place that keeps our code that way.

I had hoped that `clang-format` would be a tool with which we can document our code style (simply by providing options to it), and that we can use to port our code to that code style, and keep it that way. Actually, I had hoped, once we have our code in a common code style, re-running `clang-format` would not change it. Unfortunately, `clang-format` isn't mature enough to do that: It is not idempotent. That's a bummer.

I still hope that `clang-format` is still mature enough to at least **check**, i.e., using `--dry-run` mode.

I thought the project had agreed on using `clang-format` when we merged the `.clang-format` file to our main branch back in November.  If someone wants to propose another program to be used instead, that can be run with consistent outcome under all three development platforms (Windows, Mac, Linux), I'd like to see that.  If we allow the use of Docker for that, `clang-format` can be run on these platforms nicely.

aknrdureegaesr:
We have two PRs as this one can be ditched easily if new knowledge comes up. There is nothing in here really that represents any work worth keeping, with the notable exception of this discussion.

PR #145, opened 2026-01-17
Rename class files to match their class declarations in JS_Mode

Chris-AC9KH:
This is a continued cleanup of the codebase. It renames four class files to match their class declarations. When code documentation is referred to, this helps with maintainability of the code by making it easier to identify the files that contain the class.

Trying to be a little less invasive now with each step of the code cleanup process.


Chris-AC9KH:
Oh, and yes, Xcode clang-format removed some whitespace from the doxy hooks in DecodedText

PR #147, opened 2026-01-17
Linux build now signs the AppImage, other updates

wmiler:
1. Switched to the newer `linuxdeploy` tool with qt plugin
2. Greatly simplified building the AppDir that becomes the AppImage
3. AppImage is now signed. Use ``JS8Call-devel-x86_64.AppImage --appimage-signature`` to look at the signature
4. appstream metadata now included


wmiler:
@Joe-K0OG Ok, this should work now :) Give it a whirl!

PR #150, opened 2026-01-18
Source tree and UI fixes

Chris-AC9KH:
I think this gets the code tree a little closer to where I like it. I think I mentioned it would be big. But in the end I stripped roughly 8,000 lines of code from mainwindow.cpp. We had too many things named MainWindow so that class is gone and replaced with UI_Constructor class. So the only thing left named that is mainwindow.cpp and its header, which is what drives the UI, along with the UI_Constructor member functions.

@Joe-K0OG had requested to have the APRS inbound relays turned off by default. So I did that.

And I fixed the layout of the General->Networking & Autoreply tab in the Settings. This one has made me disgusted for a very long time about how it's layed out, so while I was thinking about it I fixed it. This is what it used to do:

Screenshot 2026-01-17 at 21 49 54

Now if you stretch the Settings box out the entry fields expand with it so you can see everybody you got in your HB block list that sends HB's every 30 minutes. 

Screenshot 2026-01-17 at 21 52 43

I been running the build from this branch since late this afternoon (Jan 17), it runs good.

### DIFF
--- a/docs/user_guide/JS8Call_User_Guide.md
+++ b/docs/user_guide/JS8Call_User_Guide.md
@@ -1720,5 +1720,5 @@ future with the foundation laid, including:
 ### (s,c)-Dense Code:
 
 A whitepaper article is being written on this topic. In the meantime,
-see JSC.h, JSC.cpp, & JSC_map.cpp in the source repository for the
+see jsc.h, jsc.cpp, & jsc_map.cpp in the source repository for the
 complete dense code table.


JS8_Logbook/ADIF.h
wmiler: Add a @file and @brief at the top.

JS8_Main/TransmitTextEdit.cpp
wmiler: This one too @file and @brief

wmiler (approved):

Chris-AC9KH:
@Joe-K0OG don't even both with this until I get done with it. It's in a broken state and I'm force-pushing commits as I work on it. It will be in a broken state for a bit yet.

Issue #119, opened 2026-01-05
Features and bugs


karlanmitchell:
Wanted to just create a list of some issues I've seen in JS8 which I think would be relatively easy to fix/implement and isn't some fantastical wishlist.  Surely, many of these issues have been mentioned, but hopefully there is at least one undocumented one in here.

*Clicking HB will send to group if selected to the right.
*No auto reply is somewhat random, if any msg detected in waterfall maybe. (can we just block that offset + mode bandwidth?)
*Is the offset change broken in JSON API for TX? 
*Can't find good documentation of the API and it's behavior (I'll write this if you'd like, just want some input)
*HB timer resets if auto heartbeat reply is set (effectively it never sends unless band is dead)
*Messages in activity window not showing up in conversation window (that obvious should be there)
*Doubling/splitting of messages in conversation window (Sometime the splitting corrects when message is complete)
*The frequency scheduler seems a bit unintuitive.  While I don't want to call it broken, is it?

Features which would be nice:
*Adding escape character to modify decoding display for lower case (example HTTP://URL.COM/^AB^CD^^ prints to screen as HTTP://URL.COM/aBcD^)
*APRS Buttons in program
*Now the harder one, a temp style graph that shows activity last X min/hour for clear slots (even a simple implementation that adds a +1 to predefined slots/frame, say 50hz, and clears every 10 mins would be useful to avoid a large timetable of values)
*Area to enter custom CAT commands before/after PTT, on program start/shutdown.
*MSG outbox expiration date

I do have about 20 years of hobby C programming (embedded, server side, sockets), and can contribute, but time constraints don't really allow me to dive deep into anything non-work related anymore.  Like I mentioned, I'd be happy to document the API, and maybe a lot of heavy lifting or features can be implemented with 3rd party programs (JS8spotter comes to mind).

-Karlan AI5TS

wmiler:
See #87 the Great API rework. Documentation is expected as a part of that work. We are taking PRs on this now.

The freq scheduler works as intended, although not exactly as one would expect. The "end time" doesn't go back to the previous freq at the specific time. You have to add another schedule to take you back to your previous freq.

On the original js8call repo, we did have a fixed api PR, not sure if @jfrancis42 will be pushing his PR here soon.

karlanmitchell:
[...]

PR #151, opened 2026-01-19
Remove deprecated DisplayManual class

Chris-AC9KH:
Move OmniRig to deprecated group, scheduled for deprecation due to being no longer supported by Windows 11 and Qt6.9.3 This references https://github.com/JS8Call-improved/JS8Call-improved/issues/119 where the author of OmniRig clearly does not understand that his code must be ported to Qt 6.9.3 or newer. Qt does not keep deprecated stuff around to support an obsolete 32-bit program.

So OmniRig, as stated in CMakeLists.txt where the option is defined has one foot in the grave, the other on a banana peel. It is scheduled for deprecation from the codebase and is moved to that group.


Chris-AC9KH:
Yes, good idea. I'll do that.

Issue #153, opened 2026-01-19
JS8CallUltra mode


wmiler:
**Describe the bug**
JS8CallUltra mode is defined in the code but not selectable from the UI.

**Expected behavior**
To be able to select it.

**Desktop (please complete the following information):**
 - OS: Ubuntu 24.04
 - Version 2.6.0

wmiler:
Apparently you can set it via the API....

punk-kaos:
Nothing wrong with it that I can see... Its all there, just disabled. I'll pop a pull request in to enable it.

PR #155, opened 2026-01-20
Enable hidden ULTRA mode.

punk-kaos:
This enables the already existing but disabled ULTRA mode. Its a 4 second mode a step faster than turbo.

### DIFF
--- a/JS8_Include/commons.h
+++ b/JS8_Include/commons.h
@@ -24,7 +24,7 @@
 #define JS8_ENABLE_JS8B    1
 #define JS8_ENABLE_JS8C    1
 #define JS8_ENABLE_JS8E    1
-#define JS8_ENABLE_JS8I    0
+#define JS8_ENABLE_JS8I    1
 
 #define JS8A_SYMBOL_SAMPLES 1920
 #define JS8A_TX_SECONDS     15


punk-kaos:
In reference to issue #153

wmiler:
Review request is to make sure there isn't any issue with turning Ultra mode on. The code in PR is just changing the define.

aknrdureegaesr:
I'm hunting various little nasty problems with timing. We had a fairly big one that caused PTT to be pressed too late. There is a preference in the code what to do when that happens: Not send the entire sequence too late, but instead start right in the middle, where we should be if we had started in time. Which is quite how it should be done, imho. But starting late than leads to less "TX delay" than configured. And particularly in turbo mode, where timing is especially tight, it can even lead to dropped initial symbols, that are never sent. It is then up to our error correction to save the day, or else "`…`" will appear.

I have investigated these things after logging became available. I thought I was mostly done with preventing dropped symbols; still need to submit a PR for #63 . The problem that JS8Call will sometimes ignore the configured TX delay and put audio on immediately after PTT is still open.

Ultra mode is more time-critical. So when we activate that, we should double check PTT and signal timing issues.

Disclaimer: I don't know whether this was the reason Ultra was never switched on. (Wild speculation: The previous bugs could have eaten so many symbols that Ultra never really worked? But that could be utterly wrong.)

wmiler:
@aknrdureegaesr see https://github.com/JS8Call-improved/JS8Call-improved/issues/153#issuecomment-3774579792

PR #172, opened 2026-01-25
Give directed commands higher priority over messaging

Chris-AC9KH:
On groups.io it is claimed people are not getting their regularly scheduled SNR reports on time. JS8Call is primarily an automatic beaconing and signal reporting system, messaging is secondary. For weak-signal messaging it is probably advisable to switch to Olivia 8/250 with Reed Solomon ID anyway since it is faster and much more reliable than JS8, and doesn't rely on frame timing.

So this changes JS8Call to full Automatic Animal Mode so it will break MSG'ing with directed commands, which is what people want because they don't care about MSG'ing - uncontrolled automatic operation is the main MO. But it retains suppressing HB ACK for an incoming MSG.

And changes the description of the program to be more accurate.

I would suggest cherry-picking this to current release and push out a 2.5.2 patch release, since people are upset about not getting their (RSSR's) Regularly Scheduled Signal Reports.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,8 +67,7 @@ set(PROJECT_COPYRIGHT "Copyright (C) 2001-2018 by Joe Taylor, K1JT, (C) 2018 by 
 set(PROJECT_HOMEPAGE https://groups.io/g/js8call)
 set(PROJECT_SUMMARY_DESCRIPTION "${PROJECT_NAME} - Digital Modes for Weak Signal Communications in Amateur Radio.")
 set(PROJECT_DESCRIPTION "${PROJECT_SUMMARY_DESCRIPTION}
- ${PROJECT_NAME} is a computer program designed to facilitate amateur
- radio communication using very weak signals.")
+ ${PROJECT_NAME} is a computer program designed to facilitate amateur radio communication using very weak signals.")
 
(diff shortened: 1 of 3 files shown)


Chris-AC9KH:
@rruchte I can change the description back to what it was. But it's not really accurate. My rough guess is that about 75% of JS8Call users use it for automatic beacon/SNR system, and that's all they use it for.

Chris-AC9KH:
I got the builds here for Linux and Windows. I usually grab those from the run, unpack them, test them to make sure they work, then upload to the assets. I'll do that this evening when I get time.

Thanks Joe!

wmiler:
> @wmiler Wyatt, I put together a draft release. Please check and make sure I did everything correctly. I think all that needs to do is to tap "Publish" and she will go. The one thing I'm not sure of is the source code. Will that be tacked on automatically? I don't need to upload the source.zip and source.tgz files do I?
> 
> 73, -Joe- K0OG

@Joe-K0OG  Joe, you never need to grab the source.tgz/zip files. Just going with the installer(s) on windows/linux AppImages is enuf.

I apologize for my out-of-loopness, this storm has done a number on some very expensive PA finals (and a bent antenna 🙀), and I've been going non-stop since I could get back on the roads.

PR #176, opened 2026-01-29
Fix TCP API RX.DIRECTED to match DIRECTED.txt output

mgochoa57:
## Summary

- Fix duplicate callsign issue where FROM was appearing in both the FROM field and VALUE/TEXT fields in TCP API RX.DIRECTED messages
- Include the user-configured eot (end of transmission) character in TCP API output for consistency with DIRECTED.txt file
- Refactor to use single `baseText` variable with clear comments explaining the data flow

## Problem

When JS8Call sends directed messages via TCP API, there were two issues:
1. The sender's callsign appeared twice (in FROM field AND in VALUE/TEXT content)
2. The eot character (♢) appeared in DIRECTED.txt but not in the TCP API live feed

## Solution

Refactored to use a single `baseText` variable:
- `baseText` contains `"TO CMD EXTRA TEXT [eot]"` (used for TCP API where FROM is a separate JSON field)
- `text` contains `"FROM: TO CMD EXTRA TEXT [eot]"` (used for DIRECTED.txt file and UI display)

This ensures the eot character is applied once and consistently appears in both outputs, while keeping FROM out of the VALUE/TEXT fields.

## Test plan

- [ ] Verify DIRECTED.txt file output format is unchanged: `FROM: TO CMD EXTRA TEXT ♢`
- [ ] Verify TCP API RX.DIRECTED VALUE/TEXT fields contain `TO CMD EXTRA TEXT ♢` (no FROM prefix, includes eot)
- [ ] Verify TCP API RX.DIRECTED FROM field contains the callsign separately

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/JS8_Mainwindow/processCommandActivity.cpp
+++ b/JS8_Mainwindow/processCommandActivity.cpp
@@ -123,53 +123,48 @@ void UI_Constructor::processCommandActivity() {
             spotAprsCmd(d);
         }
 
-        // PREPARE CMD TEXT STRING
-        QStringList textList = {
-            QString("%1: %2%3").arg(d.from).arg(d.to).arg(d.cmd)};
+        // PREPARE CMD TEXT
+        // Build baseText without the FROM prefix: "TO CMD EXTRA TEXT [eot]"
+        // This is used for the TCP API where FROM is sent as a separate field.
+        // The full text (with FROM prefix) is used for file logging and display.
+        QStringList baseList = {QString("%1%2").arg(d.to).arg(d.cmd)};
 
         if (!d.extra.isEmpty()) {
-            textList.append(d.extra);
+            baseList.append(d.extra);
         }
 
         if (!d.text.isEmpty()) {
-            textList.append(d.text);
+            baseList.append(d.text);
         }
 
-        QString text = textList.join(" ");
+        QString baseText = baseList.join(" ");
 
-        // Construct value without FROM prefix for TCP API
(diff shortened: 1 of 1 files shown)


wmiler (dismissed):

mgochoa57:
As you know, I’ve been working to fix an issue with duplicate callsigns in the API feed. My first attempt didn’t resolve the problem (#123). The second attempt did eliminate the duplicate callsigns by introducing a second variable, but it also removed the EOT character, which had always been present in the API output.

After reviewing the logic again, I realized the second variable wasn’t necessary. I refactored the code back to a single variable, which restores the EOT character while still addressing the original issue. I haven’t tested this version yet, but I’m confident it should work. I’ll verify it tonight when I get home.

wmiler:
> As you know, I’ve been working to fix an issue with duplicate callsigns in the API feed. My first attempt didn’t resolve the problem (#123). The second attempt did eliminate the duplicate callsigns by introducing a second variable, but it also removed the EOT character, which had always been present in the API output. After reviewing the logic again, I realized the second variable wasn’t necessary. I refactored the code back to a single variable, which restores the EOT character while still addressing the original issue. I haven’t tested this version yet, but I’m confident it should work. I’ll verify it tonight when I get home.

Ah, ok, gotcha.

Chris-AC9KH:
Merged. Thanks!

Issue #126, opened 2026-01-09
Logging


SaludaLabs:
**Describe the bug**
When logging it does not capture the receive report level. I have  enter it manually
Windows 11 JS8Call v2.5.0

wmiler:
> The time off is not logging to N3FJP, Time on is

This is by default in our code. Will need to check what N3FJP will accept.

https://github.com/JS8Call-improved/JS8Call-improved/blob/ffc2697d0fbb4c3d3158adfbdf3901a02c204323/JS8_UI/mainwindow.cpp#L5181-L5200

wmiler:
And it turns out, that is all N3FJP will accept on their [API](https://www.n3fjp.com/help/api.html)

The not populating snr is a bug however.
[logqso.cpp](https://github.com/JS8Call-improved/JS8Call-improved/blob/master/JS8_UI/logqso.cppl)

PR #178, opened 2026-02-01
 Bug fix: #126 Set snr for Log window. Add debugging for LogQSO 

wmiler:


Chris-AC9KH:
Without looking thru the code, off the top of my head it should be somewhere in the m_rptRcvd variable.

wmiler:
> Without looking thru the code, off the top of my head it should be somewhere in the m_rptRcvd variable.

True, but that also isn't set, and seems to be only set when a person enters it. We don't track that anywhere except in snr? Or @hb responses and it's not saved as far as I can see.

wmiler:
Sorry for the delays, I broke my shoulder at the beginning of the month, just now starting to be able to get back in front of a computer.

wmiler:
@Chris-AC9KH I don't see a logical way to get `m_rptRcvd` since it has to be sent by the other station to us. This isn't included in any `MSG` or other station-to-station communications, and we don't have a table or such for heartbeats or `snr?`. I think for now, that can't be implemented. If in the future, we hold the recv snr, that could change.

Don't merge this fix yet, I do have a small bit of clean up to do.

PR #181, opened 2026-02-05
Message Inbox Enhancements

rruchte:
Allow message inbox to dock into the main UI
Enable column sorting in the message table
Rework mark-as-read signals to trigger on row deselection Add SemiSortableHeader subclass of QHeaderView


Chris-AC9KH:
Previously tested, merged! Thanks Rob!

Issue #182, opened 2026-02-06
Notification Sound Playback Problems


rruchte:
Several users have complained about notification sound playback speed problems. 

[https://js8call.groups.io/g/main/topic/audio_notifications_playing/117298890](https://js8call.groups.io/g/main/topic/audio_notifications_playing/117298890)
[https://js8call.groups.io/g/main/topic/sounds/117655535](https://js8call.groups.io/g/main/topic/sounds/117655535)

This seems to be due to a change made to BWFFile in the port to QT6 which removed the 8bit/16bit detection and hard coded playback to 16bit. I reworked the format detection to accurately determine the bit depth and it solves the problem on my machine.

While working on this I noticed that mono wav files play back on the left speaker only. I'm not sure if that is something we want to try to address with some bit manipulation, or leave as is.

PR #183, opened 2026-02-06
Addresses issue #182 - Notification sound playback problems

rruchte:
Fix 8bit playback - set the sample format appropriately based upon the bits per channel in the format chunk instead of defaulting to 16 bits

Enable 24bit playback

Upmix mono to stereo

### DIFF
--- a/JS8_Audio/BWFFile.cpp
+++ b/JS8_Audio/BWFFile.cpp
@@ -156,6 +156,8 @@ class BWFFile::impl final {
     InfoDictionary info_dictionary_;
     qint64 header_length_;
     qint64 data_size_;
+    quint16 bits_per_sample_{0};
+    quint16 block_align_{0};
     FileError error_;
 };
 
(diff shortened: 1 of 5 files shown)


Chris-AC9KH:
This is an interesting one because while I've heard of it, what platforms did it affect? There was never any issue with this on MacOS, for whatever reason. Just a few days ago I set my MSG notification to play a .wav file (that was the only one I could find on my system) for AC/DC "You Shook Me All Night Long". I got a 200w sound system on the Mac mini and forgot I had the volume turned up kinda loud. Had a MSG come in and my wife goes, "what is wrong with your computer? It suddenly came on and sounds like Brian Johnson and Angus Young got into your office and they're shaking the whole house!"

But it was playing normally, no issues with the playback.

rruchte:
I believe it affects all platforms, the issue is the wav file characteristics. I uploaded a zip of test wav files to #182 , try the 8bit mono file and see how it plays on your system.

During the transition to QT6, a change was made to the wav file player code that assumes 16bit instead of determining the actual bit depth of the file. When you play an 8bit file assuming it's 16bit, it does the chipmunk thing. Also, mono files were only playing on the left speaker.

Chris-AC9KH:
Those test files also play fine on the Mac thru JS8Call with no problem detected. Here is a demo of it in operation with the file I was using previously.

https://vimeo.com/1163043646

So I have to believe this is platform-dependent somehow.

Issue #187, opened 2026-02-14
PSK Reporter Failing to Spot, JS8Call 2.6.0


Joe-K0OG:
**Describe the bug**
I found that while running with the "Enable relaying Inbound APRS..." my system will not post spots of stations it hears to PSKReporter.info.  Disabling the APRS relaying immediately fixes the problem.

**Expected behavior**
I expect that the enablement of relaying inbound APRS and PSKReporter spotting should be independent of each other, and should both work simultaneously.

**Desktop (please complete the following information):**
 - OS: Linux Mint 22.3
 - JS8Call 2.6.0 (latest code as of 2/13/26, built on local system)
 - Also tested with AppImage from CI builds

I ran for a few hours today on 60m and during that whole time my station did not spot any of the stations I copied.  As soon as I disabled "Enable relaying Inbound APRS..." the stations I recently heard were immediately spotted.  However, **_I now wonder if there might be some more general problem with PSKReporter.info spotting_** because after running for a while with the APRS inbound disabled, stations I've heard have not been spotted (this is after at least 15 minutes of waiting to see if those spots appear on PSKReporter.info).

Followup a bit later:  I find that if I disable all spotting in settings, click "OK", then go back into settings and re-enable spotting "Enable spotting to reporting networks..." then recently-received stations will now spot, but subsequent receptions will not spot, even after waiting 15 or more minutes.  Perhaps there is indeed a generalized spotting problem.

It would be good if others could test and verify the problem with 2.6.0.

73,
-Joe-
K0OG

rruchte:
@Joe-K0OG the MIN_SEND_INTERVAL used to be 120, but during the big round of PSK Reporter fixes, @aknrdureegaesr [requested that be changed to 600](https://github.com/JS8Call-improved/JS8Call-improved/issues/9#issuecomment-3527230284). Context for that discussion was in @Chris-AC9KH repo during the transition period from the old repo, and is lost to time. At the time, it seemed less than ideal to me, but I'm not a PSK Reporter power user, so I deferred to those with more knowledge on the subject. I suspect that there may be some misunderstanding about frequency of reporting in general vs frequency of repeating identical spots.

TL;DR - The right thing to do is probably revert to 120.

Joe-K0OG:
> I suspect that there may be some misunderstanding about frequency of reporting in general vs frequency of repeating identical spots.
> 
> TL;DR - The right thing to do is probably revert to 120.

Rob,

[...]

PR #189, opened 2026-02-16
Addresses issue #187 - PSK Reporter Failing to Spot

rruchte:
Removes TCP socket functionality from PSKReporter
Removes PSK Reporter TCP option from configuration 
Reverts PSKReporter MIN_SEND_INTERVAL to 120 seconds

### DIFF
--- a/JS8_Network/PSKReporter.cpp
+++ b/JS8_Network/PSKReporter.cpp
@@ -27,7 +27,6 @@
 #include <QRandomGenerator>
 #include <QSharedPointer>
 #include <QString>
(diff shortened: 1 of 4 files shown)


Chris-AC9KH:
I'm pretty sure this will work fine, Rob. People that had that box checked are going to have PSKReporterTCPIP=true hardwired into their JS8Call.ini but I don't think that will cause a problem as it should just ignore it because it no longer checks for it in the code.

I have to run for tonite, but once the build checks complete I'll merge this in the morning, then we can all try it out from master

Issue #191, opened 2026-02-18
Add number of available messages to MSGS? and HB response


rruchte:
Per [this thread](https://github.com/orgs/JS8Call-improved/discussions/174#discussioncomment-15850336) in the What new features would you like in 2.6.0? discussion, we should add an indication for the number of available messages in the response to the MSGS? query and HB response in addition to the ID of the first available message.

If there is more than one pending message, the additional number of messages should be appended to the response in a `+n` format. For example, if W4CAT has three pending messages for KJ4CTD, and the first message ID is 789, the response should be `W4CAT: KJ4CTD YES MSG ID 789 +2`. If 789 is the only pending message, there should be no "+n", the response should be as it is today, `W4CAT: KJ4CTD YES MSG ID 789`

Joe-K0OG:
Are we ready to push this one now?  It's looking good to me.

73,
-Joe-
K0OG

PR #192, opened 2026-02-18
Addresses issue #191 - Add number of available messages to MSGS?, HB, and MSG response

rruchte:
Adds an indication for the number of available messages to the response to the MSGS? query, HB response, and message response in addition to the ID of the next available message.

### DIFF
--- a/JS8_Main/Inbox.cpp
+++ b/JS8_Main/Inbox.cpp
@@ -301,7 +301,7 @@ int Inbox::getLookaheadMessageIdForCallsign(const QString &callsign,
     }
 
     const char *sql = "SELECT inbox_v1.id, inbox_v1.blob FROM inbox_v1 "
-                      "WHERE inbox_v1.id > ? "
+                      "WHERE inbox_v1.id != ? "
                       "AND json_extract(blob, '$.type') = 'STORE' "
                       "AND json_extract(blob, '$.params.TO') LIKE ? "
                       "ORDER BY inbox_v1.id ASC "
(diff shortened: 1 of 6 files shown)


Joe-K0OG (commented):
This seems to work well, however if I have several messages awaiting me, (three, for example), when I pick up the first message, there is no "+2" at the end of the message pickup.  However, if I pick up one of three, then "<CALLSIGN> QUERY MSGS" it will indicate a "+2".  Shouldn't each successive message (until the next-to-last one, of course) always indicate a "+X" messages remaining after a message retreival?

Also, possibly unrelated, if I transmit "@ ALLCALL QUERY MSGS" or send an HB, the messages awaiting notifies me properly.  However, if I pick up one message (so now two more are left in the queue), then transmit "@ ALLCALL QUERY MSGS" there is no reply at all.  If I then transmit "<CALLSIGN> QUERY MSGS" I will get a reply of messages awaiting, indicating a "+2".

73,
-Joe-
K0OG

P.S. I noticed that the 160m frequency was changed.  Was this intentional?

Chris-AC9KH:
> P.S. I noticed that the 160m frequency was changed. Was this intentional?

I changed that for 2.5.0 because everybody was using it anyway, and it gets it 3.5kHz above the FT8 dial freq.

rruchte:
@Joe-K0OG OK, I'll look at the count logic. Can you test the ALLCALL bug in the main branch to see if it's happening there to? I can't think of anything that I did here that would affect that.

rruchte:
@Joe-K0OG The next message ID works differently when retrieving a message than it does when querying messages. When querying messages, the ID returned is the _first available message_. When retrieving a message, the next ID is the next available message _after the one that is being returned_. 

The +n count does not take the message position into consideration, it's the total count of available messages. So you could jump into the linked list of available messages in the middle, pull all of the messages until the end, and still have a count that indicates the first half of the list the you didn't pick up.

We could rework this, we just need to know what the new requirements are.

PR #195, opened 2026-02-19
Update JS8Call desktop file

### LINKED ISSUE #188: JS8Call.desktop: suggest changing "Comment" to distinguish from WSJT-X
I think it would be a good idea to update the "Comment" field in the desktop file here since it's identical to the one in WSJT-X:

https://github.com/JS8Call-improved/JS8Call-improved/blob/master/.github/workflows/misc/JS8Call.desktop

Image

Image

Image

Do we have a good tagline to put there?  [The original JS8Call page](https://js8call.com/) says "Weak signal keyboard-to-keyboard messaging for ham radio" which sounds pretty good.

wmiler:
Fixes #188.

### DIFF
--- a/.github/workflows/misc/JS8Call.desktop
+++ b/.github/workflows/misc/JS8Call.desktop
@@ -1,7 +1,7 @@
 [Desktop Entry]
 Version=1.0
 Name=JS8Call
-Comment=Amateur Radio Weak Signal Operating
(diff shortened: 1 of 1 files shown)


wmiler:
This can be added to a future 2.5.x if desired.

Chris-AC9KH:
Merged. Thanks!

Issue #196, opened 2026-02-21
Segfault in PTT settings due to null reference on validate() call


wmiler:
**Describe the bug**
Setting PTT control to udev alias seg faults JS8Call

**To Reproduce**
1) Make a udev rule (in this case Rigblaster P&P)
`KERNEL=="ttyUSB[0-9]*", SUBSYSTEM=="tty", ATTRS{idVendor}=="10c4", ATTRS{idProduct}=="814a", SYMLINK="ttyS7000"`
2) Open JS8Call
3) Go to Radio Settings and set `Serial Port` to `/dev/ttyS7000` 
4) Test Cat (works)
5) Go to `Rig Options` and set `Serial Port` to `/dev/ttyS7000`
6) Test PTT (JS8Call shuts down with SegFault)

**Expected behavior**
Should accept and trigger PTT without crashing.

**Desktop (please complete the following information):**
 - OS: Ubuntu 24.04
 - Version 2.6.0 master
 - Radio Icom IC-7000
 - Direct to radio

Not sure why it fails yet on the PTT, but works fine on the CAT control. BTW, this was working fine when both were set to `/dev/ttyUSB0`

Chris-AC9KH:
Wyatt, this is likely going to be a permissions/serial number issue for your /dev devices. While Unix doesn't use udev, all hardware serial devices identify themselves on the bus with a serial number. In this case, an example on MacOS with the IC-7300. The bridge controller is a hardware device, audio is virtual over the serial connection.

<img width="426" height="438" alt="Image" src="https://github.com/user-attachments/assets/e2d6ec32-ed1b-4755-95ec-b19d1b5789dc" />

The virtual device can be symlinked. Hardware devices can't because the serial number can only be used once on the stack and symlinking it does not provide the proper permissions to use it. So the kill command for that process comes from the kernel because it's not allowed. I don't know why linux would be any different, but on MacOS (BSD Unix) these are identified by the serial number which points to the **location ID on the bus** (or port number) while linux udev enumerates them as USB0, USB1, etc..

<img width="583" height="76" alt="Image" src="https://github.com/user-attachments/assets/26a3848d-d509-46ca-9dc9-e448f8f9e751" />

Chris-AC9KH:
[...]

PR #197, opened 2026-02-23
Fixes issue #196

wmiler:
Fixes a null reference issue found in #196.

### DIFF
--- a/JS8_UI/Configuration.cpp
+++ b/JS8_UI/Configuration.cpp
@@ -2762,31 +2762,20 @@ bool Configuration::impl::validate() {
 
     auto ptt_method = static_cast<TransceiverFactory::PTTMethod>(
         ui_->PTT_method_button_group->checkedId());
-    auto const ptt_port = ui_->PTT_port_combo_box->currentText();
-    bool port_invalid = false;
+    if (ptt_method == TransceiverFactory::PTT_method_DTR ||
+        ptt_method == TransceiverFactory::PTT_method_RTS) {
+        const auto ptt_port = ui_->PTT_port_combo_box->currentText();
+        auto *model = dynamic_cast<QStandardItemModel *>(ui_->PTT_port_combo_box->model());
+        const int index = ui_->PTT_port_combo_box->findText(ptt_port);
+        QStandardItem *item = (index >= 0 && model) ? model->item(index) : nullptr;
 
-    if (TransceiverFactory::PTT_method_DTR == ptt_method || TransceiverFactory::PTT_method_RTS == ptt_method) {
-        if (ptt_port.isEmpty()) {
-            port_invalid = true;
-        } else {
-            auto* model = qobject_cast<QStandardItemModel*>(ui_->PTT_port_combo_box->model());
-            int idx = ui_->PTT_port_combo_box->findText(ptt_port);
+        const bool enabled = item && item->isEnabled();
+        const bool invalid = ptt_port.isEmpty() || !enabled;
 
-            if (model && idx != -1) {
-                if (QStandardItem* item = model->item(idx)) {
-                    port_invalid = !item->isEnabled();
-                } else {
-                    port_invalid = true;
-                }
-            } else {
-                port_invalid = true;
-            }
+        if (invalid) {
+            JS8MessageBox::critical_message(this, tr("Invalid PTT port"));
+            return false;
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
So here's what happens. With this code I can symlink /dev/MyGreatSerialPort to /dev/cu.usbserial-2110 on MacOS and it will accept it vs the enumerated serial ports.

Screenshot 2026-02-22 at 20 07 19

It skips over validation of the port because I'm using CAT, but the flag that was set validates it anyway as it skips over and only checks DTR and RTS, or if the port is empty. Finds the port in /dev, everything is good to go. Press Test CAT and it turns green even though it doesn't get CAT control. Click OK to close the dialog and then the Tune button on the UI and the kernel instantly kills it by sending a SIGSEGV because we got a serial bus conflict with an invalid address.

Screenshot 2026-02-22 at 19 57 55

Now, is anybody stupid enough to do this on Mac? No. Everybody uses the enumerated ports. Will it ever cause an issue? Not likely. But the old code doesn't do this because it uses a dynamic cast to convert to classes by inheritance and returns a null pointer if it fails. Which doesn't cause anything bad to happen on Mac other than a Hamlib dialog that says the serial port is no good. Nor did it happen on Debian 13 until I tried different software and got different results. Bad serial ports, every time, depend on hardware, motherboard chipsets, drivers, etc..

So this is based on what happened on ONE system. But if this fixes your crash issue on linux due to a bad USB configuration, it is likely it won't hurt anything on the other two platforms. Even though I can verify a bad CAT configuration with this code on MacOS, it is very remote that anybody would ever discover that.

Chris-AC9KH:
@wmiler why don't you title this something like "patch for linux udev rules" and go ahead and merge it. Because I don't see it being an issue on Mac or Windows unless somebody really tries to play with serial ports. Then use it to get your udev rule issue sorted out. Test a properly functioning udev setup with the old code. And if no issue exists then this can be easily reverted if desired. Simply because if somebody DOES play with serial ports this can create a downstream issue in Hamlib.

I think this was a misunderstanding that there's a problem in our code with the dynamic_cast when there's really not. The issue was caused by invalid configuration that you were trying to force, in which case the kernel is going to protect itself from that. With any valid configuration, even if the wrong serial port is selected, the kernel is not going to kill it with the old code and it will just display the error, which it does on Debian 13 in my testing (with the old code) although it takes it awhile to decide that Test CAT failed.

I have no clue why Test CAT passed on your system, because it actually initially turns green, then wait a bit and it turns red and throws the error (with the old code). But, as usual, not all linux systems are created equal.

wmiler:
I'm up for changing it, it doesn't skip the validate call at all, you enter your own /dev/mySquirrellyPort and it will give you the "invalid PTT port". Maybe I missed a check in regards to the Test CAT mode. I can rework it so it only handles the null reference issue.

wmiler:
Ok, thank you. That helps a lot!

Chris-AC9KH:
@wmiler did you have time to test that patch I posted for linux? I have not personally had time to test it with a linux system, but verified it works ok on MacOS. I'm hoping it works on linux, as based on your description with the traceback in the issues category it looks like linux (or the gcc compiler) doesn't like nullptr being passed to ->isEnabled().

wmiler:
> > Interesting. So not all linux systems are affected. I was hoping you could verify Wyatt's original issue but it's looking like that might be only applicable to Ubuntu 24 (so far). I wonder if Ubuntu builds their own kernel, or applies their own patches to it?
> 

Yes, Ubuntu patches the mainline kernel. Not sure what all they do there...

> @Chris-AC9KH @wmiler I just booted Ubuntu 24.04 and tested it with 2.5.2 AppImage and a fresh build of the current main branch, and there is no hang if I try to access a non-existent port in rig setup. It behaves nicely, displaying the "Rig failure" window as it should, allowing me to enter the settings and fix it, or choose a different configuration -- no crashing.

I'm thinking there may be a "strangeness" in the way qserialport does things. We do build with libudev for linux, an qt is supposed to look at the udev first, and then failback to /sys/class if it can't enumerate for some reason. 

Not sure why we fail, but sdrangel succeeds at this. I'll stick it on my backlog list to poke at.

PS @Joe-K0OG If you can grab the AppImage from this last run and give it a whirl, I'd appreciate it.

PR #201, opened 2026-02-26
Fix "Requested output audio format is not valid" startup bug

rruchte:
On Linux, QT can erroneously list QAudioDevice instance as having the mode QAudioDevice::Output when they are actually inputs. This results in the "Requested output audio format is not valid" error on startup, and requires the user to open the configuration and select the audio device in the input list before we can start it.

Adds SoundInput::canOpenAsInput to test if an QAudioDevice instance is actually a viable input

Updates Configuration::impl::find_audio_device to test if device is actually in input by using Adds SoundInput::canOpenAsInput, rather than trusting the mode listed in the device.

### DIFF
--- a/JS8_Audio/SoundInput.cpp
+++ b/JS8_Audio/SoundInput.cpp
@@ -177,6 +177,18 @@ void SoundInput::stop() {
 }
 
 /**
+ * @brief Tests whether device can be opened as an input
+ */
+
+bool SoundInput::canOpenAsInput(const QAudioDevice &dev) {
+    QAudioFormat fmt = dev.preferredFormat();
+    auto src = std::make_unique<QAudioSource>(dev, fmt);
+    src->start(); // opens device
+    src->stop();
+    return src->error() == QAudio::NoError;
+}
+
+/**
  * @brief Destructs the SoundInput object.
  */
 SoundInput::~SoundInput() { stop(); }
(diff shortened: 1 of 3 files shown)


wmiler:
Ok, something about this kills my audio on my system. No audio in the waterfall, and no decodes. The AppImage from PR #199 works fine, the AppImage from this PR does not.

rruchte:
@wmiler OK, can you try something?

First try to add `src->reset()` after `src->stop()` in `SoundInput::canOpenAsInput`.

Like this:
```
bool SoundInput::canOpenAsInput(const QAudioDevice &dev) {
    QAudioFormat fmt = dev.preferredFormat();
    auto src = std::make_unique<QAudioSource>(dev, fmt);
    src->start(); // opens device
    src->stop();
    src->reset(); // <-- Add this
    return src->error() == QAudio::NoError;
}
```

If that does not solve the problem, just make `SoundInput::canOpenAsInput` return `true` and see if that works.

If neither of these things work, we can either remove the check altogether, it's a belt-and-suspenders thing, or revert this patch and people like me with janky Linux audio setups can just deal with resetting the input every time we open the app.

wmiler:
Ok doing the reset() didn't seem to do anything.

Setting it to return true, gave me this:
image

Which obviously isn't right.

Chris-AC9KH:
I had understood that Pipewire was supposed to be the audio transport layer on linux that mimics MacOS's CoreAudio or Windows' WASAPI. It's possible it's not mature enough yet to sort out these issues. The Qt guys complain because linux is the only one that implements drivers in kernel space, the other two can implement drivers in userspace as extensible runtime modules, and this does sort of appear to be a driver-related issue.

PR #213, opened 2026-03-05
RX.GET_FREE_OFFSETS API addition

WebBreacher:
Proposed new API call that allows the JS8Call API to share what offsets are available to use to send traffic. 
-   Reads optional `SPEED`, `LOW`, `HIGH` params (defaults: current mode, 500 Hz, 2500 Hz)
-   For each active offset in `m_bandActivity` (within 30 seconds), computes the blocked zone as `occ ± (requestedBW/2 + signalBW/2)` — accounting for both the occupying signal's actual bandwidth and your intended TX bandwidth
-   Sorts and merges overlapping blocked zones, then returns the gaps as `FREE` segments
-   Omits gaps narrower than your TX bandwidth (they can't fit anyway)
-   Example response: {"params":{"FREE":[{"HIGH":860,"LOW":500,"WIDTH":360},{"HIGH":2500,"LOW":1120,"WIDTH":1380}],"BANDWIDTH":160,"HIGH":2500,"LOW":500,"SPEED":2,"_ID":...},"type":"RX.FREE_OFFSETS","value":""}

### DIFF
--- a/JS8_Mainwindow/networkMessage.cpp
+++ b/JS8_Mainwindow/networkMessage.cpp
@@ -659,6 +659,8 @@ if(type == "STATION.SET_SPOT") {
                             {"BANDWIDTH", bw},
                             {"LOW", low},
                             {"HIGH", high}});
+        return;
+    }
 
     /** @brief RX.GET_FILTER: Returns current filter center, width and enabled state.
      *  @note API 2.6+ */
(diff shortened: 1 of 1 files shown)


JS8_Mainwindow/networkMessage.cpp
```diff
@@ -377,6 +377,75 @@ if(type == "STATION.SET_SPOT") {
                            });
         return;
     }
+    /**
+     * @brief RX.GET_FREE_OFFSETS: Returns contiguous free offset segments in the passband.
+     * Accounts for the bandwidth of each active signal's submode when computing occupied
+     * zones, so returned segments are guaranteed wide enough for the requested TX mode.
+     * Optional params: SPEED (submode int, defaults to current), LOW (Hz, default 500),
+     * HIGH (Hz, default 2500). Entries older than 30 seconds are treated as clear.
+     */
```
wmiler: Add the `@note API 2.6+`

PR #219, opened 2026-03-07
Reduce pre-TX CAT round-trips for FLRig/network rigs

N1MAG:
**Summary**
This PR narrows the pre-transmit rig-control path to reduce timing latency seen on Linux with FLRig/network-backed Hamlib rigs.

**Changes**
- JS8_UI/mainwindow.cpp: in setRig(), emit transceiver_frequency(...) only in non-split mode.
- JS8_Transceiver/HamlibTransceiver.cpp: skip hamlib_bug_bandaid() for RIG_PORT_NETWORK backends to avoid an extra network CAT write before TX.

**Why**
FLRig users report delayed TX start. Both call sites above can add avoidable CAT round-trips in the critical pre-PTT path.

**Scope**
No API changes. No config/schema changes.

**Suggested validation**
1. Linux + FLRig + split none
2. Linux + FLRig + split fake-it
3. Linux + serial Hamlib CAT
4. Windows + FLRig Observe TX trigger-to-PTT delay and confirm no frequency regression in split mode.

### DIFF
--- a/JS8_Transceiver/HamlibTransceiver.cpp
+++ b/JS8_Transceiver/HamlibTransceiver.cpp
@@ -767,6 +767,12 @@ void HamlibTransceiver::do_frequency(Frequency f, MODE m, bool no_ignore) {
 /** Work around hamlib bug
  * [#1966](https://github.com/Hamlib/Hamlib/issues/1966). */
 void HamlibTransceiver::hamlib_bug_bandaid(TransceiverState const &s) {
+    // Avoid an extra network CAT round-trip for network-backed rigs such as
+    // flrig. That extra write can push TX start late on Linux.
+    if (rig_->caps->port_type == RIG_PORT_NETWORK) {
+        return;
+    }
+
     if (s.frequency() == s.tx_frequency() && s.split() && s.ptt()) {
         // Change the frequency ever so slightly without telling anybody.
         // Will not matter during the upcoming transmit and will be corrected
--- a/JS8_UI/mainwindow.cpp
+++ b/JS8_UI/mainwindow.cpp
@@ -6728,9 +6728,9 @@ void UI_Constructor::setRig(Frequency f) {
     if ((m_monitoring || m_transmitting) && m_config.transceiver_online()) {
         if (m_config.split_mode()) {
             Q_EMIT m_config.transceiver_tx_frequency(m_freqTxNominal);
+        } else {
+            Q_EMIT m_config.transceiver_frequency(m_freqNominal);
         }
-
-        Q_EMIT m_config.transceiver_frequency(m_freqNominal);
     }
 }
 


JS8_Transceiver/HamlibTransceiver.cpp
```diff
@@ -767,6 +767,12 @@ void HamlibTransceiver::do_frequency(Frequency f, MODE m, bool no_ignore) {
 /** Work around hamlib bug
  * [#1966](https://github.com/Hamlib/Hamlib/issues/1966). */
 void HamlibTransceiver::hamlib_bug_bandaid(TransceiverState const &s) {
+    // Avoid an extra network CAT round-trip for network-backed rigs such as
+    // flrig. That extra write can push TX start late on Linux.
+    if (rig_->caps->port_type == RIG_PORT_NETWORK) {
+        return;
+    }
```
Chris-AC9KH: Ok, so what we're doing here is simply avoiding a rig polling event if the caps->port_type switch is set to network in the hamlib_bug_bandaid() function.
- This function only executes if split mode was enabled in the UI.
   1. If the receive frequency (s.frequency()) is equal to the transmit frequency (s.tx_frequency()) AND
   2. Split mode is enabled (s.split()) AND
   3. The rig is keyed (s.ptt()).
All of these things have to be true for the function to execute the resultant command. And if so, the code nudges the receive frequency up by 1HZ by calling:
rig_set_freq(rig_.data(), RIG_VFO_CURR, s.frequency() + 1)

Otherwise the function never sends an extra rig command because the state of these function variables is known by UI settings (split mode does not run in the rig) and normal rig polling intervals. This change just makes it skip over checking the conditionals if it's a network connection to the rig.

Chris-AC9KH:
So this delayed PTT issue has been reported clear back to ver 2.2.0 and ver 2.3.1, long before the hamlib bug function was added to the code. So I fail to see how this can do anything. @Joe-K0OG can maybe test the code in this PR because he's been the only one that can (sometimes) reproduce it. And this code doesn't hurt anything if merged, it just bypasses a split mode fix for a hamlib bug if network connection is used.

The PTT delay has never happened with serial connection direct to the rig. Only when Flrig is being used, and only on some linux systems. So I have to leave it to Joe's expertise (probably tested with Hamlib 4.6.5) as to whether or not this actually does anything.

Joe-K0OG:
I'll try to test it this afternoon.

73,
-Joe-
K0OG

Joe-K0OG:
> Plus change the logic a bit to make sure the frequency is reset before the function exits.

I can test this afternoon, but would prefer to test code that is fixed so that this frequency reset is implemented.  Let me know where to pull updated code from, then I'll test.

73,
-Joe-
K0OG

Chris-AC9KH:
GA Joe. I made the changes I would go with and pushed it to this branch:
https://github.com/Chris-AC9KH/JS8Call-improved/tree/pr/PTT_delay

You can ignore the change to the User Guide, I was working on a different project when I thew this together.

Joe-K0OG:
The code builds without error, but it doesn't work properly, so I can't fully test it.  It works fine when I first open it, but if I try to change frequencies/change bands, it fails to change bands.  It changes VFOB, but not VFOA when I change bands.  In other words, this munges VFO handling in RIG split mode.  In addition, it still exhibits PTT delays in the range of a few hundred ms to over one second.

This is a no-go...

73,
-Joe-
K0OG

Joe-K0OG:
@Chris-AC9KH  @N1MAG I'm not sure what this code is supposed to do, but if it is intended to simply suppress invocation of RIG split by Flrig when issuing PTT **when the cursor is between 1500Hz and 2000Hz**, this would NOT be a good way to partially "fix" the problem, as it would only be a partial fix and would exhibit the PTT delay problem outside of the 1500Hz to 2000Hz range.

A better "fix" would be to always use Fake-it split, and as I previously observed set the Poll Interval to 1 second, which seemed to alleviate the PTT-delay problem.

73,
-Joe-
K0OG

Chris-AC9KH:
We don't. That originally came from WSJT-x and was simply retained in the JS8Call UI configuration dialog.

Chris-AC9KH:
There's no sense to removing Rig Split. It works perfectly fine on MacOS with the current 2.6.0 master code. Yes, you can test it if you want with 2.6.0 master, but this PR does nothing to fix any PTT delays.

Chris-AC9KH:
@Joe-K0OG I have two things you can test if you want. 2.6.0 master, of course. And that branch that is here where I reverted that Hamlib patch:
https://github.com/Chris-AC9KH/JS8Call-improved/tree/pr/PTT_delay

They are both the same (master and that branch) except for that patch. I'm curious if you can detect anything there. I can't tell one bit of difference in how it works here on MacOS using a network rig with wfview and split mode.

Joe-K0OG:
@Chris-AC9KH Good - I'll test & compare those two in the morning.

I guess what I was asking is this:  Is there any NEED, any complelling reason to have it RIG split?  I'm having a hard time seeing where RIG split would be better than Fake-it split.  If there is no rational reason to have it, and it occasionally leads to problems, maybe we should remove it.  Just because it "mostly" works seems to not be a good enough reason to keep it if it's not seriously needed.

73,
-Joe-
K0OG

Chris-AC9KH:
> @Chris-AC9KH , thank you for your guidance. I'll step back & review and will share any new insights. This discussion has helped build my understanding of the many considerations.

@N1MAG I'll leave this PR open and mark it as WIP and Do Not Merge for now. Your code, as written, won't work - especially the addition of the `else` statement in mainwindow.cpp in the setRig() function. For anybody running Split mode it won't allow setting the Rx frequency anymore with that `else` statement there. Simply because that final statement in that function is what sets the frequency from the UI, and it HAS to execute no matter what or VFO A is dead in split mode.

The modification to the `hamlib_bug_bandaid()` function in the HamlibTransceiver class also doesn't execute unless split mode is being used. So bypassing it for network-based rigs doesn't really accomplish anything except skipping the conditional statements that check if Rx == Tx, split is enabled, and we're transmitting. If any of those things aren't true that function immediately exits anyway before executing any CAT commands.

Now, after reviewing that function() I don't particularly like it. I don't like the idea of bumping the freq by 1Hz and relying on the next update to set it back. Just because that can fail. But after reviewing that I have suspicions it can be removed from the code for Hamlib 4.7. I have to check with the Hamlib guys to see what they did there.

N1MAG:
Acknowledged. I'll implement per your requirements. My hope is the sequencing for pre-TX may mitigate the delay by sending only one transceiver update instead of multiple back-to-back updates. I'll work on that & test before following up.
73

Chris-AC9KH:
> Maybe something in the documentation would be helpful in case some find RIG split problematic, suggesting to use Fake-it split instead.

@Joe-K0OG it's only problematic if you use Flrig on linux. Going direct to the rig it should work fine.

Many people assume our software directly controls the hardware. It does not. All the transceiver class does is implement the Hamlib API. It is Hamlib that handles the link to the hardware, whether it be thru Flrig, wfview, USB, or whatever. So the chain of command is JS8Call <-> Hamlib <-> Flrig <-> Radio. The communications layer between JS8Call and Hamlib is fine. Where it is breaking down is Hamlib <-> Flrig.

In our software you do not need a special debug build. You can turn on debug logging at any time with environment variables using `QT_LOGGING_RULES="hamlibtransceiver.js8=true` if you want to log the debug output from the HamlibTransceiver class, for instance. You will see that when the communications layer between Hamlib and Flrig starts breaking down on linux, exhibiting the PTT delay, you will get "port_read is busy?". If you trace down where this error comes from, it is in Hamlib. In the current 4.7.0 Hamlib source code that error comes from iofunc.c line 1441, not from JS8Call.

Screenshot 2026-03-10 at 06 52 19

The assumption in this PR is that we must be sending too many commands to the rig. We're doing no such thing. Our UI settings tell Hamlib how often to the poll the rig (which is Flrig in this case). That is usually once a second. Flrig has its own setting for how often it polls the rig. We can send all the commands to Hamlib that we want thru our Hamlib API, but it is Hamlib that controls the polling intervals between itself and the rig.

It is that polling between Hamlib and Flrig that breaks down on linux. It does not break down on MacOS or Windows. So people continue to blame this on some phantom issue in our code. But our code is not handling this communications layer - Hamlib is. And Hamlib is perfectly fine with all this when going straight to the rig. But send it thru Flrig and it breaks down on linux.

The bottom line is; linux has some sort of issue with that Hamlib <-> Flrig connection. I don't know what it is, I'm not a linux developer and only use linux if I'm forced to look at it. So then these linux users say, well, it didn't happen with v2.2.0 so it has to be something in JS8Call. Well, it's not. We went to newer versions of Hamlib to support newer rigs. Our API is written for Hamlib 4 and it will, at best, be only partially functional with the versions of Hamlib that were available back in 2001 when JS8Call v.2.2.0 was released.

So if the linux guys want to find this issue, use the built-in debug logging in JS8Call to find the error. Then trace down where that error comes from. It is not coming from the JS8Call code, you will not find a single reference to "port_read" in our code because our code does not directly communicate with the rig (or Flrig).

N1MAG:
This discussion has been tremendously helpful.
I'll raise some questions to the Hamlib team to gain a better understanding. It appears the way Flrig is implemented there could be coupling with some blocking behavior in the Flrig backend and Linux makes the issue more pronounced.
I have no doubt about the quality of testing & code for JS8Call. The pronounced issue above 2.2.x naturally sparked "what changed" questions and your detailed remarks are exactly what is needed to narrow the focus and hopefully contribute to a solution.
73

PR #223, opened 2026-03-08
Linux Audio Fixes

rruchte:
Refactor audio device settings to use device ID instead of description
Misc audio device identification and handling improvements 
Audio device diagnostics

This is an attempt to alleviate some of the Linux audio problems. This does not address the problem where some oddball inputs are not listed, but it does solve the "Requested input audio format is not supported on device" problem. I'd like to get some Linux users to test this, particularly if you have experienced audio weirdness.

### DIFF
--- a/JS8_UI/Configuration.cpp
+++ b/JS8_UI/Configuration.cpp
@@ -158,15 +158,19 @@
 #include <QApplication>
 #include <QAudioDevice>
 #include <QAudioInput>
+#include <QClipboard>
 #include <QColorDialog>
+#include <QDateTime>
 #include <QDateTimeEdit>
 #include <QDialog>
 #include <QDir>
 #include <QFileDialog>
 #include <QFont>
+#include <QFontDatabase>
 #include <QFontDialog>
 #include <QFormLayout>
 #include <QIntValidator>
+#include <QLibraryInfo>
 #include <QLineEdit>
 #include <QList>
 #include <QLoggingCategory>
@@ -184,7 +188,9 @@
 #include <QString>
 #include <QStringList>
 #include <QStringListModel>
+#include <QSysInfo>
 #include <QTemporaryFile>
+#include <QTextStream>
 #include <QThread>
 #include <QTimeZone>
 #include <QTimer>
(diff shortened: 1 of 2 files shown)


rruchte:
I added Mac OS and Windows output and changed the hints to be generic and only output the Linux-specific line on Linux. I have not yet tested the changes on Mac and Win.

Chris-AC9KH:
I'll pull this and test it for you on Mac.

But I do think the diagnostic output should be left in the code long-term. I think it might come in handy.

Chris-AC9KH:
@rruchte so I connected my iPhone to see how Qt enumerates that, and it does. I'm just posting the pertinent output here:
- The Environment part does not work. On Mac this is privileged information that Qt itself is not allowed access to. You can list it for a specific application, and assume that app is installed in /Applications, but in this case it's not as I'm running it from the build directory
- The MacOS Loader does not work, and never will. No userspace application is allowed access to that. You would have to use the Xcode commandline tool `otool` and that's only going to work on a machine that is used for development and has Xcode installed in it. And further the MacOS commandline terminal does not have access to ANYTHING outside of the user directory by default. So unless the user has given Terminal access to the full filesystem it won't work even with `otool`. May as well remove that one because the security protocols in MacOS are pretty tight.
- You can remove the (best effort) for Mac because the system_profiler output worked fine
- It looks like Qt lists the default devices currently being used for audio I/O by CoreAudio as "Qt Defaults". That's ok just for info, as on Mac any device or software can use the default devices (with permission).

```
JS8Call Audio Device Diagnostics
Timestamp: 2026-03-09T20:31:51

Environment
  Qt / Cross-platform
     QT_PLUGIN_PATH=<unset>
     QT_QPA_PLATFORM_PLUGIN_PATH=<unset>
     QT_QPA_PLATFORM=<unset>
     QT_DEBUG_PLUGINS=<unset>
     QT_LOGGING_RULES=<unset>
     QT_AUDIO_BACKEND=<unset>

  macOS Loader
     DYLD_LIBRARY_PATH=<unset>
     DYLD_FRAMEWORK_PATH=<unset>

Qt Defaults
  Default input:  Chris’ iPhone Microphone | 530B17CB-77CA-4BA5-9487-DEF800000003
  Default output: PBT3014 | FC-58-FA-35-16-2A:output

QMediaDevices::audioInputs() (5)
  - description: Chris’ iPhone Microphone
    id:          530B17CB-77CA-4BA5-9487-DEF800000003
    mode:        1
    chConfig:    8
    prefFmt:     48000 Hz, 1 ch, 4
    formats:     1 2 3 4 
 
macOS Audio (best effort)

sw_vers:
ProductName:		macOS
ProductVersion:		26.3.1
BuildVersion:		25D2128

system_profiler SPAudioDataType:
Audio:

        Chris’ iPhone Microphone:
          Default Input Device: Yes
          Input Channels: 1
          Manufacturer: Apple Inc.
          Current SampleRate: 48000
          Transport: Bluetooth
          Input Source: Default
```

wmiler:
> If it's expanded beyond audio (it arguably is already), we should think about putting it somewhere else. Maybe we should address that now before it gets released.

I'd agree with that.

Chris-AC9KH:
> @Chris-AC9KH Chris, are you saying that I should remove all of the Mac OS specific output?

I would say, maybe. That information that Qt is trying to get is very hard to come by on MacOS due to the way libraries and loader paths are protected from malware trying to insert itself into the system. So the stuff that's labeled not set, if it can be removed individually for Mac, take it out. If it works on the other platforms, it doesn't hurt anything and can leave it in.

rruchte:
@wmiler That's a good place for it, I moved it to its own tab and reverted the Audio tab.


Screenshot From 2026-03-10 23-05-55

Chris-AC9KH:
> OK, I expect this to evolve a lot over time, maybe you can take ownership of the Mac aspects, and someone else can own Windows.

Yeah, that would work fine. Go ahead and work on it, merge it, if I see anything in testing after that that could be an issue for MacOS I can submit a PR to fix it.

PR #224, opened 2026-03-09
WSJT-X UDP protocol: recieve to decode state switching fixed

chrbayer84:
Found a bug: clients like Gridtracker were not notified of state changes between Receive and Decode. So decodes could be missed on occasion.


Chris-AC9KH:
There is no open issue on this one. Fairly straightforward and simple bug fix. Merged. Thanks!

PR #232, opened 2026-03-23
Add the ability to change stylesheets from within the Settings

### LINKED ISSUE #186: Theming On Linux
Qt theming is broken for linux using the appimage install.  The appimage is set to strict and will not allow the system to theme the application using standard qt6ct.  
This is seperate from the bult in UI color picking for the different parts of the UI.

App functions great just wish i could make it not so bright 

thanks for the heard work 
73

wmiler:
Includes a default stylesheet that users should copy and modify to their needs. 

The stylesheet most likely doesn't need to be in JS8_Include, but rather in one of the packaging directories. 

Also the stylesheet most likely is not complete, I just did a quick hand scraping of the ui files. Open to suggestions on where to put it.

Fixes: #186 
Relates: #228 

Please test on Mac, Windows, and Mint.


Joe-K0OG (commented):
Well @rruchte Rob, the TX audio fader dB font and many others are a little larger than I liked, but it does seem to fix the problem.  I tried two .qss files:  The JS8call-default.qss (the new one you posted) and another one I downloaded.  I found with the other one I downloaded that the slider font is way too large, but if I add the following block of code from your updated .qss file to that other .qss file, it works fine:
```
* {
  font-size: 9pt;
  background-color: white;
  color: #242424;
 	selection-background-color: #242424;
 	selection-color: white;
}
```
Where you had font-size at 11pt, I changed to 9pt to get the appearance I wanted.  So, I think it is workable now, although I would prefer that the fix be in code rather than the .qss file, so that any other .qss file can be used without modification as I had to do.

73,
-Joe-
K0OG

Joe-K0OG:
@wmiler Wyatt, I found another possible fix for the large-text in the attenuation slider.  In the file JS8_UI/mainwindow.ui, line 1823, change the font size from 16 to 10, and that works well on both Windows and Linux for me, loading a .qss file without your "fix" code in it.
 ```
 <property name="font">
              <font>
               <pointsize>16</pointsize>
              </font>
 </property>
```
73,
-Joe-
K0OG

wmiler:
I'm breaking this PR up to get this moving again. First part is #243 which just adds the code to load stylesheets. It does **not** contain the default.qss file. That thing is a bear :)

wmiler:
See #247 for the other part of this PR. Option switches will be fixed at some point in the future.

PR #233, opened 2026-03-25
Minor changes to Pills 

Beiri22:
This right now has the following:
- Small docs entry with 1 picture
- fix of unknown chars for command pill without the argument: a default text to replace the characters.
- the default cq message has a grid param; the manual shows this for all cq messages. this is now activated for pills also.
- rewording of msg to: tooltip
- restrictive group highlighting


wmiler:
> * Images for pill highlight: Shall I add 1 or 2 images directly by the added text; or shall we use a few of the later example messages (under message types) side by side with an image?

1 image is fine I think, doesn't need to be the whole screen, ie just the message window. Try to get  a tooltip in the screen grab as well. 

If you want to, you can stick it both ways (on and off) in the Messages section. Pretend you have never used the program before and somebody told you to RTFM. :)

Joe-K0OG (approved):
This appears good to me.  I think it is ready to merge.

73,
-Joe-
K0OG

PR #236, opened 2026-03-31
Add non-versioned development builds for master

Chris-AC9KH:
This is the final implementation of the new CMake versioning system. It removes versioned builds done on master. Only release builds are versioned. It adds an explanation in CMakeLists.txt where the version is set that a 0.0.0 string will result in a development build. Development builds are displayed like so:

Screenshot 2026-03-30 at 20 27 05

If you put any other valid major.minor.patch version in there then it switches to a versioned build, which in this case I changed to the latest release version of MacOS 26.

Screenshot 2026-03-30 at 20 49 43

I left in place Allan's "Not for Release" in the event a build is done without being configured with CMake. That is basically a "catch all" in the event someone would configure say Xcode or Qt Creator to do a build without configuring with CMake. The software is then labeled as "Not for Release" since it did come from our official build channels.

So this should fully implement the CMake versioning system that Allan had envisioned. Allan and I talk back and forth on occasion and I told him I would finish implementing this as the final step after the CMake overhaul.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,12 +52,13 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 12.0)
 
 #------------------------------------------------------------------------------#
-# Project
+# Project configuration. The software build version is set here. If the version
+# is set to 0.0.0 the build will be unversioned and marked as "Development"
 #------------------------------------------------------------------------------#
 
 project(
   JS8Call
-  VERSION   2.6.0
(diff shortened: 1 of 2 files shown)


wmiler:
Overall I don't have an issue with this, and it looks good. Thanks for your work Chris!

Would it be a correct assumption that we won't be supporting dev builds (excluding PRs that generate them and only for the life of the PR)? 

I'm hemming and hawing to myself that it would be hard to follow where in the overall versions the dev build of the app is at. Hopefully that made sense :) Or I could just be overthinking things, I have been in 3 different branches today, lol.

PR #237, opened 2026-04-01
clean up some legacy wsjtx code

Chris-AC9KH:
This should remove the last traces of legacy wsjtx code in our codebase. I wouldn't have even submitted a PR for this since it does not change any functionality in the code. But I had already applied the commit to a PR branch instead of directly to master. So going the PR route with it.


wmiler:
> Incidentally, the Windows resource file (wsjt.rc) was also left behind but using the JS8Call .ico file. So I renamed that to jscall.rc and fixed that in CMakeList.txt as well. This won't affect your CI builds since all that resource file does is point to the location of the Windows .ico file. As long as CMake can find that all is good.

Even the Windows CI installer generation only uses `icons\windows-icons\js8call.ico` so that doesn't affect anything.

PR #239, opened 2026-04-02
Dockable Message Panel Fixes

### LINKED ISSUE #238: MSG Inbox flag handling
This would likely be @rruchte department and I'll admit I have not looked at the source code for this.

I prefer to use the MSG Inbox docked with the UI. I really like that. When a new MSG comes in I think it would be better to clear the unread flag when that MSG is selected instead of clearing it on de-select.

The other issue I note, and maybe this only happens on MacOS in which case it will be up to me to identify the issue and fix it; the flag that appears in the Call Sign list does not clear unless the MSG Inbox is closed and reopened. The expected behavior would be that the flag in the Call Sign list would clear when the new MSG is selected in the MSG inbox list and it clears the unread flag there.

Chris-AC9KH:
- Change logic of MessagePanel so the first row is not automatically selected when the table refreshes
- clear the callsign list unread flag when the MSG is selected in the MSG table
- Make column 0 a NonSortableColumn to prevent unwanted UI anomoly that would sent the selected row to the bottom of the unread rows

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/238

Rob already looked at this in the Issues department. It has only been tested on MacOS.

### DIFF
--- a/JS8_UI/MessagePanel.cpp
+++ b/JS8_UI/MessagePanel.cpp
@@ -54,6 +54,7 @@ MessagePanel::MessagePanel(QString inboxPath, QWidget *parent)
 
   auto *hdr = new SemiSortableHeader(Qt::Horizontal, table);
   hdr->addNonSortableColumn(6);
+  hdr->addNonSortableColumn(0);
   hdr->setSectionResizeMode(QHeaderView::Interactive);
   hdr->setStretchLastSection(true);
   hdr->attachTo(table);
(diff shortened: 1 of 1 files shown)


Joe-K0OG (approved):
Built and tested on Win11x64 (JTSDK64-Tools), and Linux Mint 22.3, and all builds without problems, and seems to function properly.

73,
-Joe-
K0OG

rruchte (approved):
@Chris-AC9KH This looks good to me.

PR #242, opened 2026-04-05
Update links for libraries now hosted on js8lib

Chris-AC9KH:
New change for 3.0, libraries and build assets are now hosted on js8lib releases instead of on the main repository releases. This changes the links to where the libraries are hosted. As soon as this is merged so CI builds don't get broken, the build assets 2.6 release on the main repo will be removed.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -41,12 +41,12 @@ jobs:
 
     - name: Grab build dependeneces
       run: |
-        curl --location --output /tmp/js8lib2.6-MacOS_AppleSilicon_Qt.tar.gz https://github.com/JS8Call-improved/JS8Call-improved/releases/download/lib%2F2.6/js8lib2.6-MacOS_AppleSilicon_Qt.tar.gz
+        curl --location --output /tmp/js8lib3.0-MacOS_AppleSilicon_Qt.tar.gz https://github.com/JS8Call-improved/js8lib/releases/download/lib%2F3.0/js8lib3.0-MacOS_AppleSilicon_Qt.tar.gz
         cd /tmp
-        tar zxf /tmp/js8lib2.6-MacOS_AppleSilicon_Qt.tar.gz
+        tar zxf /tmp/js8lib3.0-MacOS_AppleSilicon_Qt.tar.gz
         sudo mv /tmp/js8lib /usr/local/js8lib
 
-        curl --location --output /tmp/ffmpeg_MacOS_arm64.tar.gz https://github.com/JS8Call-improved/JS8Call-improved/releases/download/lib%2F2.6/ffmpeg_MacOS_arm64.tar.gz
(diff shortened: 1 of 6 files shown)


wmiler:
How are you building QT for Linux? I don't see it listed in js8libs at all.

Follow on question, do we want to go ahead and build all the Linux libraries as well, and make a big package like you did for MacOS?

wmiler:
> * The Qt 6.9.3 builds for linux were done separately a long time ago, before 2.4.0. I just transferred those builds to the assets for 3.0. I left them in the old build assets for now, as well. But I believe the only place they are used is in the linux build scripts. They might be usable for CI builds as well, but I wouldn't have a clue how to set that up for linux.

Ok.

> Now, on to your question on a complete package for linux; I THINK this could be done? And I'm pretty sure it could be done by creating a linux build branch that is similar to the MacOS ones? The MacOS branches have a Build script that fully automates the process of building a new library. So I'm assuming that could be done for linux as well, which would be more desirable for both the build scripts and (possibly) AppImage builds.
> 

> The problem with the build scripts is that we depend on distro-supplied libraries like Hamlib, boost, et al. So if you build on Mint with the build scripts for instance, you will not have support for new rigs because Mint still uses Hamlib 4.6.2. So I think there may be some advantages to building a Hamlib/boost/fftw/libusb package for linux. We already have a Qt6.9.3 package, it's just a change to the cmake configuration to use them both.
> 
> So this may make CI builds less complicated (and not have to rely on caching) if you only have to pull two linux packages from js8lib - one for the libraries and one for Qt.
> 

CI builds are a special case, as I have debug builds going for hamlib and Js8call setup since we tend to have a fair amount of Js8->hamlib->rig issues, and having debug turned on for hamlib and Js8 makes those easier to spot. The other libs are fine without debug.

For normal users, having the optimized builds for release/candidate would be best.

> I could create a js8lib linux branch and try to set it up similar to MacOS where it would install in /usr/local/js8lib. I'm pretty sure this can be done without conflicting with other apps the user might have installed like WSJT-x which probably relies on system libraries.
> 

I would install everything under ~/.local/lib/js8lib/ that way they won't have to deal with sudo and the various security policies that relate to things under /usr/ I would also install QT there to keep everything together.

>  For the build scripts we sandbox it by installing it in ~/.local/lib/Qt. So for this reason, I would recommend against making a one-piece linux package, and go with a two-piece package.

Agreed.

wmiler (approved):

Chris-AC9KH:
@wmiler I am working on a arm64 linux library build. I took your advice and am installing it to ~/.local/lib/js8lib. But I am finding out that the linux "build-essential" package does not contain many of the tools needed to build libraries. This all comes on Mac by default in Xcode, which sets up a full development environment. There does not appear to be any such thing for linux.

So I'm going thru it step by step, analyzing the build errors, and installing whatever build tools are missing as I go.

For linux, I should include in the first part of the BUILD.sh script to install those missing tools, otherwise it will be a nightmare on linux to build libraries.

PR #245, opened 2026-04-07
restrict pill display to callsigns in relay hops

### LINKED ISSUE #240: Minor Pills formatting with ">"
This may not be something that can (or should) be changed.  I noticed that when I use the ">" sign in order to trigger a simple ACK when sending a message, the pill encapsulation seems inappropriate.  Example from just now:
Image

In this case, the first word of a message is interpreted as a callsign by the pill encapsulator.  If it can be fixed so that only the "AC9KH>" is encapsulated in this case, yet it will still work properly if text following a ">" is indeed a callsign, then that would be good.  Perhaps this can be handled by parsing the text after "AC9KH>" and before the trailing space, and if it is in the form of a valid callsign or group name (text that begins with "@"), then it is encapsulated, yet if it is not a valid callsign then only the "AC9KH>" is encapsulated.

73,
-Joe-
K0OG

Beiri22:
fixes #240

right now only allow callsigns in relay chains; waiting to get more information on validity of group calls in relay chains.

### DIFF
--- a/JS8_Main/DirectedMessageParser.cpp
+++ b/JS8_Main/DirectedMessageParser.cpp
@@ -123,6 +123,12 @@ const QRegularExpression &grid4Re() {
     return re;
 }
 
+const QRegularExpression &runtimeRelayFollowHopStandaloneRe() {
+    static const QRegularExpression re(
+        R"(^\b(?<prefix>[A-Z0-9]{1,4}\/)?(?<base>([0-9A-Z])?([0-9A-Z])([0-9])([A-Z])?([A-Z])?([A-Z])?)(?<suffix>\/[A-Z0-9]{1,4})?\b$)");
+    return re;
+}
+
 const PillCommandDef *findCommandDef(const QString &cmd) {
     for (const auto &def : s_commandDefs) {
         if (cmd == QLatin1String(def.command))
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
@Beiri22 groupcalls are not a valid relay chain item.

@Joe-K0OG are you up for testing this to see if it solves your issue? The code looks ok to me. If it meets your expectations then I'll run the CI builds and merge it.

Joe-K0OG:
> @Beiri22 groupcalls are not a valid relay chain item.
> 
> @Joe-K0OG are you up for testing this to see if it solves your issue? The code looks ok to me. If it meets your expectations then I'll run the CI builds and merge it.

Yes, I'll test it this AM.  Thanks guys.

73,
-Joe-
K0OG

Beiri22:
> @Beiri22 groupcalls are not a valid relay chain item.
> 
> @Joe-K0OG are you up for testing this to see if it solves your issue? The code looks ok to me. If it meets your expectations then I'll run the CI builds and merge it.

ok.

Joe-K0OG:
@Beiri22 I just built & tested it, and as far as I can tell the callsign checking with the ">" symbol is working properly.

I also notice that `MSG TO: [callsign]` and `QUERY CALL [callsign]?` don't do callsign checking.  Can you add the same checking to those two as well?  As it is now I can type unlimited strings of text where there ought to be a qualified callsign.

73,
-Joe-
K0OG

Beiri22:
You're right. Arguments of commands are till now only plain strings. Just for clarity: 
- Both arguments are only callsigns, no groups allowed?
- are there any other commands that have calls or groups as parameter?
- are there any other commands that have arguments which can be checked with something like a regex?

Joe-K0OG:
@Beiri22 To see a summary of the command list, right-click on a callsign in the Call Activity window:

image

As you can see, there are only ">"  `MSG TO: [callsign]` and `QUERY CALL [callsign]?` that need to do callsign checking.

As I understand it, they should only be callsigns.  Maybe @Chris-AC9KH can verify this.

It does not appear that anything else needs checking.  The only other thing you *could* check would the message ID, but I don't think it's worth doing anything about.

73,
-Joe-
K0OG

Chris-AC9KH:
>As I understand it, they should only be callsigns. Maybe @Chris-AC9KH can verify this.

@Joe-K0OG you are correct. Groupcalls are not a vaiid item in a relay chain.

Beiri22:
What about Heartbeat and CQ argument of type Grid Locator? What Grid Locator lengths are allowed? Are other values to HB and CQ are possible?

Joe-K0OG:
> What about Heartbeat and CQ argument of type Grid Locator? What Grid Locator lengths are allowed? Are other values to HB and CQ are possible?

Well, if you want to check all that stuff, I guess you could.

Grid locators would range from a minimum of four characters, to a maximum of twelve, for example:
Minumum:  EM47
Maximum:  EM47CW99VT40

If you really want to get picky about checking grid locators, it's always:
AANNAANNAANN
...where AA = alpha characters and NN = numerals.

I don't think you need to do any checking on HB or CQ.

73,
-Joe-
K0OG

Joe-K0OG:
@Beiri22 Can you re-base your repository and check your modifications to make sure all is good with the latest changes to master from yesterday?  

I made a patch and applied it to the current master and built it on Windows, and it seems to be good.

If there are no problems, maybe we can finish this one.

73,
-Joe-
K0OG

Beiri22:
Just rebased. A quick look looks ok; the other changes do not seem to interfere with this one

Joe-K0OG (approved):
In my testing on Linux and Windows, this seems to work as it should, so I think it is ready to merge, functionally speaking.  @Beiri22 thanks for doing this.

73,
-Joe-
K0OG

rruchte (approved):
Looks good to me.

wmiler (approved):
Looks fine. Can be merged on CI passing. Thanks!

Issue #244, opened 2026-04-06
Cleanup UI - Remove "JS8" from Status Bar


Joe-K0OG:
I suggest that we remove the "JS8" from the status bar, since no other mode would ever be present.  This is leftovers from WSJT-X code I think.

<img width="508" height="82" alt="Image" src="https://github.com/user-attachments/assets/ff9a8216-1b8d-4a58-9402-1d653027ed85" />

Perhaps that space can be allocated to the `TX:` and `Last TX:` fields.

73,
-Joe-
K0OG

Joe-K0OG:
@Chris-AC9KH That would be redundant since the speed is prominently displayed at the top-right, but would be more useful than just "JS8".  I would prefer to remove it and use that space for something else.

Let's see what others think.  If they like your suggestion, then go for it.

73,
-Joe-
K0OG

Chris-AC9KH:
The mode button is also redundant with the repeat button for HB status.

A status bar, which is what we're talking about, has the purpose of being a summary. For instance, if you select to not show frequency and clock your mode button on the top right disappears. In that case the status bar indicates what speed mode is selected. It doesn't now, because it's a static text string. But that branch I made the mods on makes it dynamic.

There is plenty of people that run JS8Call without that top header because it's a bunch of wasted screen space - Tom K1TWH is one. In that case I could also add the frequency being used to the status bar.

Chris-AC9KH:
@Joe-K0OG I added an Auto Reply: widget like so:

<img width="735" height="368" alt="Image" src="https://github.com/user-attachments/assets/bf0ca992-3808-4c61-bbc1-e3120b489bb5" />

The watchdog timer can turn Auto Reply off. So when looking at UI_Constructor::on_the_minute() I noticed that the watchdog does NOT turn off Auto Reply unless a Tx event is attempted after the timer expires. I'm think, well, this is totally stupid. So somebody could have JS8Call running with the watchdog enabled. They're doing something else on their computer and the watchdog timer expires. Nothing happens. Suddenly somebody sends them a MSG and when it tries to ACK the MSG suddenly the watchdog kicks in and the ACK don't work.

How about notifying when the timer actually expires instead? That way, if they're at the console they can reset it. So I fixed that.

<img width="771" height="476" alt="Image" src="https://github.com/user-attachments/assets/1df7927a-a989-4fbf-967b-6e577cffd9e0" />

Chris-AC9KH:
[...]

PR #247, opened 2026-04-08
QT Stylesheet of default styles

wmiler:
Broken off from PR #232.

Task list to be completed
1. Fully add everything style relevant from the ui files (fonts, etc)
2. As well as items buried in the cpp and header files
3. Verify that the QSS agrees with ui files, such that a user can "switch back"

Caveats:
As @Joe-K0OG Joe [found](https://github.com/JS8Call-improved/JS8Call-improved/issues/244#issuecomment-4207350888) out, there are a few things that are defined in the cpp files that are not in the ui files, and currently need fixes to enable access from a stylesheet.

Currently (as far as I've searched) the following have no QSS mapping, see [comment](https://github.com/JS8Call-improved/JS8Call-improved/issues/244#issuecomment-4208139155) 
```
tx_status_label
config_label
mode_label
last_tx_label
auto_tx_label
progressBar
wpm_label
```

### DIFF
--- a/.github/workflows/misc/JS8Call-default.qss
+++ b/.github/workflows/misc/JS8Call-default.qss
@@ -1,0 +1,355 @@
+/* JS8Call Default Style Sheet */
+
+/* Reset a few things to catch edge cases as reported by Rob Ruchte (Manjaro with Gnome) */
+
+* {
+  font-size: 11pt;
+  background-color: white;
+  color: #242424;
+  selection-background-color: #242424;
+  selection-color: white;
+}
+
+/* Configuration.ui */
+
+QPushButton:checked#test_PTT_button {
+  background-color: red;
+  border-style    : outset;
+  border-width: 1px;
+  border-radius: 5px;
+  border-color: black;
+  min-width: 5em;
+  padding: 3px;
+}
+
+QLabel#tableSelectionBackgroundLabel {
+  background-color: #3498db;
+}
+
+QLabel#labCQ {
+  background-color: #66ff66;
+}
+
+QLabel#primaryHighlightLabel {
+  background-color: #f1c40f;
+}
+
+QLabel#tableForegroundLabel {
+  background-color: white;
+}
+
+QLabel#tableBackgroundLabel {
+  background-color: white;
+}
+
+QLabel#cqMessagesLabel {
+  background-color: #66ff66;
+}
+
+QLabel#labMyCall {
+  background-color: #ff6666;
+}
+
+QLabel#secondaryHighlightLabel {
+  background-color: #ffff66;
+}   
+
+QLabel#rxLabel {
+  background-color: #ffeaa7;
+}
+
+QLabel#txForegroundLabel {
+  background-color: #ffeaa7;
+}
+
+QLabel#rxForegroundLabel {
+  background-color: #ffeaa7;
+}
+
+QLabel#composeLabel {
+  background-color: white;
+}
+
+/* Pill highlighting */
+
+QLabel:pillRecipientLabel {
+  font-weight:bold;
+  background-color: #2980b9;
+  color: #ffffff;
+}
+
+QLabel:pillCommandLabel {
+  font-weight:bold;
+  background-color: #e67e22;
+  color: #ffffff;
+}
+
+Qlabel:pillGroupLabel {
+  font-weight:bold;
+  background-color: #e74c3c;
+  color: #ffffff;
+}
+
+QLabel:pillSenderLabel {
+  font-weight:bold;
+  background-color: #7f8c8d;
+  color: #ffffff;
+}
+
+/* End Pill highlighting */
+
+/* mainwindow.ui */
+
+QFrame#logWidget {
+  background-color: black;
+}
+
+QFrame#frame {
+  background-color: black;
+  color: white;
+}
+
+QPushButton#dialFreqUpButton {
+  background-color:black  ;
+  background-image:url(":/images/artwork/up.png");
+  background-position: center center;
+  background-repeat:no-repeat;
+  border:1px solid;
+  border-radius:2px;
+}
+
+QLabel#labDialFreqOffset {
+  font-size: 14pt;
+  line-height:12pt;
+  color: white;
+}
+
+QFrame#frame_6 {
+  background-color: black;
+}
+
+QLabel#labCallsign {
+  font-size: 14pt;
+  line-height:12pt;
+  color: white;
+}
+
+Qlabel#labUTC {
+  font-size: 14pt;
+  line-height:12pt;
+  color: white;
+}
+
+QFrame:QPushButton#buttonGrid {
+  background-color: lightgray;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QFrame:QPushButton:checked#buttonGrid {
+  background-color: #6699ff;
+  padding:0.25em 0.25em; font-weight:bold;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton#monitorTxButton {
+  background-color:lightgray  ;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#monitorTxButton {
+  background-color:#22FF22;
+}
+
+QPushButton#monitorTxButton[transmitting="true"] {
+  background-color:#FF2222;
+}
+
+QPushButton#monitorButton {
+  background-color:lightgray  ;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#monitorButton {
+  background-color:#22FF22;
+}
+
+QPushButton#logQSOButton {
+  background-color:lightgray  ;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#logQSOButton {  
+  background-color:#22FF22;
+}
+
+QPushButton#tuneButton {
+  background-color:lightgray  ;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#tuneButton {
+  background-color: #6699ff;
+}
+
+QPushButton#modeButton {
+  background-color: #6699ff  ;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#modeButton {
+  background-color: #6699ff;
+}
+
+QPushButton#spotButton {
+  background-color:lightgray;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#spotButton {
+  background-color: #6699ff;
+}
+
+QTextEdit#textEditRx {
+  background-color: #ffeaa7;
+}
+
+/* Custom Widget is defined in JS8_Main/TransmitTextEdit.h */
+
+QTextEdit:TransmitTextEdit#textEditTx {
+  background-color: white;
+  font-size: 11pt;
+}
+
+QTextEdit:TransmitTextEdit:QTextEdit[readOnly="true"] {
+  background: #EEE;
+  font-style: italic;
+}
+
+QTextEdit:TransmitTextEdit:QTextEdit[transmitting="true"] {
+  background: #EEE;
+  font-style: italic;
+}
+
+QPushButton#hbMacroButton {
+  background-color:lightgray;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#hbMacroButton {
+  font-weight:bold;
+  color:black;
+}
+
+QPushButton#cqMacroButton {
+  background-color:lightgray;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#cqMacroButton {
+  font-weight:bold;
+  color:black;
+}
+
+QPushButton#startTxMacroButton {
+  background-color:lightgray;
+  padding:0.25em 0.25em; font-weight:normal;
+  border-style:solid;
+  border-width:0px;
+  border-radius:2px;
+}
+
+QPushButton:checked#startTxMacroButton {
+  background-color: #22FF22;
+  color: #222;
+}
+
+QPushButton#startTxMacroButton[transmitting="true"] {
+  background-color: #FF2222;
+  color: #222;
+}
+
+QWidget#widget_3 {
+  background-color: black;
+}
+
+QPushButton#readFreq {
+  font-family: helvetica;
+  font-size: 9pt;
+  font-weight: bold;
+  background-color: white;
+  color: black;
+  border-style: solid;
+  border-width:0px;
+  border-color: gray;
+  min-width:40px;
+  min-height:30px;
+}
+
+QPushButton#readFreq[state="warning"] {
+  background-color: orange;
+}
+
+QPushButton#readFreq[state="ok"] {
+  background-color: #22FF22;
+}
+
+QFrame#logWidget {
+  background: black;
+}
+
+/* Custom Widget is defined in JS8_Main/AttenuationSlider.h */
+Qslider:AttenuationSlider#outAttenuation {
+  font-size: 10;
+}
+
+/* Custom Widget is defined in JS8_Main/SignalMeter.h */
+/* Meter component, which displays to the right of the scale, as a  */
+/* level gauge with a peak hold indicator. Displays green when the  */
+/* level is good, yellow when it's too low, red when it's too high. */
+
+QWidget:SignalMeter#signal_meter_widget {
+  color: white;
+  font-size: 11pt;
+}
+
+/* MessagePanel.ui */
+
+QTextEdit#messageTextEdit {
+  background-color: #ffeaa7;
+}
+
+/* WideGraph.ui */
+
+QPushButton:checked#autoDriftButton {
+  font-weight:bold;
+  color:black;
+}


Chris-AC9KH:
@wmiler I thought there was not going to be any default stylesheet loaded? The danger with attempting to define the UI theming, look and feel, with a stylesheet is that the three platforms all have their own individual Qt style. It is bundled in the Qt Plugins. There is no way you're going to come up with a default Qt style sheet for MacOS 26 with Liquid Glass UI. It will make JS8Call stick out like a sore thumb, looking like somebody tried to drop a linux app into MacOS.

As long as there is no stylesheets loaded at runtime by default, and any use of stylesheets at all is totally optional (including a default one), this might fly.

Joe-K0OG:
> As long as there is no stylesheets loaded at runtime by default, and any use of stylesheets at all is totally optional (including a default one), this might fly.

Yes, that is what I understood would be done:  No default .qss will be loaded on running by default, but we would provide one for people to look at as a starting point, and allow people to load one if they wish.  It would be good to have a command-line switch for that (the default from Qt, "`-stylesheet <filename>`" doesn't work right now for some reason).

73,
-Joe-
K0OG

wmiler:
> > As long as there is no stylesheets loaded at runtime by default, and any use of stylesheets at all is totally optional (including a default one), this might fly.
> 
> Yes, that is what I understood would be done: No default .qss will be loaded on running by default, but we would provide one for people to look at as a starting point, and allow people to load one if they wish. It would be good to have a command-line switch for that (the default from Qt, "`-stylesheet <filename>`" doesn't work right now for some reason).
> 
> 73, -Joe- K0OG

What @Joe-K0OG said. 

I'm not sure why you (@Chris-AC9KH ) keep going about how this will be a stylesheet that will be loaded at startup. This stylesheet simply breaks out all the addressable QLabels and what not, that **could be** styled if one would like to do so, and then would be loadable via the Settings->UI->Stylesheet button. Nothing here forces anybody into anything.

Joe, regarding the switch issue, that is a WIP on my side, there are issues with other flags as well.

Chris-AC9KH:
> I'm not sure why you (@Chris-AC9KH ) keep going about how this will be a stylesheet that will be loaded at startup.

Because that is typically what a default stylesheet would be used for. So just want to verify we're not going to stylesheets to define the UI. QSS is horribly slow. It requires parsing strings and re-calculating styles every time a widget is created, reparented, or refreshed. Historically stylesheets do not mix well with QStyle subclasses so it would require an invasive overhaul of the code to put hooks in it for stylesheets since there are elements of the JS8Call UI that are hard-coded for a reason - the main one being cpu and memory usage, and performance.

QSS is ok to play with for theming on borders, backgrounds and text. But once you start trying to modify QLabels and widgets it has to be realized that relying on it for layout modifications or dynamic updates is generally not a good idea.

Chris-AC9KH:
@wmiler I'm overhauling the "look and feel" of JS8Call to be more inline with modern standards. So I defined three new helper methods in the mainwindow header. I'm using the QGuiApplication QOperatingSystemVersion macros to determine how to apply "look and feel". This can now be defined in the header

https://github.com/JS8Call-improved/JS8Call-improved/commit/57b2291ed4775ada656fb2a38912071670962b6f

The result on MacOS 26 is that I get this look

Screenshot 2026-04-09 at 21 21 19

This could be applied platform-wide like it is with the message composition "pills". But I was told that linux users (and possibly Windows too) prefer the old "designed on Linux Windows 95-copied look" with ugly square widgets, buttons and labels.

So I've only gotten started with this but I'll complete the UI overhaul by the end of this weekend. On Linux and Windows it should have no effect. But on MacOS (and iOS) it will make JS8Call look like a native Swift-coded MacOS app.

Chris-AC9KH:
@wmiler @Joe-K0OG the beauty of defining this in the header is that I can use something like this in the source file and define any custom colors as bg, fg variables for any StatusLabel.
```
auto style = preferRoundedStatusLabels()
           ? roundedStatusLabelStyle("#6699ff", "#000000")
           : squareStatusLabelStyle("#6699ff", "#000000");

label->setStyleSheet(style);
```

Or I can define it at compile time with pre-processor directives in the header with something like:
```
#if defined (Q_OS_MACOS)
    return roundedStatusLabelStyle(bg, fg);
#elif defined (Q_OS_WIN)
    return QStringLiteral(
        blah blah blah......
```

Windows has a standard Qt style, which is presently being forced to the old "designed on Linux Windows 95-copied look" but I can define that in the header for Windows to give the app native look on Windows 11. I don't think there is one for linux - at least not that I can find in any Qt documentation. So I guess Linux stays with the old "designed on Linux Windows 95-copied look" unless somebody can define a different one.

There is no need to muddy up the code to apply QSS StyleSheets. If I design the code the way it should've been designed in the first place, you can add your QSS StyleSheets to the header definitions instead of plastering the code with hooks that make QSS work for linux.

Chris-AC9KH:
@wmiler ok, so I fixed this for you. I decided to use the preprocessor and compile the platform styles into the binary because it uses less runtime code and is more efficient at applying the platform-specific styles. I added a definition for linux QSS with some example code that you can use to call it.

This currently only applies to the labels, but it is an example of how you can load a StyleSheet at runtime if you want to on linux, but it preserves the native platform styles for Mac and Windows, which are well defined. Linux is not well-defined and is a hodge-podge, so you can StyleSheet away all you want on linux and it won't affect the other two platforms on those labels.

https://github.com/Chris-AC9KH/JS8Call-improved/commit/2d113da318b0340a7ad32b6e05f4983da8c8e2a8

All that's left to do now is define the rest of UI for the preprocessor, but since I was working on the labels this is a good start. I can PR this if you want to use it.

Joe-K0OG:
> There is no need to muddy up the code to apply QSS StyleSheets. If I design the code the way it should've been designed in the first place, you can add your QSS StyleSheets to the header definitions instead of plastering the code with hooks that make QSS work for linux.

@Chris-AC9KH Are you saying a user will no longer have the ability to apply a .qss stylesheet on an already-built application?  If so, then this is not the way to go.  But maybe I misunderstand this aspect of what you are saying?

73,
-Joe-
K0OG

Joe-K0OG:
> All that's left to do now is define the rest of UI for the preprocessor, but since I was working on the labels this is a good start. I can PR this if you want to use it.

@Chris-AC9KH Again, are you saying that I *can* load a stylesheet at run time on Linux, but *NOT* on Windows?  I would like to be able to do this on Windows too.

73,
-Joe-
K0OG

Chris-AC9KH:
I currently set the preprocessor to only allow USE_QSS_THEMES on linux. We've been forcing the "designed on Linux Windows 95-copy look" on Windows, and that's not right. Windows 11, like MacOS, has a well-defined API that should be adhered to.

At present, with code posted in the link above, this only applies to the labels. On Linux you are free to do anything with them you want with QSS. On Windows and Mac you're not - it's restricted to the native style.

This is example code on how to apply QSS the right way. QSS should NOT be allowed to modify anything in the UI that you want when it already defined by an API. QSS is very inefficient because it requires parsing strings and re-calculating styles every time a widget or label is created, reparented, or refreshed.

So I provided this as an example of how to apply QSS and limit what it can do. It should only be applied to static UI elements like borders, backgrounds and text theming. Never on dynamic UI elements. So this is provided as an example of how to apply it, and how to restrict it.

Joe-K0OG:
@Chris-AC9KH O.k., I think I understand where you are going.  I can still apply a .qss theme at run-time on Windows or Linux to adjust the overall look & feel, but not specific dynamic elements like the buttons, or status-bar (like it is now - statically defined).  Can Mac users apply some generalized theme at run time (whether by .qss or some other Mac tool)?

In terms of general feature philosophy, it would be a nice feature to be able to theme the whole look on all OS variants.  This can already be partially done with existing color/font selections, but the .qss approach was nice because it was more extensive.  I don't want to lose that, in general.

73,
-Joe-
K0OG

Chris-AC9KH:
> @Chris-AC9KH Are you saying a user will no longer have the ability to apply a .qss stylesheet on an already-built application? If so, then this is not the way to go. But maybe I misunderstand this aspect of what you are saying?

Yes, you're not understanding. In current master you cannot modify the labels with QSS. With the code I posted above you can, but only on Linux. I defined this with preprocessor directives so the native styles are compiled into the binary. This makes for more efficient code than a bunch of `if-else` statements and defining color variables for each instance of the UI that is already defined by the preprocessor. So I no longer need all those background colors you were having me put in. They are compiled into the binary. And they can only be overridden on linux with QSS.

Chris-AC9KH:
I'll provide an explanation of what I did.
- first we define the standard background and foreground colors for labels
- we put in a USE_QSS_THEMES for labels that is ONLY valid on linux because that's defined at compile time.
- we define the native styles for each platform. These are well-defined in both Windows 11 and MacOS developer documentation as well as Qt documentation.
- For linux there is no standard.
- we define a default for linux or Android if USE_QSS_THEMES is not set
- this compiles the definitions for those labels into the binary for each platform - no extra code cruft needed.
- the default on linux can be overridden with the example code I put above this code block. On MacOS and Windows your QSS StyleSheet will have no effect on the labels.
- define this for the rest of the UI and you can use QSS on linux if you want to auto-load a StyleSheet at runtime without wrecking the UI on Mac or Windows where users have higher expectations of what should be presented, and UI design is more tightly controlled.

Like I said, I can create a PR and merge this to master since it only affects the labels here. Then you got a template as to how to apply QSS the right way. And it moves defining defaults in the UI to one place instead of being scattered thruout the code.

This doesn't have to be defined in mainwindow.h. If you want, you can create say a "styles.h", put it in includes, and include that header in any source file where you want to define the UI.

```
inline QString statusLabelStyle(const QString& bg = "#6699ff",
                                const QString& fg = "#000000")
{
#if defined(USE_QSS_THEMES)
    return QString();
#else
  #if defined(Q_OS_MACOS)
    return QStringLiteral(
        "QLabel{background-color: %1; color: %2; "
        "border-radius: 8px; padding: 2px 8px; "
        "border: 1px solid rgba(0,0,0,0.15)}"
    ).arg(bg, fg);
  #elif defined(Q_OS_WIN)
    return QStringLiteral(
        "QLabel{background-color:%1; color:%2; "
        "border-radius:4px; padding:2px 8px; "
        "border:1px solid rgba(0,0,0,0.25)}"
    ).arg(bg, fg);
  #else // Linux/other (when USE_QSS_THEMES is not set)
    return QStringLiteral(
        "QLabel{background-color:%1; color:%2; "
        "border-radius:0px; padding:2px 6px; "
        "border:1px inset rgba(0,0,0,0.20)}"
    ).arg(bg, fg);
  #endif
#endif
}
```

Chris-AC9KH:
> Can Mac users apply some generalized theme at run time (whether by .qss or some other Mac tool)?
> 
> In terms of general feature philosophy, it would be a nice feature to be able to theme the whole look on all OS variants. This can already be partially done with existing color/font selections, but the .qss approach was nice because it was more extensive. I don't want to lose that, in general.

Mac has advanced built-in "theming" in the operating system. QSS not needed. The danger with QSS is that it is very inefficient. So it needs to be implemented properly. This is an example of how to do it vs trying to grab elements out of the UI and modify them on-the-fly with QSS by sticking in statements here and there in the code.

Joe-K0OG:
> I can still apply a .qss theme at run-time on Windows or Linux to adjust the overall look & feel, but not specific dynamic elements like the buttons, or status-bar (like it is now - statically defined).

Is the above statement true for Windows users?  Again, I think it would be good for Windows users to be able to do some custom themeing of the overall look and feel (even if dynamic elements are not changeable).

73,
-Joe-
K0OG

Joe-K0OG:
@Chris-AC9KH Saying it another way:  It would be good to be able to "theme" JS8Call without having to theme the whole OS to get there.  We can partially do that with the colors/fonts settings, but the .qss stylesheets idea was nice because it would theme the whole layout.

73,
-Joe-
K0OG

wmiler:
@Joe-K0OG Joe, it is my understanding that Apple forces their users to only use their styling. So full stop, no qss is allowed. Additionally, it is my understanding, that the hardcoded style elements are also ignored by Apple. Therefore, the only OSes under discussion should be Windows and Linux.

Edit: Although, on reflection, that's not totally true, otherwise, pill formatting would fail to work.

Externally loaded QSS has a load/processing time of around 150ms, the way that master currently implements styling as a load/processing time of around 90ms at first load. Converting to the "proper way" as defined by QT6 using `QStyle->polish` and friends `unpolish and update`, would significantly reduce the compile time loading of hardcoded styles to around 10ms, this would make sense for things that change often.

Joe-K0OG:
@wmiler Yes, that all makes sense - that we are really only talking about Linux & Windows.  However, I am a little unclear about some of the comments @Chris-AC9KH made, possibly implying that he wants to remove the .qss stylesheet load option from Windows users, which I would object to.  It certainly makes sense to hard-code several elements so that the loading of .qss stylesheets or OS theme changing won't break their appearance, as we demonstrated with the excellent modifications to the status bar (thanks Chris!).  Of course, whatever is done for Windows & Linux users, we need to make sure it doesn't cause any problems for MacOS, but it seems like Chris is confident that we are safe there.

73,
-Joe-
K0OG

Joe-K0OG:
Another general comment - and surely I'm not alone here - as much as I like the MacOS look, I don't really want that look on my Windows or Linux systems.  I want them to look, as Chris puts it, "...like they came from Win95" -- I actually *like* that look!  Sometimes, old things are just fine...  I say that more and more, the older I get, ha!

73,
-Joe-
K0OG

Chris-AC9KH:
Come to think of it, I'm going to create a styles.h and put it in JS8_Include. There's so many elements of the JS8Call UI that need to be brought up to modern UI standards that it's almost overwhelming. And don't worry, it won't change on Linux because Linux users seem to like that kind of stuff. But it needs to be fixed on MacOS and Windows at the minimum to meet their UI design standards. The UI in this program has not changed since it was invented, and it is a dated design.

You need to start with a baseline, which right now is the new status bar labels. Then expand it to cover the rest of the UI. Once we get the native "look and feel" working for all three platforms, then can decide on definitions for linux, or if QSS is applicable on Windows 11. Even on linux the wayland graphics compositor has a newer "look and feel" compared to the old GTK/Windows 95 style that JS8Call primarily uses. That can be defined for linux too as the default before QSS is applied. Linux has three native platform plugins, wayland, xcb and X11, none of which are being used in the UI.

The bottom line here is that you have to do your homework on the basic UI design BEFORE trying to "fix it" with QSS StyleSheets. The QSS StyleSheets, while the linux users will find it fun because that's what they spend a lot of their time on, is putting the cart before the horse without fixing the basic UI problems first to get rid of all the hardcoded stuff in it that overrides native styles.

You can try to come up with a "default" style sheet, but how accurate is it when you're started from a dated design that needs to be brought up to modern standards? Why not fix the UI first? Then you can declare one StyleSheet for that source file that is valid on platforms it can be used on, others default to the native style instead of hardcoded stuff that looks like it was copied from Windows 95.

That's the purpose of the code I posted in the link above - to show how to do it starting at the bottom of the UI with the status bar that I just overhauled. You compile that code and you will get three different styles on the three platforms. But, unfortunately, we should be using the wayland style on linux instead of the old GTK/Windows 95 style by default. But there's no consensus among linux users as to how to even turn the computer on, much less what it should look like, so what you see is what you get on linux. If I PR that code to master it's up to you to define it for linux.

Joe-K0OG:
@Chris-AC9KH All that sounds good (I think!).  Yes it would be good to have proper OS-appropriate look and feel by default.

But please answer my question with a simple yes or no:  On *both* Window and Linux will I be able to apply a custom theme at run time? 

I would personally like that feature, even it it's somewhat limited (like it is with the status bar now).  A possible solution would be to have a built-in "Light" and "Dark" application theme, even if I can't apply a .qss style sheet.  Again, I would like to be able to somewhat theme (especially "Light" vs. "Dark") the application without themeing the whole OS.

73,
-Joe-
K0OG

P.S.  I don't run Wayland on Linux Mint 22.3 as it's not integrated there yet - but I'm guessing eventually will.

Chris-AC9KH:
> @wmiler Yes, that all makes sense - that we are really only talking about Linux & Windows. However, I am a little unclear about some of the comments @Chris-AC9KH made, possibly implying that he wants to remove the .qss stylesheet load option from Windows users, which I would object to.

@Joe-K0OG no, you are misunderstanding what the code I posted does. That particular code does NOT prevent you from loading a StyleSheet on Windows. It only prevents you from using a StyleSheet to change the native appearance of the Status Bar on Windows 11. Those are dynamically updated UI elements, so I coded it to let linux users break it, but it's locked in on Windows and Mac.

The code I posted is to show you how to do that to prevent QSS from causing the UI to be unusable on the two platforms where UI design is more tightly controlled. If you don't want to use it, then fine, we'll move on. You'll have to figure out how to put your QSS hooks into the Status Bar on your own. But what I posted is the proper way to implement QSS, with a demonstration of how you put controls on what it is allowed to do.

You're also misunderstanding wayland. It is now the standard graphics compositor on linux, replacing X11. Any definitions of platform-specific styles for the preprocessor should be using the wayland style. But like I said, getting any sort of consensus from linux users is impossible.

Chris-AC9KH:
> Edit: Although, on reflection, that's not totally true, otherwise, pill formatting would fail to work.

@wmiler, the pill formatting of the MSG composition is a MacOS style. That can be defined in Qt to make applicable across all platforms which @Beiri22 has done.

What I provided is a template that can be used to apply MacOS style, Windows 11 style, or no style for linux to the Status Bar. And allow you to modify it on linux with QSS. It eliminates extra cruft in the code for defining status labels, and is more efficient because it's compiled into the binary by the preprocessor. That's why I included a `#pragma once` for that header so it can be only used once at compile time, with the intention that it should be moved to a header that defines native styles.

Joe-K0OG:
> You're also misunderstanding wayland. It is now the standard graphics compositor on linux, replacing X11. Any definitions of platform-specific styles for the preprocessor should be using the wayland style. But like I said, getting any sort of consensus from linux users is impossible.

I'm not misunderstanding Wayland, I'm saying that Linux Mint 22.3 which I primarily run doesn't implement it (yet) - but it is coming sometime...  I don't know when.  My Mint box still runs X11, which I'm fine with.  No big deal either way. 

I'm completely not concerned with performance, since I run a very old (over 15-years-old) laptop with a dual-core processor, 6GB RAM, and even with loading extensive .qss stylesheets performance is very good, using very little CPU and memory resources.  This improved dramatically with the improvements Allen made to the code.  My Mint system is significantly lower spec that many RPi systems out there, and it runs great.

73,
-Joe-
K0OG

wmiler:
Wayland isn't standard yet, it will be in the next year or so (Ubuntu 26.04 releases this month). But again, most hams aren't using the most recent version of the OS in the hamshack, and with the stuff M$ is pulling with Win11, I don't see it changing fast.

Also, once again, I will point out that this stylesheet, strictly pulls out the "addresses" of what is already in the existing UI files that get compiled into the binary. I'm not going to point that out anymore.

Joe-K0OG:
Just a little tidbit:  I found in playing around with .qss stylesheets, it is convenient to have a "clear.qss" file, which is a blank file, such that when I load it, the screen style reverts to default, if I don't want a style I previously applied.  That way I don't have to close/re-open JS8Call to revert.  Maybe it would be good to have a "Clear Style" button there that does the same thing?

73,
-Joe-
K0OG

Chris-AC9KH:
> Also, once again, I will point out that this stylesheet, strictly pulls out the "addresses" of what is already in the existing UI files that get compiled into the binary. I'm not going to point that out anymore.

You don't need to point that out. What I'm pointing out is that if the UI gets updated to more modern standards, your QSS calls are likely not going to work because functions() are used to define them. So I'm providing you with a template as to how to do that to try to prevent something from being designed on legacy, and that will have future maintainability.

wmiler:
> > Also, once again, I will point out that this stylesheet, strictly pulls out the "addresses" of what is already in the existing UI files that get compiled into the binary. I'm not going to point that out anymore.
> 
> You don't need to point that out. What I'm pointing out is that if the UI gets updated to more modern standards, your QSS calls are likely not going to work because functions() are used to define them. So I'm providing you with a template as to how to do that to try to prevent something from being designed on legacy, and that will have future maintainability.

Ok, I'll ponder on that a bit. I think we are all in the same general ballpark, but not quite making the connection. There's one thing to be seeking a standard, but that standard is NOT to be MacOS on all platforms, just as Win11 is not on all platforms, etc. 

The only way the QSS would cease to be backward compatible, is if things get changed in the UI files.

Moving on, since I really didn't understand what you were trying to accomplish Chris, in the status bar PR, your intention is to change the "look and feel" to our own JS8Call vs what came from WSJT-X? Is that correct?

Chris-AC9KH:
I'll say this one last time, I'm not going to say it again. Implementing QSS is more than just grabbing stuff that already exists in the code because that changes and is a bad design. I'm trying to show you how to implement QSS in a way that makes it long-term maintainable. Which will require making some changes in the UI in how various elements are defined.

I would like to move to a styles.h header where there's at least a template as to how to maintain it in the future, and PR that for the freshly overhauled Status Bar.  If not interested I'll move on, I got better things to do.

wmiler:
> Just a little tidbit: I found in playing around with .qss stylesheets, it is convenient to have a "clear.qss" file, which is a blank file, such that when I load it, the screen style reverts to default, if I don't want a style I previously applied. That way I don't have to close/re-open JS8Call to revert. Maybe it would be good to have a "Clear Style" button there that does the same thing?
> 
> 73, -Joe- K0OG

That can be added, but I think we need to figure this UI thing out first, so we all have a clear understanding of goals and intentions.

Chris-AC9KH:
> in the status bar PR, your intention is to change the "look and feel" to our own JS8Call vs what came from WSJT-X? Is that correct?

JS8Call does not have a "look and feel". It's a hodge-podge of source file hardcoded and ui-> generated that does not correctly implement platform-specific styles. The new Status Bar is the first UI element in JS8Call that does. You can compile the code on the various platforms from the branch in my private repo to get an understanding of how it works.

Joe-K0OG:
> That can be added, but I think we need to figure this UI thing out first, so we all have a clear understanding of goals and intentions.

Yeah @wmiler, I didn't expect it right now - what @Chris-AC9KH is doing needs to get done first as you said.  I just wanted to document the idea of clearing a style while I was thinking about it -- I'll likely forget about it later, ha!

Joe-K0OG:
> I would like to move to a styles.h header where there's at least a template as to how to maintain it in the future, and PR that for the freshly overhauled Status Bar. If not interested I'll move on, I got better things to do.

@Chris-AC9KH If you are willing, I say go for it.  If not, it's acceptable as it is.  If future maintenance can be improved, that's certainly worthwhile.

73,
-Joe-
K0OG

Chris-AC9KH:
> your intention is to change the "look and feel" to our own JS8Call vs what came from WSJT-X? Is that correct?

The JS8Call Status Bar is not the same as WSJT-x. The goal is to use the native platform UI by default, which is not being done in JS8Call. The freshly overhauled Status Bar is the first UI element in JS8Call that uses native platform styles.

When I get time later I'll fix up that branch to create a new styles.h header and PR it. Because I don't think anybody here understands how it works. Then you can compile the code on the three platforms and you will see that the MacOS style is not the "standard". You will get MacOS style on Mac. You will get Windows 11 style on Windows. And you will get the old squared off GTK/Windows 95 style on Linux, which doesn't have a style.

This is the very first time, in just one UI element, that native styles are applied to that element. Nobody seems to understand how to do it so we're running thru the code assigning foreground colors to labels for theming issues on linux. This can be fixed or changed in one place if the code was designed right.

If you don't like the PR when it appears, then scrap it. But I hope it gives people a better understanding of what's wrong in certain areas of the UI code.

Joe-K0OG:
> This is the very first time, in just one UI element, that native styles are applied to that element. Nobody seems to understand how to do it so we're running thru the code assigning foreground colors to labels for theming issues on linux. This can be fixed or changed in one place if the code was designed right.

Looks like a good idea to me.  The way it worked out with the status bar is good I think.

73,
-Joe-
K0OG

Chris-AC9KH:
> Wayland isn't standard yet, it will be in the next year or so

Again, there's a misunderstand of what's going on here. I'm not coding JS8Call to use wayland, ok? I'm pointing out that wayland has a style that uses a very slightly rounded corner on QLabels and it uses a different more attractive border than the old deprecated Sunken border style that was hardcoded into JS8Call and doesn't even exist anymore in Qt 6.11 - the code won't compile.

These styles are defined per platform. And since the wayland style is more visually appealing than the old squared-off GTK/Win95 style, it is what I would choose for the Status Bar labels as the default on linux with no QSS applied. But we definitely don't want to send some linux user into a coma if wayland-style labels suddenly appear in JS8Call.

Chris-AC9KH:
> Looks like a good idea to me. The way it worked out with the status bar is good I think.

Ha! your "Fusion" fix does not work on MacOS. I just fixed the progress bar on MacOS 26, which hasn't worked since it was released. The new definitions are in styles.h and I used an inline preprocessor directive in the mainwindow.cpp code to compile the progress bar type specific-by platform.

wmiler:
I'm going to self-mark this as on hold until Discussion #251 New UI completes. No sense making changes while backend code is changing.

Chris-AC9KH:
@wmiler this should be fully functional. In the final implementation of syles.h I removed the example restriction on QSS.

The only caveat, as I noted, is where you declare your qApp->setStyleSheet(). It must be done by the compiler after the declaration in QApplication in main.cpp (same for any Qt project). And any style sheet MUST include definitions for anything defined in styles.h or no style will be applied. That is basically the only change. Since Linux is untouched by that, other than the status bar labels and progress bar, this should be fairly straight forward on linux. For Windows a qss style sheet will have to include definitions for things defined for Windows in styles.h to make a usable "default" style sheet.

I suspect this will be more of a linux thing. Very few Windows users will probably mess with it. Mac won't use it at all.

Chris-AC9KH:
@wmiler I took a glance at your code and it looks like you are declared in JS8_UI Configuration.cpp. So all should be good here. The object code is compiled after main.cpp in the compile order in CMakeLists.txt. I took care to make the styles.h declaration in the QApplication method in main.cpp, which is the very first thing that gets compiled that defines the app UI and API, to ensure this wouldn't be broken. And it looks fine to me for runtime injection of QSS.

Chris-AC9KH:
I should mention that are some things in mainwindow.ui that are VERY inconsistent. When designing the new MacOS UI I ran across a bug in there that I fixed - not pushed to any branch yet.

But the important thing to realize is that stuff that is hardcoded in mainwindow.ui gets overridden by declared styles. The hardcoded stuff in there ONLY applies to a default configuration where no custom styles are defined. So as I run into this stuff that has been mis-configured in mainwindow.ui I'll fix them as found. But it's not the end of the world.

wmiler:
> I should mention that are some things in mainwindow.ui that are VERY inconsistent. When designing the new MacOS UI I ran across a bug in there that I fixed - not pushed to any branch yet.
> 
> But the important thing to realize is that stuff that is hardcoded in mainwindow.ui gets overridden by declared styles. The hardcoded stuff in there ONLY applies to a default configuration where no custom styles are defined. So as I run into this stuff that has been mis-configured in mainwindow.ui I'll fix them as found. But it's not the end of the world.

That's fine, the style file is mainly just copying out of the ui files, so peeps would know how to address locations. I'll do a fixup later when you are happy with your stuffs.

Joe-K0OG:
> That's fine, the style file is mainly just copying out of the ui files, **so peeps would know how to address location**s. I'll do a fixup later when you are happy with your stuffs.

@wmiler Yes, Wyatt, this is what I need to successfully do further experimentation with stylesheets.  I've been occasionally hacking away at it, but somewhat blindly.

73,
-Joe-
K0OG

Chris-AC9KH:
@wmiler @Joe-K0OG This should now be really easy to finish. The things you need to theme the header bar with runtime qss injection are listed in UI_Constructor.cpp:

```
    ui->frame->setStyleSheet(logFrameStyle());
    ui->logWidget->setStyleSheet(Styles::LogWidgetStyle);
    ui->dialFreqUpButton->setStyleSheet(Styles::DialFreqUpDownButtonStyle);
    ui->dialFreqDownButton->setStyleSheet(Styles::DialFreqUpDownButtonStyle);
    ui->labDialFreqOffset->setStyleSheet(Styles::LabDialFreqOffsetStyle);
    ui->labCallsign->setStyleSheet(Styles::LabCallsignStyle);
    ui->labUTC->setStyleSheet(Styles::LabUTCStyle);
    ui->buttonGrid->setStyleSheet(Styles::ButtonGridStyle);
    ui->monitorTxButton->setStyleSheet(Styles::MonitorTxButtonStyle);
    ui->monitorButton->setStyleSheet(Styles::MonitorButtonStyle);
    ui->logQSOButton->setStyleSheet(Styles::LogQSOButtonStyle);
    ui->tuneButton->setStyleSheet(Styles::TuneButtonStyle);
    ui->modeButton->setStyleSheet(Styles::ModeButtonStyle);
    ui->spotButton->setStyleSheet(Styles::SpotButtonStyle);
```

And the default QSS definitions for those are in styles.h

The frame is the Frequency display. The one Joe was having a problem with is logWidget - that is the parent widget for all the rest and is the background color for the entire thing.

I'm not sure why all those have such weird names? I took them direct from the .ui file. But styles.h is well commented so you can figure it out.

The linux guys will have to submit a PR for the default linux style yet. It is mostly not defined in styles.h. For instance, @rruchte had white text on the header bar control buttons, but the QPushButton style is not defined for linux in styles.h. There's a suggested configuration for it but he was trying to change the background color of the button grid instead. So somebody needs to figure out the defaults for linux and finish defining their default style in styles.h

Chris-AC9KH:
Also, there is no font for the frequency display on linux. MacOS uses Monaco, Windows uses Consolas. I did not define one for linux because whatever font I would pick would be wrong. So the frequency display is not going to be same from one linux flavor to the next until somebody defines that stuff. I did make one attempt at defining a font for the date/time display, but I don't know if it's right or not.

Joe-K0OG:
@Chris-AC9KH Excellent work Chris - thanks for doing it so quickly, but don't burn yourself out!

On Windows 11 24H2 and 25H2, it runs fine in both light and dark system themes.  I don't see any problems there.  It built without any problems in JTSDK64-Tools.

On Linux Mint 22.3, the following build warning was flagged:
```
[ 99%] Building CXX object CMakeFiles/JS8Call.dir/JS8_Widgets/LazyFillComboBox.cpp.o
/js8call/js8call-prefix/src/JS8_UI/mainwindow.cpp: In member function ‘void UI_Constructor::createStatusBar()’:
/js8call/js8call-prefix/src/JS8_UI/mainwindow.cpp:1451:18: warning: variable ‘textColor’ set but not used [-Wunused-but-set-variable]
      |                  ^~~~~~~~~
[100%] Linking CXX executable JS8Call
[100%] Built target JS8Call
```

This did not cause any problems, and here is the result on a light system theme:
image

This is excellent, no problem with it.

However, with a dark system theme on Linux, the button text colors are white as was mentioned, so we will need to recommend something for that.  I suggest that we just set those to black all the time since on Linux the button colors don't change much with system theme, and the background remains the same.
image

Big progress here - great work Chris!

73,
-Joe-
K0OG

Joe-K0OG:
Just as an example to see where we are with applying a .qss stylesheet file during runtime, here is a dark .qss file I applied:
image

With a few tweaks to the .qss stylesheet, this can probably be made to work well.

73,
-Joe-
K0OG

Joe-K0OG:
> The linux guys will have to submit a PR for the default linux style yet. It is mostly not defined in styles.h. For instance, @rruchte had white text on the header bar control buttons, but the QPushButton style is not defined for linux in styles.h. There's a suggested configuration for it but he was trying to change the background color of the button grid instead. So somebody needs to figure out the defaults for linux and finish defining their default style in styles.h

As I mentioned earlier, since whether on light or dark themes, at least on Linux Mint, the button colors don't change much, I suggest that the top bar button text be set as black for all states, by default, so that when running a dark system theme where the text turns white, it will remain readable as black.  I just edited styles.h and built it on Linux, testing it on both light and dark themes, and with .qss stylesheets applied at run-time.  If you guys want to try my modified styles.h on Linux, here it is attached.
[styles.h.txt](https://github.com/user-attachments/files/26791076/styles.h.txt)

I think this is a reasonable solution for default settings.

73,
-Joe-
K0OG

rruchte:
@Joe-K0OG That was my assessment as well. Default text color styles falling through to those buttons makes no sense.

Joe-K0OG:
> @Joe-K0OG That was my assessment as well. Default text color styles falling through to those buttons makes no sense.

My result with a dark system theme on Linux Mint:
image

And I tested this will all the button presses, and the text is always nice and legible.

Shall we push this and be done with it?

73,
-Joe-
K0OG

rruchte:
Looks good to me.

Chris-AC9KH:
@Joe-K0OG right here is where you set your button text color and it's commented out. I didn't define anything for linux. But if you want black text un-comment the preprocessor directives and set your color to what you want. You don't have to define the rest if you don't want. If you look above this section you'll see Mac and Windows are set to black.

```
  // Suggested possibility for linux - not defined at this time
    // #elif defined(Q_OS_LINUX)
    //    return R"(
    //        QPushButton {
    //            background-color: #E0E0E0;
    //            color: #212121;
    //            border: 1px solid #BDBDBD;
    //            border-radius: 1px;
    //            padding: 5px 15px;
    //            font-family: "Ubuntu", "Noto Sans";
```

wmiler:
> Just as an example to see where we are with applying a .qss stylesheet file during runtime, here is a dark .qss file I applied: 
> 
> With a few tweaks to the .qss stylesheet, this can probably be made to work well.
> 
> 73, -Joe- K0OG

Yeah I see those boxes with the qss file on this pr as well. That's a "deal with it later" sort of thing :)

wmiler:
> @Joe-K0OG right here is where you set your button text color and it's commented out. I didn't define anything for linux. But if you want black text just comment out the preprocessor directives and set your color to what you want. You don't have to define the rest if you don't want. If you look above this section you'll see Mac and Windows are set to black.
> 

See the other PR #257

wmiler:
This is next on the list for me after I finish clearing repo security issues. So maybe this weekend...

wmiler:
Moved to the 3.1.0 milestone. Still no free time at home to work on this.

wmiler:
I should have mentioned on the close that I'm thinking of just providing this on the website vs including it in the code base.

Chris-AC9KH:
I actually think the website would be better. We already have the functionality to load stylesheets. Users just need a place where they can contribute their own, download ones that others have created (if we want to offer that), and maybe documentation on how to create one.

That could all be implemented pretty easily on the website by adding either a tab in the header to it, or maybe inline links to it, or maybe even a blurb in the User Guide on it.

wmiler:
Yeah, User Guide is already slated for some major updates including breaking it up (chapters/sections) for better readability. See #285 and add any thoughts there please.

Seems like @rruchte Rob was supposed to be doing something with the web site, did that get completed? If so, I'll submit this as a PR over there.

Still recovering from the 80+ hour work week I just came off of, oddly 20 years ago I recovered faster, lol!

rruchte:
@wmiler My changes are complete and merged.

wmiler:
> @wmiler My changes are complete and merged.

Ok, thanks.

PR #249, opened 2026-04-10
Use preprocessor directives to apply platform-specific styles

Chris-AC9KH:
- allow QSS theme injection on linux for status bar labels
- remove color definitions from labels that are already defined by the preprocessor
- fix progress bar on MacOS 26 which uses a different style from other platforms

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,6 +230,14 @@ configure_file(
   "${CMAKE_CURRENT_BINARY_DIR}/js8call_config.h"
 )
 
+#------------------------------------------------------------------------------#
+# On Linux we allow QSS theme injection
+#------------------------------------------------------------------------------#
+if (UNIX AND NOT APPLE)
+      target_compile_definitions(
+      ${TARGET} PRIVATE USE_QSS_THEMES)
+endif()
+
 target_precompile_headers(
   ${TARGET} PRIVATE
   ${CMAKE_CURRENT_BINARY_DIR}/js8call_config.h
--- a/JS8_UI/mainwindow.cpp
+++ b/JS8_UI/mainwindow.cpp
@@ -6,6 +6,7 @@
  */
 
 #include "mainwindow.h"
+#include "styles.h"
 
 #include "moc_mainwindow.cpp"
 
@@ -1389,27 +1390,24 @@ void UI_Constructor::createStatusBar() // createStatusBar
 {
     tx_status_label.setAlignment(Qt::AlignCenter);
     tx_status_label.setMinimumSize(QSize{150, 18});
-    tx_status_label.setStyleSheet("QLabel{background-color: #22ff22; color: #000000}");
-    tx_status_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    tx_status_label.setStyleSheet(txStatusLabelStyle(TxStatusAppearance::Receiving));
     statusBar()->addWidget(&tx_status_label);
 
     last_tx_label.setAlignment(Qt::AlignCenter);
     last_tx_label.setMinimumSize(QSize{150, 18});
-    last_tx_label.setStyleSheet("QLabel{background-color: #6699ff; color: #000000}");
-    last_tx_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    last_tx_label.setStyleSheet(statusLabelStyle());
     statusBar()->addWidget(&last_tx_label);
 
     config_label.setAlignment(Qt::AlignCenter);
     config_label.setMinimumSize(QSize{80, 18});
-    config_label.setStyleSheet("QLabel{background-color: #6699ff; color: #000000}");
-    config_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    config_label.setStyleSheet(statusLabelStyle());
     statusBar()->addWidget(&config_label);
     config_label.hide(); // only shown for non-default configuration
 
     mode_label.setAlignment(Qt::AlignCenter);
     mode_label.setMinimumSize(QSize{80, 18});
-    mode_label.setStyleSheet("QLabel{background-color: #6699ff; color: #000000}");
-    mode_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    mode_label.setStyleSheet(statusLabelStyle());
+
     {
         QString modeLabelText;
         switch (m_nSubMode) {
@@ -1438,32 +1436,26 @@ void UI_Constructor::createStatusBar() // createStatusBar
 
     frequency_label.setAlignment(Qt::AlignCenter);
     frequency_label.setMinimumSize(QSize{110, 18});
-    frequency_label.setStyleSheet("QLabel{background-color: #6699ff; color: #000000}");
-    frequency_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    frequency_label.setStyleSheet(statusLabelStyle());
     frequency_label.setText(QString("Freq: %1").arg(Radio::pretty_frequency_MHz_string(dialFrequency() + freq())));
     statusBar()->addWidget(&frequency_label);
     
     auto_reply_label.setAlignment(Qt::AlignCenter);
     auto_reply_label.setMinimumSize(QSize{110, 18});
-    auto_reply_label.setStyleSheet("QLabel{background-color: #6699ff; color: #000000}");
-    auto_reply_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    auto_reply_label.setStyleSheet(statusLabelStyle());
     QString autoReplyState = ui->actionModeAutoreply->isChecked() ? "On" : "Off";
     statusBar()->addWidget(&auto_reply_label);
 
     statusBar()->addPermanentWidget(&progressBar);
     progressBar.setMinimumSize(QSize{100, 18});
-    if (QStyle *fusion = QStyleFactory::create("Fusion")) {
-        progressBar.setStyle(fusion);
-    }
+    const bool small = true;
     const QColor textColor = this->palette().color(QPalette::WindowText);
-    progressBar.setStyleSheet(QString("QProgressBar { color: %1; text-align: center; }")
-                                  .arg(textColor.name()));
+    progressBar.setStyleSheet(progress_bar_stylesheet(textColor, small));
     progressBar.setFormat("%v/%m");
 
     statusBar()->addPermanentWidget(&wpm_label);
     wpm_label.setMinimumSize(QSize{120, 18});
-    wpm_label.setStyleSheet("QLabel{background-color: #6699ff; color: #000000}");
-    wpm_label.setFrameStyle(QFrame::Panel | QFrame::Sunken);
+    wpm_label.setStyleSheet(statusLabelStyle());
     wpm_label.setAlignment(Qt::AlignCenter);
 }
 
@@ -2734,8 +2726,8 @@ void UI_Constructor::guiUpdate() {
         }
 
         if (m_transmitting) {
-            tx_status_label.setStyleSheet(
-                "QLabel{background-color: #ff2222; color:#000}");
+            tx_status_label.setStyleSheet(txStatusLabelStyle(TxStatusAppearance::Transmitting));
+
             if (m_tune) {
                 tx_status_label.setText("Tx: TUNE");
             } else {
@@ -2747,12 +2739,10 @@ void UI_Constructor::guiUpdate() {
             transmitDisplay(true);
         } else if (m_monitoring) {
             if (m_tx_watchdog) {
-                tx_status_label.setStyleSheet(
-                    "QLabel{background-color: #000; color:#fff}");
+                tx_status_label.setStyleSheet(txStatusLabelStyle(TxStatusAppearance::IdleTimeout));
                 tx_status_label.setText("Idle timeout");
             } else {
-                tx_status_label.setStyleSheet(
-                    "QLabel{background-color: #22ff22; color: #000}");
+                tx_status_label.setStyleSheet(txStatusLabelStyle(TxStatusAppearance::Decoding));
                 tx_status_label.setText(m_decoderBusy ? "Decoding"
                                                       : "Receiving");
             }
@@ -6953,8 +6943,9 @@ void UI_Constructor::tx_watchdog(bool triggered) {
         if (m_auto)
             auto_tx_mode(false);
         stopTx();
-        tx_status_label.setStyleSheet(
-            "QLabel{background-color: #000000; color:#ffffff; }");
+        {
+            tx_status_label.setStyleSheet(txStatusLabelStyle(TxStatusAppearance::IdleTimeout));
+        }
         tx_status_label.setText("Idle timeout");
 
         // if the watchdog is triggered...we're no longer active
--- a/JS8_UI/styles.h
+++ b/JS8_UI/styles.h
@@ -1,0 +1,131 @@
+/**
+ * @file styles.h
+ * @brief header file that defines platform specific label, button, widget and dialog
+ * styles used in functions in the UI_Constructor class that assembles the UI and provides
+ * the proper "look and feel" for each desktop platform
+ *
+ * styles.h was added on 11 Apr, 2026 and is intended to eventually become the default
+ * StyleSheet configuration for the JS8Call user interface.
+ */
+
+#pragma once
+#include <QtGlobal>
+#include <QtGui/QGuiApplication>
+#include <QtCore/QOperatingSystemVersion>
+#include <QtCore/QString>
+
+inline QString statusLabelStyle(const QString& bg = "#6699ff",
+                                const QString& fg = "#000000")
+{
+#if defined(USE_QSS_THEMES)
+    // If a global application stylesheet is active, defer to it and don't override.
+    if (qApp && !qApp->styleSheet().isEmpty()) {
+        return QString();
+    }
+#endif
+
+// Status bar label StyleSheet
+#if defined(Q_OS_MACOS)
+    return QStringLiteral(
+        "QLabel{background-color: %1; color: %2; "
+        "border-radius: 8px; padding: 2px 8px; "
+        "border: 1px solid rgba(0,0,0,0.15)}"
+    ).arg(bg, fg);
+#elif defined(Q_OS_WIN)
+    return QStringLiteral(
+        "QLabel{background-color:%1; color:%2; "
+        "border-radius:4px; padding:0px 8px; "
+        "border:1px solid rgba(0,0,0,0.25)}"
+    ).arg(bg, fg);
+#else // Linux/other
+    return QStringLiteral(
+        "QLabel{background-color:%1; color:%2; "
+        "border-radius:2px; padding:1px 6px; "
+        "border:1px inset rgba(0,0,0,0.18)}"
+    ).arg(bg, fg);
+#endif
+}
+
+// We need special handling for tx_status_label since it technically has four states,
+// however, Receiving and Decoding are the presently the same colors. But this allows
+// Decoding to be set differently if we choose to do so
+enum class TxStatusAppearance {
+    Receiving,    // green on black text
+    Transmitting, // red
+    Decoding,     // same as Receiving
+    IdleTimeout   // black bg, white fg
+};
+
+QString txStatusLabelStyle(TxStatusAppearance appearance);
+
+static inline QString makeStyle(const QString& bg, const QString& fg) {
+#if defined(Q_OS_LINUX)
+    return QString(
+        "QLabel{background-color:%1; color:%2; "
+        "border-radius:2px; padding:1px 6px; "
+        "border:1px inset rgba(0,0,0,0.18);}"
+    ).arg(bg, fg);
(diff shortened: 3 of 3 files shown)


Chris-AC9KH:
This creates a new styles.h header that is in JS8_UI since it is only used for UI styles definitions - currently defines the status bar labels and the progress bar style. Also fixes the progress bar, which hasn't worked on MacOS 26 since it was released.

Applies native styles to all three platforms for status bar labels. If wayland-style labels are desired for linux they can now be defined in styles.h. The styles are compiled into the binary using preprocessor directives but can be overridden on linux with QSS.

Joe-K0OG:
@Chris-AC9KH Built & tested on Windows.  The boxes need a little more vertical space.  I personally prefer the "clunky square boxes" - but I guess it will be this way, I got no say...

The descenders on the text are clipped (look at "q" "g" and "y" descenders):
image

73,
-Joe-
K0OG

wmiler:
Oh, ok, I think I figured out where all the confusion came from. It was never **my** intention to externally stylize last_tx_label and friends via qss. I just said it could be done. Well, what you have done here makes a lot more sense now.

Chris-AC9KH:
> @Chris-AC9KH Built & tested on Windows. The boxes need a little more vertical space. I personally prefer the "clunky square boxes" - but I guess it will be this way, I got no say...

That is the native Windows 11 style. I'll look at the font issue this evening. That doesn't look like the native Windows Segoe UI-Variable font that automatically scales to different widget and button sizes and screen sizes. If I have to I'll reduce the padding on the labels.

Edit: if you want the old square GTK/Windows 95-style labels go to linux, that's what you'll get.

Chris-AC9KH:
> Oh, ok, I think I figured out where all the confusion came from. It was never **my** intention to externally stylize last_tx_label and friends via qss. I just said it could be done. Well, what you have done here makes a lot more sense now.

It can be done on linux. I just locked it down for Mac and Windows as an example of how to control what QSS is allowed to do. Especially on dynamic UI elements, QSS is not a good idea.

Joe-K0OG:
> That is the native Windows 11 style. I'll look at the font issue this evening. That doesn't look like the native Windows Segoe UI-Variable font that automatically scales to different widget and button sizes and screen sizes. If I have to I'll reduce the padding on the labels.

@Chris-AC9KH Yes, it is the default Windows Segoe font.  This is a fresh settings, so it's all default:
image

73,
-Joe-
K0OG

Joe-K0OG:
F.Y.I.  Here is what it looks with default fresh settings on Windows, Dark app mode (no .qss has been applied):
image

73,
-Joe-
K0OG

Joe-K0OG:
Here is how it looks on Linux (clean, no .qss applied), light system theme:
image

...and on Linux with a dark system theme:
image

As you can see, the fields in the status bar have no background definition to them.  Can a default background be added for Linux?

73,
-Joe-
K0OG

Chris-AC9KH:
So Windows does not do as good of a job with their auto-scaling fonts as MacOS does. With the default Apple System UI 13 font it scales perfectly

Screenshot 2026-04-10 at 17 39 52

Windows doesn't with their default UI font. I had to decrease the vertical padding to 1 pixel for Windows and could probably decrease it to 0 pixels.

Screenshot 2026-04-10 at 17 40 38

Chris-AC9KH:
> As you can see, the fields in the status bar have no background definition to them. Can a default background be added for Linux?

On Linux it was left up to QSS to theme your status bar because USE_QSS_THEMES is defined in CMakeLists.txt for linux. So statusLabelStyle() returns an empty StyleSheet for linux unless you define it with a .qss file. I had understood that's what the linux folks want.

Chris-AC9KH:
@Joe-K0OG OTOH, if you want the status bar to be defined on linux with no .qss loaded I can make it do that too. But there was no consensus on whether we're using the old squared-off GTK/Win95 style, or the more modern wayland style for the labels. So I didn't go there.

Maybe @wmiler has input on this as well if you want a default, or if you want to use .qss to theme your status bar. And if you want a default, which one is it? And do you want me to provide a way to override the default with QSS? Or do you prefer to not theme the status bar with QSS?

I believe I told you that on linux you're going to get what you get since those status bar labels were listed in the other thread as not having QSS hooks. So I gave you a way to puts hooks in them to define them with QSS.

wmiler:
> @Joe-K0OG OTOH, if you want the status bar to be defined on linux with no .qss loaded I can make it do that too. But there was no consensus on whether we're using the old squared-off GTK/Win95 style, or the more modern wayland style for the labels. So I didn't go there.
> 
> Maybe @wmiler has input on this as well if you want a default, or if you want to use .qss to theme your status bar. And if you want a default, which one is it? And do you want me to provide a way to override the default with QSS? Or do you prefer to not theme the status bar with QSS?
> 
> I believe I told you that on linux you're going to get what you get since those status bar labels were listed in the other thread as not having QSS hooks. So I gave you a way to puts hooks in them to define them with QSS.

Everything should have a default set, even if it was the original default (ie from 2.3.2). 

The QSS file is only if people wanted to change things after the fact. Most Linux users are ok with the defaults, until they need to adjust things for say, ADA issues.

Chris-AC9KH:
OK, I'll set a default for linux. Which one do you want?  IF you select a default, make sure there's a consensus on it. I don't want to hear later from some linux user that there's a bug in the UI because we got nice MacOS-style "pills" for MSG composition. But why do we have ugly square buttons and labels from the 1990's when MacOS and Windows got modern ones and linux is using that now for wayland? And you have know that's what will happen. That's why I didn't go there in the first place.

Chris-AC9KH:
I will add that it's not recommended to use the old Sunken Qt style because it's deprecated. It needs to be defined the new "Qt way" and I can duplicate the old Sunken Qt style with the more modern Qt code. I just need to know if we're using old style or wayland style?

Chris-AC9KH:
This is an example of wayland style panes, buttons or labels. Buttons and labels have 2 pixel rounded corners. Panes have 2 pixel rounding at the top, square corners on the bottom.

Screenshot_20241105_181425

Joe-K0OG:
@Chris-AC9KH @wmiler Yes, I think there should be a default that can be styled for Linux.  I'm fine with either what we designed earlier, or with Wayland style.  Wayland style might make more sense in the long run, since that will eventually dominate, and I'm fine with that.  Just make sure that the default will work with either light or dark system theme.

73,
-Joe-
K0OG

Joe-K0OG:
@Chris-AC9KH It's closer, but the text still needs a little more room on Windows.  The descenders are clipped by a pixel or two still:
image

73,
-Joe-
K0OG

Joe-K0OG:
> This is an example of wayland style panes, buttons or labels. Buttons and labels have 2 pixel rounded corners. Panes have 2 pixel rounding at the top, square corners on the bottom.

@Chris-AC9KH If that's the standard, that certainly looks fine to me.

Chris-AC9KH:
@Joe-K0OG on the next push I reduced the vertical padding to 0px for Windows.

Ok, I'll code up wayland style labels for the default for the next push.

Joe-K0OG (commented):
Looking good now!  Win11 (Light system theme):
Image

Win11 (Dark system theme):
Image

The only very minor thing is that the progress bar darkness is slightly too dark when it covers the text, so could be a slightly lighter shade.  However, that might just be my eyes & screen settings.  I can still read the text, so it's not too bad as is.

The Linux Wayland status bar style looks good now on both light and dark system themes.  Should the Decoding/Receiving box have a default color fill like it does on Windows, and like the other boxes do?  
Image

Thanks @Chris-AC9KH for the good work here!  I busted open a can of worms when I asked for something to be done with the status bar - I had no idea!  I'm glad you tackled this, those worms needed to be released!  ;)

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG the progress bar is not defined for color. It uses palette(base) for QProgressBar which is system dependent. On MacOS, for instance, that status bar does not change color in Dark Mode and the bar itself is always blue. Little bit more attention to detail on MacOS with graphics rendering.

Screenshot 2026-04-10 at 20 41 27

Screenshot 2026-04-10 at 20 40 56

Windows 11, according to Microsoft's own documentation, has known issues in Dark Mode. And instead of fixing it, they got 20 pages of propaganda on how to work around it. So I'll look at defining a set color for the bar.

You might know that linux does not obey the stylesheet for the tx_status_label, but Mac and Windows do. I'm attempting to come up with a workaround for that one.

Joe-K0OG:
> Windows 11, according to Microsoft's own documentation, has known issues in Dark Mode. And instead of fixing it, they got 20 pages of propaganda on how to work around it. So I'll look at defining a set color for the bar.
> 
> You might know that linux does not obey the stylesheet for the tx_status_label, but Mac and Windows do. I'm attempting to come up with a workaround for that one.

@Chris-AC9KH Yeah, I was looking through the code and wondered why the tx_status_label colors don't work on Linux like it did before.  

I'm fine with the way the Windows color is on the progress bar.  I don't think it's really a problem.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG it should work now. I'm building it on Fedora 43 to test it. Why linux doesn't obey the style but Mac and Windows do? It's because it's linux. Should put in the code someplace `Linux == "Problems"`

Chris-AC9KH:
Or better yet:
```
if defined (Q_OS_LINUX)
    "scrap it and get a Mac"
endif
```

Chris-AC9KH:
@Joe-K0OG so I got it to work on Linux.

On Windows the color of the progress bar and the text, is chosen by the native Qt system base palette. Why Windows chose those non-contrasting colors in Dark Mode is beyond my understanding. It is poor attention to detail in Windows graphics rendering. I could probably fix it and hard-code a blue progress bar for Windows. But how many people actually use Windows in Dark Mode? Like none? So would it be worth it? Probably not.

wmiler:
> Or better yet:
> 
> ```
> if defined (Q_OS_LINUX)
>     "scrap it and get a Mac"
> endif
> ```

Fanboys...sheesh

Chris-AC9KH:
@wmiler you try to code it up for linux then.

Joe-K0OG:
> Or better yet:
> 
> ```
> if defined (Q_OS_LINUX)
>     "scrap it and get a Mac"
> endif
> ```

```
AND_IF (You_Got_Lottsa_Dough)
        "Get a Mac"
otherwise (Live_Cheeeeep)
```

Hey, I'm super-cheap, so can't afford modern Macs any more.  Besides, I enjoy the challenge of running with minimal equipment as much as possible.  ;)

The latest Linux fix builds & runs fine on both Windows & Linux, and plays well with dark OS themes, and .qss stylesheets.  Good fix @Chris-AC9KH - thanks!  From Linux Mint 22.3:
image

73,
-Joe-
K0OG

Joe-K0OG:
Maybe this should be another PR, but when I try -stylesheet <filename.qss> it flags parsing errors in the terminal window.  Not a big deal for now, but I thought I would test it just to see if that works yet.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG I got linux to work with preprocessor directives in mainwindow.cpp to compile separate code for linux vs Mac and Windows. But that's not an ideal fix for the linux problems. That tx_status_label is more complicated because it switches states so I'm going to have to come up with an enum method in styles.h to fix this.

But I'll do that tomorrow.

Joe-K0OG:
> I could probably fix it and hard-code a blue progress bar for Windows. But how many people actually use Windows in Dark Mode? Like none? So would it be worth it? Probably not.

@Chris-AC9KH The minor problem is not dark mode, but that in either light or dark theme, the progress bar slider darkness is dark, somewhat obscuring the black text, but it's not an issue worth messing with.  Since it's the default for Windows, just leave it.  No reason to muddy up the code or waste your time on such a minor thing.  I'm happy with it the way it is.

73,
-Joe-
K0OG

Joe-K0OG (approved):
Works great, and is consistent on both Linux and Windows, in all conditions:  Light system theme, Dark system theme, and with .qss stylesheets loaded.

Thanks @Chris-AC9KH!

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG @wmiler I designed a new unified progress bar similar to what Allen did with that slider thing. Except it's defined totally in styles.h. I eliminated use of palette(base) and went to defined colors for it. And I eliminated the Fusion block in the code since it is now a unified design.

Plus I added a new enum TxStatusAppearance class that defines the states of the tx_status_label and applies the proper one with a switch. This totally eliminated inline color definitions in mainwindow.cpp for the Status Bar and you can redesign the Status Bar from one place now (styles.h). I'm a lot happier with this.

I made the new progress bar slightly lighter color than the labels. Since the thing is sweeping back and forth all the time, I wanted it to be somewhat attractive, but not something that jumps right out at you like it should be the only thing to look at. So I made it slightly less obtrusive. I gave the bar slightly rounded ends and just a little hinting on the color.

In Dark Mode Windows 11 insists on putting white text on a light blue background. I have not found a way around that despite trying everything in Microsoft's 20 pages of propaganda on workarounds for Dark Mode problems. Otherwise I think it looks pretty good.

If this is all good, then I'll doxy the new styles.h and other people can start using it to define platform styles in the UI in one place and have a template to go by to see how it did it. Then this one is ready to go after I squash it and do a final review of code cleanup.

Joe-K0OG (commented):
@Chris-AC9KH I kind of preferred it with the Fusion style applied to the progress bar, but really no big deal either way.  The progress bar comes out "skinny" as compared to before, and in relation to the buttons in the status bar, and this is true on both Linux and Windows:
Image

If keeping Fusion applied to the progress bar is a problem, then this is fine.  If it can be kept for the progress bar and cause no other problems, I would prefer that.

73,
-Joe-
K0OG

P.S.  Or, without Fusion applied to the progress bar, can you set the width to be the same as the other elements in the status bar?

Joe-K0OG:
> In Dark Mode Windows 11 insists on putting white text on a light blue background. I have not found a way around that despite trying everything in Microsoft's 20 pages of propaganda on workarounds for Dark Mode problems. Otherwise I think it looks pretty good.

Well, it looks fine on my Win11 PC in Dark mode (fresh settings, no .qss stylesheet applied):
image

Maybe some Win11 updates I don't have mess with that?  I refuse do do any further Win11 updates because according to many horror stories it may permanently crash and/or lock me out of the OS.  I'm at 24H2 until it dies, then it's Linux all the way. ;o

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG I got rid of Fusion and set the unified progress bar design to 10 pixels height. I did that to make it less obtrusive and use a unified design across all platforms.

Still can't figure out the Windows dark mode issue with the text on the progress bar. According to Microsoft's propaganda on Dark Mode problems, it is a problem in the design of your app. Windows overrides the style and we're gonna put white text there even if you define it as black. Because Microsoft knows best. If you can't read it because there's white text on a white background either make your app display a black background, or highlight the text with your mouse pointer, which usually solves the problem so you can read it.

This is what it actually says in their developer documentation on this issue. They don't take into account that Windows is not the only operating system that exists.

Chris-AC9KH:
> P.S. Or, without Fusion applied to the progress bar, can you set the width to be the same as the other elements in the status bar?

I can, but it's not being done at this time. You can define this boolean as true if you want the bigger progress bar. The option is there, it's wired to use the smaller one for the unified design. If you want to use the bigger one you'll have to define it by platform. But the unified design follows Qt's QProgressBar developer recommendations for a cross-platform, clean-looking and modern design. So that's what we're going with.
```
// QProgressBar StyleSheet
inline QString progress_bar_stylesheet(const QColor& textColor, bool small = false);

inline QString progress_bar_stylesheet(const QColor& textColor, bool small)
{
    const QString base = QString(
        "QProgressBar {"
        "  border: 0px;"
```
I'm running Windows 11 25H2, Build 26200.8037 and it is fully up-to-date with Windows Update. It is arm64 Windows and it is very fast and responsive running in a VM on my Mac mini M4Pro.

Screenshot 2026-04-11 at 13 17 36

I've switched to using clang to compile on Windows with Qt Creator. Microsoft has clang tuned right up on Win 11 arm64 and it is WAAY faster than MinGW gcc. Like 4x faster. And it can compile either native arm64, or x86_64 with Microsoft's Prism translator. Quite impressive, actually.

Joe-K0OG:
> I can, but it's not being done at this time. You can define this boolean as true if you want the bigger progress bar. The option is there, it's wired to use the smaller one for the unified design. If you want to use the bigger one you'll have to define it by platform. But the unified design follows Qt's QProgressBar developer recommendations for a cross-platform, clean-looking and modern design. So that's what we're going with.

@Chris-AC9KH Nahhh - it's fine the way it is, so leave it.  I only mentioned it because I noticed it was different, and wasn't sure if that was an expected or intended result.  It's all good.

>I'm running Windows 11 25H2, Build 26200.8037 and it is fully up-to-date with Windows Update. 

Maybe that's the reason my Win11 doesn't display white text in Dark mode since I'm at 24H2.  I know that a **lot** of Win11 users are not updating to 25H2, so there is going to be a mix of users in the field.  Too bad it's not the same on all versions of Win11 - what a mess MS & Win is!  We're stuck with it for now...

I just remembered I do (or did?) have a 25H2 test system that I think still works.  I may try the latest with it and see what it does.

73,
-Joe-
K0OG

Chris-AC9KH:
@wmiler @Joe-K0OG so this is ready to go. Rebased down to one commit
- created new styles.h header which is your new default StyleSheet configuration after the rest of the code is cleaned up to use it
- removed unneeded code from mainwindow.cpp
- provides a template as how to apply platform-specific styles to the UI
- provided a definition of USE_QSS_THEMES in CMakeLists.txt as an example of how QSS can be controlled or applied on a per-platform basis
- styles.h is doxy hooked

Joe-K0OG:
@Chris-AC9KH Built & tested on Win11 and Linux Mint - works great!  ...with one minor exception that you noted...

I booted a Win11 25H2 test system, and sure enough there the progress-bar numbers are white when the OS is in dark mode.  Too bad!  However, it's not a killer situation in my view because when the blue covers the numbers, they are somewhat readable.  Interesting that this does *not* happen on Win11 24H2.  Yeah, 25H2 is an improvement - right!  Good job `$MS$` !!!

73,
-Joe-
K0OG

Chris-AC9KH:
Yeah, Microsoft's developer documentation doesn't point out a way to fix it either unless you are developing a Windows-specific application. Qt's documentation just says Windows overrides styles in certain cases and you can maybe do something like change your colors to be contrasting on Windows. Since I don't think many people use Dark Mode on Windows anyway, I'm not loosing any sleep over it.

Joe-K0OG:
@Chris-AC9KH Agreed - it's not a big deal since it's still quite useable in Dark mode.  We might hear a complaint or two, but not enough to be worried about I don't think.  Who knows what the next Win update will do anyway...

73,
-Joe-
K0OG

Joe-K0OG:
@Chris-AC9KH I'm just asking a question here, since I don't fully understand all this...

Why don't we just set the progress-bar text color to black and leave it there?  At least on Linux and Windows, I think this would work fine.  I just built the following and tested it on Win11 24H2, 25H2, and Linux Mint 22.3, Light and Dark themes, and it works perfectly in all cases, locking the progress-bar numbers to black.

   ```
     const QString base = QString(
        "QProgressBar {"
        "  border: 0px;"
        "  background-color: #ffffff;"
        "  color: %1;"  // Need to keep this here to maintain progress bar geometry
        "  color: #000000;"  // Added to force font color to black
        "  text-align: center;"
        "  padding: 0px;"
        "  %2"
        "}"
```

Is there some reason why we can't do this?  Does it break something on MacOS?

At first I tried just changing the "%1" to "#000000" but the progress-bar geometry was affected by this.  That's why I decided to leave it and just add another line with the color set to black, and that worked perfectly.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG that is a malformed QString. You can't do that. In the function declaration it takes a Qt color object by const reference, which forms the basic look of the progress bar, and applies it to the %1 variable. What you did will cause the progress bar to not show on MacOS at all. Why? MacOS uses its native style engine and if you use a malformed QString like that where it declares a variable, then immediately overwrites it with hard-code, the widget is broken.

Ham-fisted linux developers have been in the code before doing this kind of stuff with "hack fixes", and the progress bar was broken on Mac for ages. So I eliminated it back in the pre-2.3.1 days for Mac, and I just fixed it again with the new unified bar design.

The main issue is that Qt on Windows in Dark Mode overrides stylesheet colors with system theme colors on PermanentWidgets. This needs to be fixed in Windows itself. But I like I said earlier, Microsoft does not think any other operating systems exist. So the "fix" they provide in their 20 or so pages of propaganda on Dark Mode problems in Windows they say the problem is your app - fix your app. But the more explicit and forceful you try to be with the StyleSheet code, then it breaks things on other platforms.

The code is solid, and I'm not changing it. Maybe the next Windows Update will fix it. But don't hold your breath.

Joe-K0OG:
> What you did will cause the progress bar to not show on MacOS at all.

Ahhh - good - that's why I was asking.  We're good with what we have then, no problem.

Thanks & 73,
-Joe-
K0OG

PR #253, opened 2026-04-12
Final full implementation of styles.h

Chris-AC9KH:
- Add platform-specific styles for QPushButton on Windows and Mac only, Linux uses default style
- Fully document styles.h with doxy
- Fix Windows 11 dark mode issue in the progress bar
- Remove QSS injection restriction from the styles.h This will be up to linux developers to define

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -230,14 +230,6 @@ configure_file(
   "${CMAKE_CURRENT_BINARY_DIR}/js8call_config.h"
 )
 
-#------------------------------------------------------------------------------#
-# On Linux we allow QSS theme injection
-#------------------------------------------------------------------------------#
-if (UNIX AND NOT APPLE)
-      target_compile_definitions(
-      ${TARGET} PRIVATE USE_QSS_THEMES)
-endif()
-
 target_precompile_headers(
   ${TARGET} PRIVATE
   ${CMAKE_CURRENT_BINARY_DIR}/js8call_config.h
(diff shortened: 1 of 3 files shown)


Chris-AC9KH:
@wmiler one thing that should be noted here is that Qt's stylesheet system works by last-write-wins at the QApplication level. So IF you want to be able to completely style the app with a .qss your qApp->setStyleSheet() needs to be declared immediately following my declaration in main.cpp where the Qt object that owns the event loop is created, and before a.exec() is called. That way your style sheet that's loaded at runtime replaces the platform style, it doesn't merge or append.

One caveat: anything defined in styles.h has to be defined in your stylesheet or it won't be there.

Joe-K0OG:
> One caveat: anything defined in styles.h has to be defined in your stylesheet or it won't be there.

Yes, good to know @Chris-AC9KH I was confused by some experiments I did with some partial stylesheets, and this explains it.

I built and tested on Linux Mint 22.3, Win 11 24H2 and 25H2, light and dark modes on all, and loading some stylesheets, and all works as it should - looking good Chris!  Thanks for ALL the hard work you did on this!

73,
-Joe-
K0OG

Joe-K0OG (approved):
Looks good - no problems found in my testing on Linux, Win11 24H2 and 25H2.

Thanks!
73,
-Joe-
K0OG

Chris-AC9KH:
I'll merge this one. It's already been thoroughly thrashed.

PR #256, opened 2026-04-15
New, more modern header bar design for JS8Call 3

Chris-AC9KH:
See the discussions at https://github.com/orgs/JS8Call-improved/discussions/251 for more information on this PR.

### DIFF
--- a/JS8_UI/mainwindow.ui
+++ b/JS8_UI/mainwindow.ui
@@ -55,7 +55,7 @@
       </property>
       <widget class="QWidget" name="logHorizontalWidget" native="true">
        <property name="sizePolicy">
-        <sizepolicy hsizetype="Preferred" vsizetype="MinimumExpanding">
+        <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
          <horstretch>0</horstretch>
          <verstretch>1</verstretch>
         </sizepolicy>
@@ -63,7 +63,7 @@
        <property name="minimumSize">
         <size>
          <width>0</width>
-         <height>125</height>
+         <height>0</height>
         </size>
        </property>
        <property name="maximumSize">
@@ -90,6 +90,9 @@
         </property>
         <item>
          <widget class="QFrame" name="logWidget">
+          <property name="styleSheet">
+           <string>QFrame#logWidget { background-color: #F2F2F0; }</string>
+          </property>
           <property name="sizePolicy">
            <sizepolicy hsizetype="Expanding" vsizetype="Minimum">
             <horstretch>0</horstretch>
@@ -99,7 +102,7 @@
           <property name="minimumSize">
            <size>
             <width>0</width>
-            <height>125</height>
+            <height>10</height>
            </size>
           </property>
           <property name="maximumSize">
@@ -107,9 +110,6 @@
             <width>16777215</width>
             <height>16777215</height>
            </size>
-          </property>
-          <property name="styleSheet">
-           <string notr="true">QFrame { background:black; }</string>
           </property>
           <property name="frameShape">
            <enum>QFrame::Shape::NoFrame</enum>
@@ -144,12 +144,6 @@
              <property name="autoFillBackground">
               <bool>false</bool>
              </property>
-             <property name="styleSheet">
-              <string notr="true">QFrame {
-	color:white;
-	background-color:rgb(0,0,0);
-}</string>
-             </property>
              <property name="frameShape">
               <enum>QFrame::Shape::NoFrame</enum>
              </property>
@@ -161,69 +155,35 @@
                <number>6</number>
               </property>
               <property name="leftMargin">
-               <number>10</number>
+               <number>20</number>
               </property>
               <property name="topMargin">
-               <number>6</number>
+               <number>0</number>
               </property>
               <property name="rightMargin">
                <number>0</number>
               </property>
               <property name="bottomMargin">
-               <number>6</number>
+               <number>0</number>
               </property>
-              <item>
-               <spacer name="verticalSpacer_6">
-                <property name="orientation">
-                 <enum>Qt::Orientation::Vertical</enum>
-                </property>
-                <property name="sizeHint" stdset="0">
-                 <size>
-                  <width>20</width>
-                  <height>10</height>
-                 </size>
-                </property>
-               </spacer>
-              </item>
               <item>
                <layout class="QHBoxLayout" name="horizontalLayout_18">
                 <property name="topMargin">
                  <number>0</number>
                 </property>
                 <item>
-                 <widget class="QLCDNumber" name="currentFreq">
+                 <widget class="QLabel" name="currentFreq">
                   <property name="minimumSize">
                    <size>
                     <width>200</width>
-                    <height>40</height>
+                    <height>0</height>
                    </size>
                   </property>
-                  <property name="toolTip">
-                   <string>Set Current Frequency</string>
+                  <property name="alignment">
+                   <set>Qt::AlignmentFlag::AlignCenter</set>
                   </property>
-                  <property name="frameShape">
-                   <enum>QFrame::Shape::NoFrame</enum>
-                  </property>
-                  <property name="frameShadow">
-                   <enum>QFrame::Shadow::Plain</enum>
-                  </property>
-                  <property name="lineWidth">
-                   <number>0</number>
-                  </property>
-                  <property name="smallDecimalPoint">
-                   <bool>false</bool>
-                  </property>
-                  <property name="digitCount">
-                   <number>10</number>
-                  </property>
-                  <property name="segmentStyle">
-                   <enum>QLCDNumber::SegmentStyle::Flat</enum>
-                  </property>
-                  <property name="value" stdset="0">
-                   <double>1234567890.000000000000000</double>
-                  </property>
-                  <property name="intValue" stdset="0">
-                   <number>1234567890</number>
+                  <property name="text">
+                   <string>1234567890</string>
                   </property>
                  </widget>
                 </item>
@@ -237,18 +197,6 @@
                   </property>
                   <item>
                    <widget class="QPushButton" name="dialFreqUpButton">
-                    <property name="minimumSize">
-                     <size>
-                      <width>30</width>
-                      <height>24</height>
-                     </size>
-                    </property>
-                    <property name="maximumSize">
-                     <size>
-                      <width>30</width>
-                      <height>24</height>
-                     </size>
-                    </property>
                     <property name="cursor">
                      <cursorShape>PointingHandCursor</cursorShape>
                     </property>
@@ -258,36 +206,27 @@
                     <property name="autoFillBackground">
                      <bool>false</bool>
                     </property>
+                    <property name="text">
+                     <string>▲</string>
+                    </property>
                     <property name="styleSheet">
-                     <string notr="true">QPushButton {
-	background-color:#000;
-	background-image:url(&quot;:/images/artwork/up.png&quot;);
-	background-position: center center;
-	background-repeat:no-repeat;
-	border:1px solid;
-	border-radius:2px;
-}
-
-QPushButton:pressed {
-	background-color:#222;
-}</string>
+                     <string notr="true">
+                    QPushButton {
+                        background-color: #000000;
+                        color: #39FF14;
+                        font-size: 12pt;
+                        border:1px solid;
+                        border-radius:2px;
+                    }
+                    QPushButton:pressed {
+                        background-color: #222;
+                    }
+                    </string>
                     </property>
                    </widget>
                   </item>
                   <item>
                    <widget class="QPushButton" name="dialFreqDownButton">
-                    <property name="minimumSize">
-                     <size>
-                      <width>30</width>
-                      <height>24</height>
-                     </size>
-                    </property>
-                    <property name="maximumSize">
-                     <size>
-                      <width>30</width>
-                      <height>24</height>
-                     </size>
-                    </property>
                     <property name="cursor">
                      <cursorShape>PointingHandCursor</cursorShape>
                     </property>
@@ -297,19 +236,22 @@ QPushButton:pressed {
                     <property name="autoFillBackground">
                      <bool>false</bool>
                     </property>
+                    <property name="text">
+                    <string>▼</string>
+                    </property>
                     <property name="styleSheet">
-                     <string notr="true">QPushButton {
-	background-color:#000;
-	background-image:url(&quot;:/images/artwork/down.png&quot;);
-	background-position: center center;
-	background-repeat:no-repeat;
-	border:1px solid;
-	border-radius:2px;
-}
-
-QPushButton:pressed {
-	background-color:#222;
-}</string>
+                     <string notr="true">
+                      QPushButton {
+                       background-color: #000000;
+                       color: #39FF14;
+                       font-size: 12pt;
+                       border:1px solid;
+                       border-radius:2px;
+                      }
+                      QPushButton:pressed {
+                       background-color: #222;
+                      }
+                    </string>
                     </property>
                    </widget>
                   </item>
@@ -332,11 +274,13 @@ QPushButton:pressed {
                  <string>Set Current Frequency Offset</string>
                 </property>
                 <property name="styleSheet">
-                 <string notr="true">QLabel {
-font-size: 14pt;
-line-height:12pt;
-color : white;
-}</string>
+                 <string notr="true">
+                  QLabel {
+                   font-size: 14pt;
+                   line-height:12pt;
+                   color : black;
+                  }
+                 </string>
                 </property>
                 <property name="text">
                  <string>1500 Hz</string>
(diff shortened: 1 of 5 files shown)


Chris-AC9KH:
Example screenshot of the new design for JS8Call 3.0. This is now configurable in styles.h on a per-platform basis using the preprocessor to compile the default style for that platform into the binary. Mac and Windows styles are designed based on their individual developer guidelines for "look and feel". Linux (at present) retains the GTK/Windows 95 style for buttons, widgets and fonts and uses the system default font for the frequency display. This can be tweaked for the default on linux but there is no consensus at present if Linux should use old GTK style, Wayland style, etc. This will be up to Linux developers to decide upon.

Screenshot 2026-04-15 at 07 41 53

JS8_UI/mainwindow.ui
```diff
@@ -486,7 +407,7 @@ color : white;
                  <number>0</number>
                 </property>
                 <property name="text">
-                 <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p align=&quot;center&quot;&gt;&lt;span style=&quot; font-size:12pt;&quot;&gt;01:23:45&lt;br/&gt;2015 Jun 17 &lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+                 <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p align=&quot;center&quot;&gt;&lt;span style=&quot; font-size:18pt;&quot;&gt;01:23:45 2015 Jun 17&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
wmiler, line 350: Is there a reason this is in html?

wmiler (approved):
This looks fine, after this one merges, I will need to do some Linux related mods (enable via styles.h, etc), waterfall runs into macro buttons being the main thing, and adjusting the colors abit.

wmiler:
One style question: does it make sense, for the frequency, adjustment arrows, and date/time backgrounds to match the rest of the background that it sits on?

JS8_UI/mainwindow.ui
Chris-AC9KH: Yes, it allows rich formatting of the string vs using plain text in a QLabel.

Chris-AC9KH:
> One style question: does it make sense, for the frequency, adjustment arrows, and date/time backgrounds to match the rest of the background that it sits on?

No, it can have just a border with no background. Or no border if desired. I chose to fill it with the frequency display label background to signify that it applies only to the frequency adjustment for the up/down arrows.

I originally designed the date/time to be on the standard background. Making a LCD-style display "dressed up" the header bar a bit and made it look nicer.

Joe-K0OG:
> This looks fine, after this one merges, I will need to do some Linux related mods (enable via styles.h, etc), waterfall runs into macro buttons being the main thing, and adjusting the colors abit.

@wmiler I'm not wure what you mean by the waterfall running into the macro buttons.  Here is my screen (no .qss stylesheet applied, so default appearance) on Linux Mint 22.3:
image

It looks really good to me.

73,
-Joe-
K0OG

Chris-AC9KH:
If you want me to, for runtime qss-injection, I can try to move the the hardcoded stuff out of mainwindow.ui to C++ code. Then your questions above can be solved by defining a default style for every platform in styles.h. If the linux guys want faded orange background with royal purple on blaze pink LCD displays they can define that in styles.h for their default and the MacOS and Windows people don't have to deal with the linux folks' weird-looking configurations.

I can do that for this PR, but it will take me awhile to define all that in C++ and will probably make your runtime qss-injection project easier.

Joe-K0OG:
> faded orange background with royal purple on blaze pink LCD displays they can define that

Ha, ha, ha!!  Now you're talkin'!  Thanks Chris for the little chuckle.

73,
-Joe-
K0OG

rruchte:
> If the linux guys want faded orange background with royal purple on blaze pink LCD displays they can define that 

I'm all for making JS8Call the Winamp of ham radio applications.

wmiler:
This is the way it initially shows on Ubuntu 24.04:

image

Chris-AC9KH:
Give me a couple days in my spare time and I'll move the hard-coded XML definitions to C++. I think that would be better anyway because then I can apply the default platform style to the six buttons, which are presently hard-coded. IF I get in the right frame of mind I could probably do it in a couple hours. But we run a firewood processing business here and wife is all excited about getting the processor mounted on her track loader to see if the hydraulic hoses she blew on it last fall hold oil now.

Joe-K0OG:
> This is the way it initially shows on Ubuntu 24.04:

Confirmed:  I see what you are talking about.  If you grab the puller at the top of the waterfall and pull down a few pixels, it will pull away from the buttons.  I found on my Linux Mint 22.3 if I push up the puller above the waterfall, it will slightly cover the button bar like your screenshot:
image

So, I guess the movement limit needs to be adjusted to prevent the overlap.

73,
-Joe-
K0OG

Joe-K0OG:
> I can do that for this PR, but it will take me awhile to define all that in C++ and will probably make your runtime qss-injection project easier.

It seems reasonable to me to hold off the PR and go ahead and do this.  I don't think there is any rush to push this PR.  If it takes a few weeks before we are ready for this redesign PR, that's fine since this isn't any kind of basic functionality issue.

73,
-Joe-
K0OG

Chris-AC9KH:
That thing is called a QSplitter handle and it's just the difference in the way Qt renders on linux vs Windows vs Mac. It is made to be adjustable, so adjust it to fit what you want. The default window layout itself is broken and unusable when the program is first opened with no JS8Call.ini. It is expected that you drag stuff around to make it look like you want.

wmiler:
So I'll be up on a mountain, starting Friday thru Sunday prepping a foundation for a ham tower at EM86sa, so I will be out of pocket this weekend.

Chris-AC9KH:
I'll point out that right here is the problem for the linux people. There's no consensus on what style they want. So it's not defined for Linux so you have no built-in padding on your QPushButtons unless somebody decides on what style they're gonna use. Apparently, some of them go into a coma if the buttons aren't square, some of them don't like it unless the buttons are blah-grey, so nothing is defined for a linux default. I told you guys I was not touching that one with a 10ft pole until the linux folks all get on one page.

```
// Suggested possibility for linux - not defined at this time
//#elif defined(Q_OS_LINUX)
//    return R"(
//        QPushButton {
//            background-color: #E0E0E0;
//            color: #212121;
//            border: 1px solid #BDBDBD;
//            border-radius: 1px;
//            padding: 5px 15px;
//            font-family: "Ubuntu", "Noto Sans";
//        }
//        QPushButton:hover {
//            background-color: #F5F5F5;
//        }
//        QPushButton:pressed {
//            background-color: #BDBDBD;
//        }
//        QPushButton:disabled {
//            background-color: #ececec;
//            color: #888888;
//        }
//    )";
```

Joe-K0OG:
> until the linux folks all get on one page.

My opinion here, as a primarily Linux user:  The default appearance, as-delivered, should be as similar as possible across all platforms, then should be .qss stylesheet themeable at runtime where it can (I guess primarily on Windows and Linux).

73,
-Joe-
K0OG

rruchte:
On Linux, the appearance is (or should be) largely determined by the desktop environment. There is no same page for Linux users to get onto. There are tools like Kvantum that attempt to theme all QT apps on a system, with varying degrees of success. QT also provides the QT_QPA_PLATFORMTHEME environment variable that provides some options for theming. I sometimes have to add QT_QPA_PLATFORMTHEME=gnome to my .desktop files to make QT apps respect the system theme. The goal should be to make the app respect the system theme as much as possible, not create a theme that's supposed to satisfy all Linux users, which is an impossible task.

Chris-AC9KH:
Well, I already pulled everything related to styles out of the .ui file. Hammering out the C++ for it was easy. Breaking that mainwindow.ui file, OTOH, is quite easy to do. That's why I keep track of what I've done with Xcode - I can revert it easily from the diff if it don't run or work!

Screenshot 2026-04-15 at 18 47 05

I'll have an initial push to the PR here shortly.

Chris-AC9KH:
So this is what I got on MacOS with my new code. Sure looks good to me.

Screenshot 2026-04-15 at 18 56 06

So, the way I understand the linux thing; Mac and Windows get their native styles. Linux gets old square buttons and frames in that header bar. If that's not what you want, shoot now or forever hold your piece.

Chris-AC9KH:
@Joe-K0OG try this code out now. If there's anything you want changed for the defaults on Linux it is very easy to make a change to the "other" preprocessor section in the new Styles namespace at the bottom of styles.h. So you can make changes there, compile the code to see what it looks like, and if you want it to be the default for Linux post the code block here and I'll change it in this PR before it goes in.

Don't get all hot and bothered with your qss stylesheets. That's not going to work any better than it did before. I just made it easier for Wyatt to modify it with runtime qss injection.

Chris-AC9KH:
You might know it doesn't build on linux. I'll fix that.

rruchte:
On my system (Manjaro, Wayland, Gnome), there are a couple of things that seem off out of the box, aesthetics aside. 

The font color on the buttons is white (this is also the case in master, so not the fault of this change), which is tolerable on the black background, but doesn't really work on the lighter background. 

The freq display is defaulting to some variable width font instead of the fixed width font that the time display is using.

Screenshot From 2026-04-15 21-52-38

Chris-AC9KH:
I had another build error on linux that I pushed a fix for again. @rruchte the date/time has to be monospace or it "jumps" every time the second changes. There's no font set for Linux for the frequency display because there is no standards like there is for Mac and Windows. Mac uses Monaco. Windows uses Consolas. Linux uses whatever because nobody can come up what to define it with.

Chris-AC9KH:
@rruchte right here in the code is where you can define what you want your buttons to look like on Linux. Otherwise it's a crap shoot. I declared when I started this project that I'm not touching linux with a 10ft pole.

```
    // Suggested possibility for linux - not defined at this time
    // #elif defined(Q_OS_LINUX)
    //    return R"(
    //        QPushButton {
    //            background-color: #E0E0E0;
    //            color: #212121;
    //            border: 1px solid #BDBDBD;
    //            border-radius: 1px;
    //            padding: 5px 15px;
    //            font-family: "Ubuntu", "Noto Sans";
    //        }
    //        QPushButton:hover {
    //            background-color: #F5F5F5;
    //        }
    //        QPushButton:pressed {
    //            background-color: #BDBDBD;
    //        }
    //        QPushButton:disabled {
    //            background-color: #ececec;
    //            color: #888888;
    //        }
    //    )";
```

rruchte:
OK, I'll take a look at that. Is there a way to just use the same monospace font in the freq display that the time display uses?

Also, for fahrvergnügen, what do you think about using heredoc to define the QSS? Much easier to read and work with, you don't have to escape quotes, etc.

```cpp
constexpr const char *LabUTCStyle = R"QSS(
QLabel {
    border-radius:0px;
    font-size: 18pt;
    line-height:18pt;
    font-family: "DejaVu Sans Mono", "Liberation Mono", "Noto Mono", "Ubuntu Mono", monospace;
    font-weight: bold;
    background-color: black;
    color : #39FF14;
}
)QSS";
```

vs

```cpp
constexpr const char *LabUTCStyle =
    "QLabel {"
    "    border-radius:0px;"
    "    font-size: 18pt;"
    "    line-height:18pt;"
    "    font-family: "DejaVu Sans Mono", "Liberation Mono", "Noto Mono", "Ubuntu Mono", monospace;
    "    font-weight: bold;"
    "    background-color: black;"
    "    color : #39FF14;"
    "}";
```

Chris-AC9KH:
@rruchte the code builds on all three platforms. I'm happy with the default styles for Mac and Windows 11 after one final tweak to adjust the background color of the bar because it matches exactly the color of the windows title bar. After I get that tweaked for Windows I'll get this merged in. Then the linux guys can duke it out as to what they want for a default "look and feel". It's super-easy to modify it in styles.h and whoever wins the beauty contest can submit a PR later for the linux default style.

rruchte:
Can we just add color: black; to the ButtonGridStyle? Since we're already customizing the background color there, it makes sense to explicitly set the text color too.

rruchte:
@Chris-AC9KH Setting the color to black on ButtonGridStyle and setting the font-family on QLabel#currentFreq to the same as LabUTCStyle solves my problems. I think both of these are The Right Way To Do It, not polluting the styles with anything Linux specific.

Screenshot From 2026-04-15 22-39-53

Chris-AC9KH:
@rruchte after I merge this you can do that on linux if you want. Don't do anything with Mac/Windows because they both use mouse-over custom colors. The design convention on MacOS is that button text changes from black to white on select. That is also defined for Windows.

Chris-AC9KH:
After this builds on all the platforms I'm going to merge it. Then the linux guys can decide on what they want in styles.h and PR it. And Wyatt can figure out his runtime qss injection. The header bar pointers are declared in the UI_Constructor member function right after  ui->setupUi(this). So runtime qss-injection has to happen after that to override those defined styles that are now different for each platform.

Chris-AC9KH:
@rruchte this built on all the CI builds. So I'm going to merge it.

You can also define the header bar components to use the wayland-style 2px rounding on the corners if you want. The linux style for the header bar is defined in styles.h right near the bottom after the `#else` preprocessor directive. The Frequency display is further up in the file, starting on line 279 for linux. You'll have to define whatever font-family you want to use there because I didn't define one.

Chris-AC9KH:
On MacOS the color of the button turns slightly darker blue with the switch from black to white text on select or mouseover like it did here when I moused over the button before I took the screenshot. And this applies to the entire UI, not just the header bar or macro buttons. This is applied globally to the mainwindow AND Configuration settings, etc..

Screenshot 2026-04-15 at 22 22 06

Chris-AC9KH:
Sidenote: the buttonStyle() function is not defined on linux at all. I wrote it out, then there was a revolt because the buttons weren't square. So I reduced the radius to 1px and commented the whole thing out so linux just gets the default.

Chris-AC9KH:
> @Chris-AC9KH Setting the color to black on ButtonGridStyle and setting the font-family on QLabel#currentFreq to the same as LabUTCStyle solves my problems. I think both of these are The Right Way To Do It, not polluting the styles with anything Linux specific.

@rruchte you won't pollute the styles with anything linux-specific. I think the reason your header bar buttons have white text is because the QPushButton style is not declared for linux. So something in the theming on your system makes the text white while it's black on every linux system I've tested.

The way I wrote the namespace Styles{} is that I went thru the mainwindow.ui file and took the properties for every stylesheet declared in there for logWidget, and wrote them out in C++ for MacOS. After I had every one of those stylesheets gone from mainwindow.ui I copied and pasted what I wrote for MacOS for the other two preprocessor sections. Then customized it per platform.

The Frequency display was already done earlier so I didn't include that the namespace Styles{}.

So basically, if you want Blaze Orange Frequency digits on Florescent Green background on linux, you can do that and it won't affect the other two. If the linux guys can come up with a consensus on what they want for defaults, just define it in styles.h and PR it. I assume it will have to be tested on several linux distributions before it is set in stone. But that's why I did this - in theory, you can switch linux back to the old Big Black Bar if you want and Mac and Windows will have the new one. In the end, the only thing that was done here was to remove the hardcoded stuff in mainwindow.ui and move it C++. The layout in mainwindow as to where stuff is at isn't changed, other than reducing the minimum size so Mac and Windows users don't have to have that Big Black Bar anymore.

And I did change the code to get rid of the QLCDNumber that was used for the frequency display and went to a QLabel instead. Why? There's very little you can do with QLCDNumber for styling.

Joe-K0OG:
Wow - a FLURRY of activity I see when I look at my email this morning!  I'll do some testing on Linux & Windows this AM.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG there's not so much "testing" that needs to be done as there is somebody coming up with default configuration for linux. It is not fully defined in styles.h. Somebody needs to decide what the defaults are for linux and submit a PR to set them in styles.h

So when you are "testing" on linux keep in mind that the QPushButton style is not even defined for linux so Rob has white text on his buttons, you maybe got black text on yours. Nothing is going to be the same.

Joe-K0OG:
>So when you are "testing" on linux keep in mind that the QPushButton style is not even defined for linux so Rob has white text on his buttons, you maybe got black text on yours. Nothing is going to be the same.

I answered your question (bold below about button text color), but I posted in the wrong place a while ago @Chris-AC9KH -- here is my test result:
=-=-=-
@Chris-AC9KH Excellent work Chris - thanks for doing it so quickly, but don't burn yourself out!

On Windows 11 24H2 and 25H2, it runs fine in both light and dark system themes.  I don't see any problems there.  It built without any problems in JTSDK64-Tools.

On Linux Mint 22.3, the following build warning was flagged:
```
[ 99%] Building CXX object CMakeFiles/JS8Call.dir/JS8_Widgets/LazyFillComboBox.cpp.o
/js8call/js8call-prefix/src/JS8_UI/mainwindow.cpp: In member function ‘void UI_Constructor::createStatusBar()’:
/js8call/js8call-prefix/src/JS8_UI/mainwindow.cpp:1451:18: warning: variable ‘textColor’ set but not used [-Wunused-but-set-variable]
      |                  ^~~~~~~~~
[100%] Linking CXX executable JS8Call
[100%] Built target JS8Call
```

This did not cause any problems, and here is the result on a light system theme:
image

This is excellent, no problem with it.

However, with a dark system theme on Linux, the button text colors are white as was mentioned, so we will need to recommend something for that.  **I suggest that we just set those to black all the time** since on Linux the button colors don't change much with system theme, and the background remains the same.
image

Big progress here - great work Chris!

73,
-Joe-
K0OG

PR #257, opened 2026-04-16
Implement Linux styles

wmiler:
This turns on the default style in styles.h for linux. Uses either Ubuntu fonts, or Noto Sans as available. Normalized the colors with Windows/Mac a bit.

Needs testing across the various Linux derivatives to see if we are close enuf.

### DIFF
--- a/JS8_UI/styles.h
+++ b/JS8_UI/styles.h
@@ -394,7 +394,7 @@ constexpr const char *MonitorButtonStyle =
 
 constexpr const char *LogQSOButtonStyle =
     "QPushButton {"
-    "    background-color:lightgray;"
+    "    background-color:#6699ff;"
     "    color:black;"
     "    padding:0.25em 0.25em; font-weight:normal;"
     "    border-style:solid;"
@@ -406,7 +406,19 @@ constexpr const char *LogQSOButtonStyle =
     "    color:black;"
     "}";
 
-constexpr const char *TuneButtonStyle = LogQSOButtonStyle; // Same as above
+constexpr const char *TuneButtonStyle =
+    "QPushButton {"
+    "    background-color:lightgray;"
+    "    color:black;"
+    "    padding:0.25em 0.25em; font-weight:normal;"
+    "    border-style:solid;"
+    "    border-width:0px;"
+    "    border-radius:6px;"
+    "}"
+    "QPushButton:checked {"
+    "    background-color:#6699ff;"
+    "    color:black;"
+    "}";
 
 constexpr const char *ModeButtonStyle =
     "QPushButton {"
--- a/JS8_UI/styles.h
+++ b/JS8_UI/styles.h
@@ -263,7 +263,7 @@ static inline QString logFrameStyle() {
                           " min-height: 40px;"
                           "}");
 #elif defined(Q_OS_WIN)
-    return QStringLiteral("QFrame#frame { background-color: #F2F2F0; }"
+    return QStringLiteral("QFrame#frame { background-color: #DDEEFF; }"
                           "QLabel#currentFreq {"
                           " color: #39FF14;"
                           " background-color: black;"
@@ -361,6 +361,7 @@ constexpr const char *ButtonGridStyle =
 constexpr const char *MonitorTxButtonStyle =
     "QPushButton {"
     "    background-color:lightgray;"
+    "    color:black;"
     "    padding:0.25em 0.25em; font-weight:normal;"
     "    border-style:solid;"
     "    border-width:0px;"
@@ -368,14 +369,17 @@ constexpr const char *MonitorTxButtonStyle =
     "}"
     "QPushButton:checked {"
     "    background-color:#22FF22;"
+    "    color:black;"
     "}"
     "QPushButton[transmitting=\"true\"] {"
     "    background-color:#FF2222;"
+    "    color:black;"
     "}";
 
 constexpr const char *MonitorButtonStyle =
     "QPushButton {"
     "    background-color:lightgray;"
+    "    color:black;"
     "    padding:0.25em 0.25em; font-weight:normal;"
     "    border-style:solid;"
     "    border-width:0px;"
@@ -383,11 +387,13 @@ constexpr const char *MonitorButtonStyle =
     "}"
     "QPushButton:checked {"
     "    background-color:#22FF22;"
+    "    color:black;"
     "}";
 
 constexpr const char *LogQSOButtonStyle =
     "QPushButton {"
     "    background-color:lightgray;"
+    "    color:black;"
     "    padding:0.25em 0.25em; font-weight:normal;"
     "    border-style:solid;"
     "    border-width:0px;"
@@ -395,6 +401,7 @@ constexpr const char *LogQSOButtonStyle =
     "}"
     "QPushButton:checked {"
     "    background-color:#6699ff;"
+    "    color:black;"
     "}";
 
 constexpr const char *TuneButtonStyle = LogQSOButtonStyle; // Same as above
@@ -406,14 +413,17 @@ constexpr const char *ModeButtonStyle =
     "    border-width:0px;"
     "    border-radius:6px;"
     "    background-color:#6699ff;"
+    "    color:black;"
     "}"
     "QPushButton:checked {"
     "    background-color:#6699ff;"
+    "    color:black;"
     "}";
 
 constexpr const char *SpotButtonStyle =
     "QPushButton {"
     "    background-color:lightgray;"
+    "    color:black;"
     "    padding:0.25em 0.25em; font-weight:normal;"
     "    border-style:solid;"
     "    border-width:0px;"
@@ -421,6 +431,7 @@ constexpr const char *SpotButtonStyle =
     "}"
     "QPushButton:checked {"
     "    background-color:#6699ff;"
+    "    color:black;"
     "}";
 
 }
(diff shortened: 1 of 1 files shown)


Joe-K0OG:
@wmiler Wyatt, thanks for doing this.  I'll test on Linux Mint 22.3, a couple of Ubuntu variants, Debian 12 and 13, MX Linux, and Fedora 39 and 41, as well as Windows -- just to make sure...  ;)  I'll try to get back to you on this in a few hours.

73,
-Joe-
K0OG

Chris-AC9KH:
@wmiler you can do more tweaking for linux in there too, like maybe setting an appropriate monospace font for the frequency display. There is nothing set for linux because I didn't know what to use. It is set to use Monaco on Mac and Consolas on Windows. But a font-family is not defined for linux so it's likely going to use a default font that's not monospace. You might also notice Mac and Windows use a smaller font size for the frequency display so it looks more like this:

Screenshot 2026-04-16 at 12 31 25

I was not sure what the linux folks wanted. So somebody who runs linux all the time really needs to go thru there and set whatever they come up with for defaults on linux. It is only partially defined right now compared to Mac and Windows.

rruchte:
@wmiler Can you add the monospace font stack to QLabel#currentFreq so the frequency display font matches the date/time font?

Current:
Screenshot From 2026-04-16 13-35-27

Font applied:
Screenshot From 2026-04-15 22-39-53

Chris-AC9KH:
And I'll also mention that if the linux people don't like the unified progress bar, you can change that too.

It's pretty simple now:
- just add a preprocessor directive `#if defined(Q_OS_LINUX)`
- write out what you want for the progress_bar_stylesheet() function
- put what's there now under `#else`
- add a `#endif` on the bottom
- voila! linux has a custom progress bar and only Mac and Windows use the unified one.

Joe-K0OG:
> And I'll also mention that if the linux people don't like the unified progress bar, you can change that too.
> 
> It's pretty simple now:
> 
>     * just add a preprocessor directive `#if defined(Q_OS_LINUX)`
> 
>     * write out what you want for the progress_bar_stylesheet() function
> 
>     * put what's there now under `#else`
> 
>     * add a `#endif` on the bottom
> 
>     * voila! linux has a custom progress bar and only Mac and Windows use the unified one.

My opinion:  Keep it default as it is now, the same as for MacOS and Windows -- I think it is fine for Linux.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG that's fine: I'm just pointing out what you CAN do with styles.h if the linux folks want a completely different style. Nothing there is set in stone - it was just a starting point to get it into everybody's hands so they can decide what they want for "look and feel", while making it easy to do from one central point. There's many other UI elements that could be added to it if you want.

With this design it should also be very easy for Wyatt to insert runtime QSS to override it.

Joe-K0OG:
@wmiler Test results of wmiler:styles -- All these were built and tested from source, and tested with several light and dark system themes.
* Debian 12 - OK
* Debian 13.2 - OK
* Ubuntu 25.04 - OK
* LUbuntu 24.04 - Works fine, but will not follow system theme like other distros do.  However, other apps don't either, so it's an LUbuntu OS problem, not a JS8Call problem.
* MX Linux - Works fine, but will not follow system theme like other distros do.  However, other apps do follow the system theme.
* Fedora 39 - OK
* Fedora 41 - OK
* Windows 11 24H2 - OK

In addition, I tested all these with a modified styles.h to add the mono font family to the currentFreq in the Linux section, line 283:
`" font-family: \"DejaVu Sans Mono\", \"Liberation Mono\", \"Noto Mono\", \"Ubuntu Mono\", monospace;"`

And that worked fine to provide the same monospaced font as the date/time text.

In all cases, the following warning shows during build, as I mentioned before in another place:
```
Building CXX object CMakeFiles/JS8Call.dir/JS8_UI/mainwindow.cpp.o
/home/jcounsil/development/JS8Call-improved/JS8_UI/mainwindow.cpp: In member function ‘void UI_Constructor::createStatusBar()’:
/home/jcounsil/development/JS8Call-improved/JS8_UI/mainwindow.cpp:1451:18: warning: variable ‘textColor’ set but not used [-Wunused-but-set-variable]
      |                  ^~~~~~~~~
[  5%] Linking CXX executable JS8Call
[100%] Built target JS8Call
```

73,
-Joe-
K0OG

wmiler:
> @wmiler Test results of wmiler:styles -- All these were built and tested from source, and tested with several light and dark system themes.
> 

Thanks, that's helpful!

> * MX Linux - Works fine, but will not follow system theme like other distros do.  However, other apps do follow the system theme.

Hmm, I wonder what's different about MX? I'll have to do some research on that.

> 
> In addition, I tested all these with a modified styles.h to add the mono font family to the currentFreq in the Linux section, line 283: `" font-family: \"DejaVu Sans Mono\", \"Liberation Mono\", \"Noto Mono\", \"Ubuntu Mono\", monospace;"`
> 
> And that worked fine to provide the same monospaced font as the date/time text.
> 

So you think everybody wants monospace there? I can added it, if it looks better that way.

> In all cases, the following warning shows during build, as I mentioned before in another place:
> 
> ```
> Building CXX object CMakeFiles/JS8Call.dir/JS8_UI/mainwindow.cpp.o
> [snip]
> ```
> 
> 73, -Joe- K0OG

Yeah, that's just an unused var that somebody stuck in there at some point.

Chris-AC9KH:
> In all cases, the following warning shows during build, as I mentioned before in another place:
> 
> ```
> Building CXX object CMakeFiles/JS8Call.dir/JS8_UI/mainwindow.cpp.o
> /home/jcounsil/development/JS8Call-improved/JS8_UI/mainwindow.cpp: In member function ‘void UI_Constructor::createStatusBar()’:
> /home/jcounsil/development/JS8Call-improved/JS8_UI/mainwindow.cpp:1451:18: warning: variable ‘textColor’ set but not used [-Wunused-but-set-variable]
>  1451 |     const QColor textColor = this->palette().color(QPalette::WindowText);
>       |                  ^~~~~~~~~
> [  5%] Linking CXX executable JS8Call
> [100%] Built target JS8Call
> ```

That was left in as a fallback when I removed the QStyleFactory::create("Fusion") experiment. gcc will complain about it because you have -Wunused-but-set-variable turned on in your compiler flags. gcc is a little brain dead in this regard because it's a const variable that is not set. clang is a little smarter realizing it is there for breakpoint purposes.

Chris-AC9KH:
I'll add that if that line of code is worrisome, I can remove it. But if using qss injection if you do not define every last item, it will rely on palette queries. So I'll leave it up @wmiler as to whether or not he wants to remove that line of code. It's all going to depend on how thorough you are with designing your third-party qss stylesheets. You can comment that line out and try it, and if it works, then you can remove it. Otherwise it's there for fallback.

Chris-AC9KH:
FYI, I left one other breakpoint in the code. And that is with the up/down.png images in artwork. They are no longer used because I switched to unicode instead of using those. But I left them defined in CMakeLists.txt until we are sure the unicode characters display properly on linux.

So upon completion of this PR, the images themselves, and the definition in CMakeLists.txt can be removed if everything is ok with the unicode.

Joe-K0OG:
> So you think everybody wants monospace there? I can added it, if it looks better that way.

@wmiler I really don't care either way.  It was fine the way it was.  I'll defer to someone else's opinion on that.  However, it *does* make a little bit of sense to make in mono to be the same as the date/clock font, and to agree with the default MacOS and Windows fonts.  I'm good either way though.

73,
-Joe-
K0OG

Joe-K0OG:
> But I left them defined in CMakeLists.txt until we are sure the unicode characters display properly on linux.

In all my Linux variant testing today, they displayed fine, so I think you are good to remove those if you want.

73,
-Joe-
K0OG

Chris-AC9KH:
Ok, @wmiler can remove them as part of this PR. I have a request for Wyatt though. On Windows 11 there is no contrast between the title bar, menu bar (which MacOS does not have) and the header bar. I verified this with the Digital Color Meter on Mac. So I changed the Windows header bar to a baby-blue Windows 11 color to give it some contrast. If you could change these two lines for me in styles.h would appreciate it.....

Screenshot 2026-04-16 at 19 03 37

Screenshot 2026-04-16 at 19 04 02

Chris-AC9KH:
@wmiler I added a commit to this PR to fine-tune some elements on Mac and Windows.

Chris-AC9KH:
@wmiler @rruchte @Joe-K0OG

one consideration for your linux design is that I changed the macro button bar to be stretchable only from a minimum of 25px to maximum 40px. This will accommodate different linux button styles - the user just has to resize it for whatever buttons he/she is using. The QPushButton height is standardized at 15px for Mac and Windows so they both use the 25px minimum that gives 10px buffer top and bottom. There is nothing set for linux as far as I know but the stretchable limit should allow anything the linux guys come up with to be displayed properly by moving the QSplitter handle for the macro button bar.

Joe-K0OG (approved):
I don't find any problems on either Windows or Linux.  Seems to build and run, and resize, just fine.

73,
-Joe-
K0OG

Chris-AC9KH:
Great, I'll merge this one. Any further tweaks can be applied any time. I got another one ready that replaces the offset frequency label in the header bar with a frequency offset slider for Mac only. With this merged I can rebase that one and get it in.

PR #261, opened 2026-04-19
New Design of Frequency Offset Control (Mac only)

Chris-AC9KH:
This cleans up the old artwork and code that was switched to unicode. Cleans up the associated compiler warning. It doesn't clean up the compiler warning from Rob's PR but I didn't know if I should remove those variables or not, so I didn't. Sometimes developers leave that stuff in the code for diagnostics, testing or fallback even though it gives non-developers a jolt and they think something is horribly wrong in the code.

The old label style frequency offset setter can be set up to 999.999999 MHz which puts a HF radio into the gigahurts range. Since this might be a feature for subspace communications I didn't do anything about it.

Screenshot 2026-04-19 at 07 40 46

So for Mac I designed a new QSlider that replaces it and has a range of 0-3000 Hz, which is normal sideband audio range.

Screenshot 2026-04-19 at 08 35 40

It took me a little while to figure out how to replace a QLabel in a .ui file with a QWidget on-the-fly in C++. But I managed to make it work so this does not affect Windows or Linux. Only Mac has the slider. I did test it on Windows 11 and it works fine. It does not work on linux. Linux can use the slider for input, but it does not communicate with the back end two ways. So if the offset is changed via the waterfall or context menu item, or by sending a HB, on linux the slider does not move. This is a problem in linux in other software too, like wfview where it doesn't work on linux for SDR's where you move a knob on the radio and the slider in the software does not move to the new position.

So this is very easy to implement in Windows by simply adding the new OffsetSliderWidget class to the Windows section in styles.h and adding Windows to the preprocessor guards for it. But I left Windows as-is for now.


Chris-AC9KH:
I'll add that this pretty much completes the UI overhaul I had envisioned for JS8Call 3.0. I've heard from people that say they got one of the newer versions and it still looks like it did before - what's actually changed? Well, there should be no doubt now that 3.0 is not the same as 2.2.0.

So after this, everybody can return to their regularly scheduled programming.

Chris-AC9KH:
I added inline comments in styles.h for the namespace Styles for the linux section that better explains what each item defines. Hopefully, this will make it easier for the qss folks to figure out what they need to define in their custom StyleSheets on linux. The header bar is handled completely different from the rest of the UI for background colors, buttons, etc. to make sure it doesn't get broken by dark mode theming. The doxy blocks are great for code documentation. But for people trying to design custom style sheets it may not be immediately apparent what each of those items sets.

Chris-AC9KH:
And I added a little enhancement that locks the appearance of the slider widget for dark mode theming. While this is not a problem on MacOS, IF somebody duplicates the code for the Windows section this will prevent Windows 11 from breaking it in Dark Mode.

Chris-AC9KH:
I just verified that if you set a time drift it still will display in the new display. And it does. I had not thought to verify that until now, and wasn't sure if anybody else had verified it either.

Screenshot 2026-04-19 at 13 13 49

Joe-K0OG:
Concerning the offset slider:

It would be nice if you could still click on the number, get a pop-up and enter a number

It would be nice to figure out how to implement this on Linux so that all three platforms could benefit

Chris-AC9KH:
@Joe-K0OG bringing the popup out from the slider widget is doable, but not good UI design. Mac users expect clean, well-thought-out designs that don't have a bunch of a popups, right click context menus, buttons and switches et al. While that kind of thing might fly on linux, it don't make it out of the gate on Mac.

With the slider you don't need anything like that. You can simply mouse over it, and use your scroll wheel or trackpad scrolling to set it to what you want. It is actually REALLY nice if you choose to not display the waterfall and use the waterfall in your SDR software instead. Which a lot of people on Mac do because they got either FlexRadios or newer Icom's with the waterfall display. So they use SmartSDR (FlexRadio), SDR-Control (Icom, only available for Mac or iOS) or WFView (open-source alternative to SDR Control). The Frequency Offset slider has been one of the most-requested features for Mac (from all the emails I get).

So at this point I implemented it on Mac only. If the Windows developers want it, they can easily add it to Windows. If the Linux folks want it, same thing - they'll have to figure out why the QSlider class doesn't have duplex communications on linux.

Chris-AC9KH:
Come to think of it, there might be other problems in linux with slider controls not moving to the place where the hardware is set at. On KDE, for instance, I note that when I set a volume setting with a slider it never "saves" it. Next time I start up KDE guess what? The volume is at zero again. And that's a QSlider class widget in KDE.

Same with WFView. Start it up on linux and all the sliders are at zero and do not match the rig settings. Move the volume knob on the rig, and the volume slider in WFView does not move. On Mac and Windows this problem does not exist.

So I suspect there might be only a partial implementation of the class on linux. I'm not sure. Maybe it takes a code "hack" to work around it. I'm not sure.

Chris-AC9KH:
I'm going to merge this. MacOS compiles completely different code from Windows or Linux here, and this does not affect Windows or Linux. The code is there if other developers want to implement it on the other platforms. The reason for this is that not everybody runs the "standard" JS8Call layout. Some prefer layouts like this:

Screenshot 2026-04-19 at 18 56 39


Starting tomorrow I'll be gone for about a month. Well, not gone gone. But wife is starting her spring wood processing for her firewood business so the wood is cured and ready for sale next fall. We got about 2,700 cord of wood to go thru in a month. I spend 12-14 hrs a day just keeping equipment running - if it's not a chainsaw or processor with a burnt up bar and chain, it's a track loader with a blown hydraulic hose that lubricated everything in a 30ft radius. So we're looking forward to long days at the log yard for the next month, not gonna have much time to work on JS8Call. But I've pretty much finished my project list for 3.0.

Chris-AC9KH:
I was studying the code last night and found the problem. There is three circular reference functions in mainwindow.cpp that have a potential memory leak and are not thread safe. They deal with how the frequency offset is reset after a HB event by setting booleans and it never releases the memory allocation after it runs. This whole circular reference was put into place to allow HB outside the HB subchannel and was obviously added on as an afterthought. It is using the waterfall cursor because the label itself has no control over the offset - only the popup frequency setter has any control over it.

When I designed the new frequency offset slider I did it differently, bypassed all that by compiling different code for MacOS. I used a simple variable that is set before the HB system changes the offset (or doesn't change it, whatever the case may be). On stopTx() I restore the frequency by feeding the value of the stored variable to the slider, which controls the waterfall cursor, then clear the variable. No booleans left hanging in various switch positions required.

The old code is throwing a bunch of switches and nobody is keeping track of the switch positions. They're left thrown in whatever on/off position they're in and we'll figure it out later for the next time. I would say this needs to be fixed before 3.0 is released, and was the reason linux don't work. The easiest way to fix it it going to be to apply the offset slider to all three platforms and get rid of that old circular reference switch-throwing code. This is the problem with using too many booleans to set the state - impossible to keep track of the logic because you don't know what position the switches are in.

I don't know if I'll get time to look at this for the next month. But even if the linux folks DON'T want the slider, the base switch-throwing code needs to be fixed.

PR #265, opened 2026-04-22
Apply offset frequency slider widget across all platforms

Chris-AC9KH:
This adds the offset frequency slider widget to Windows and linux, along with fixes in the code that previously broke it on linux.

### DIFF
--- a/JS8_Mainwindow/UI_Constructor.cpp
+++ b/JS8_Mainwindow/UI_Constructor.cpp
@@ -51,7 +51,7 @@ UI_Constructor::UI_Constructor(QString const &program_info,
       m_appDir{QApplication::applicationDirPath()}, m_palette{"Linrad"},
       m_txFrameCountEstimate{0}, m_txFrameCount{0}, m_txFrameCountSent{0},
       m_txTextDirty{false}, m_driftMsMMA{0}, m_driftMsMMA_N{0},
-      m_previousFreq{0}, m_hbInterval{0}, m_cqInterval{0}, m_hbPaused{false},
+      m_hbInterval{0}, m_cqInterval{0}, m_hbPaused{false},
       m_msAudioOutputBuffered(0u),
       m_framesAudioInputBuffered(JS8_RX_SAMPLE_RATE / 10),
       m_audioThreadPriority(QThread::HighPriority),
@@ -456,11 +456,8 @@ UI_Constructor::UI_Constructor(QString const &program_info,
             return true;
         },
         this));
-      
-#if defined(Q_OS_MACOS)
+
     freqOffsetWidget = new Styles::OffsetSliderWidget(nullptr);
-    // Find the label's parent layout and replace it on Mac to display a
-    // QSlider widget instead of a QLabel
     QLayout *parentLayout = ui->labDialFreqOffset->parentWidget()->layout();
     if (auto *vbox = qobject_cast<QVBoxLayout *>(parentLayout)) {
         int index = vbox->indexOf(ui->labDialFreqOffset);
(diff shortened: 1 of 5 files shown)


Joe-K0OG (approved):
This works great on both Linux and Windows.  The slider handle (MacOS style) looks good on both platforms, and fits will with the overall look & feel.  The slider can be generally-tuned by clicking the slider and pulling the handle left/right.  Clicking on the waterfall causes the handle to follow, as it should.  Fine tuning to 1Hz resolution is possible by Left/Right or Up/Down arrows on the keyboard.  During Split TX, the frequency and handle follow as they should.  Nice functional piece of eye candy ;)  Thanks Chris!

73,
-Joe-
K0OG

rruchte (approved):
This works well. Page Up/Page Down moves by increments of 10, so it's pretty easy to click to get in the neighborhood, then add a few keystrokes get to the exact offset you want without fiddling too much.

I think it would be cool to have the slider be quantized by a larger amount and snapped to 10s when a key (shFit maybe) is down. That's not something QT gives you out of the box though, we would have to subclass QSLider, but it's definitely possible, maybe a nice-to-have for later.

Joe-K0OG:
> This works well. Page Up/Page Down moves by increments of 10, so it's pretty easy to click to get in the neighborhood, then add a few keystrokes get to the exact offset you want without fiddling too much.
> 
> I think it would be cool to have the slider be quantized by a larger amount and snapped to 10s when a key **(shiFt maybe)** is down. That's not something QT gives you out of the box though, we would have to subclass QSLider, but it's definitely possible, maybe a nice-to-have for later.

Hey, watch your language Rob Ha, ha, ha ha!!!

I didn't notice the PgUp/PgDwn, and Home/End key functions.  Nice!

73,
-Joe-
K0OG

rruchte:
@Joe-K0OG Oops, LOL

Chris-AC9KH:
I figured people would like it once they tried it out. @rruchte I did not replace the QLabel in mainwindow.ui. Instead I chose to leave it there as a placeholder, hide it and replace it with the QSlider widget in code, done in the UI_Constructor. And I did make a class out of it in the Styles namespace so it is quite configurable for the future without having to mess with XML code.

Chris-AC9KH:
> Fine tuning to 1Hz resolution is possible by Left/Right or Up/Down arrows on the keyboard.

Yes, that works. You can also tune it with the trackpad (or mouse scroll wheel). I haven't used a mouse for years - I use the Apple desktop trackpads with everything, including my iPad. You don't have to click on the handle - simply mouse over and scroll left or right and it can be tuned in 1Hz increments.

rruchte:
@Chris-AC9KH Oh, that's really nice. I too am a Magic Trackpad enjoyer.

Chris-AC9KH:
@rruchte Apple builds the best desktop trackpads there is, bar none. And they're not cheap. But once you make the transition it's pretty hard to use a mouse again.

But anyway, if you want to add PC-specific keyboard shortcuts you should be able to do that without affecting Mac. The Mac keyboard is different from PC and the MacOS UI is only partially functional when using a PC keyboard. Keyboard shortcuts between the two are not compatible, especially when using the option and Command keys on Mac vs alt and Windows key on a PC keyboard. So the OffsetSliderWidget class should be fully customizable with custom keyboard commands on PC with the preprocessor guards used in styles.h without breaking anything on Mac. If the linux folks want a more vertical handle that is slightly taller than it is wide, you can do that. The three platforms no longer have to be the same, since they are not the same, which is the advantage of defining UI elements in code rather than XML in the .ui files. So if you open mainwindow.ui in Qt Designer, it will not show a QSlider widget where the slider is. It will show a QLabel, which is merely a placeholder for the general UI layout. Everything defining the slider is done in code instead.

Issue #260, opened 2026-04-19
Multiple notification audio connections left open


JohBod:
**Describe the bug**
The connection to the notification soundcard is properly opened and closed when a single notification occurs. When multiple notifications occur in rapid succession (overlapping), for example when receiving many responses to a HB, a number of connections are opened but only the _last_ used connection is closed. It becomes a mess of unused connections after a while. I can only get rid of them by closing and restarting JS8Call.

**To Reproduce**
Enable notifications and receive a bunch of notification triggering messages at the same time. 

**Expected behavior**
No connections to the notification soundcard after the last notification sound has ended.

**Screenshots**

<img width="1071" height="1011" alt="Image" src="https://github.com/user-attachments/assets/56563bca-0a01-48b2-b659-99e5e0102f2f" />

**Desktop (please complete the following information):**
 - OS: Linux Mint 22.3
 - Version: v2.5.2 official appimage from https://github.com/JS8Call-improved/JS8Call-improved/releases
 - Radio: IC-7300 Mk2
 - Direct to radio: Yes
 - via FLrig or others: No

**Additional context**
I don't know if this has any adverse effect but maybe there is a risk of running out of handles somewhere or perhaps some table will run out of space?

73 Johan SM6LKM

Chris-AC9KH:
I'm not sure what to make of this. I don't know if it's an issue or not. Perhaps @rruchte could comment since he did some recent work on audio for linux.

rruchte:
I can take a look, I spent some time looking at the notifications code while diagnosing the notification playback speed issue.

rruchte:
I was not able to reproduce the dangling audio connections on my system, but I was able to consistently induce a segfault by rapidly playing the notifications in Settings/Notifications, which is probably another symptom of the same problem. I have a potential fix, and will post a PR momentarily.

Chris-AC9KH:
@JohBod we think we have this issue fixed in master. If you are able to build the code and test the current master branch, that would be a more definitive test since none of us were able to duplicate your issue.

JohBod:
I don't think that I have enough knowledge to build it with ease and without frustration :) As the change is already merged into master I suppose it will appear in the next public release? If so, I will be happy to wait for the next appimage release and test it then. 73

Chris-AC9KH:
@JohBod JS8Call v3.0 will be coming out within the next 30 days. I will leave this open for now, please remember to report back and close this issue if v3.0 fixes it.

[...]

PR #271, opened 2026-04-25
Addresses issue #260 Multiple notification audio connections left open

rruchte:
Refactor notification sound playback to prevent orphaned audio connections when simultaneous notifications sounds are played.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,6 +245,7 @@ target_sources(
   JS8_Audio/AudioDevice.cpp
   JS8_Audio/BWFFile.cpp
   JS8_Audio/NotificationAudio.cpp
+  JS8_Audio/NotificationSoundOutput.cpp
   JS8_Audio/SoundInput.cpp
   JS8_Audio/SoundOutput.cpp
   JS8_JSC/JSC.cpp
--- a/JS8_Audio/NotificationAudio.cpp
+++ b/JS8_Audio/NotificationAudio.cpp
@@ -19,10 +19,10 @@ Q_DECLARE_LOGGING_CATEGORY(notificationaudio_js8)
  * @return None.
  */
 NotificationAudio::NotificationAudio(QObject *parent)
-    : QObject{parent}, m_stream{new SoundOutput} {
-    connect(m_stream.data(), &SoundOutput::status, this,
+    : QObject{parent}, m_stream{new NotificationSoundOutput} {
+    connect(m_stream.data(), &NotificationSoundOutput::status, this,
             &NotificationAudio::status);
-    connect(m_stream.data(), &SoundOutput::error, this,
+    connect(m_stream.data(), &NotificationSoundOutput::error, this,
             &NotificationAudio::error);
 }
 
@@ -115,17 +115,9 @@ void NotificationAudio::stop() { m_stream->stop(); }
 /******************************************************************************/
 
 void NotificationAudio::playEntry(Cache::const_iterator const it) {
-    if (m_buffer.isOpen())
-        m_buffer.close();
-
     auto const &[format, data] = *it;
-
-    m_buffer.setData(data);
-
-    if (m_buffer.open(QIODevice::ReadOnly)) {
-        m_stream->setDeviceFormat(m_device, format, m_msBuffer);
-        m_stream->restart(&m_buffer);
-    }
+    m_stream->setDevice(m_device, m_msBuffer);
+    m_stream->play(data, format);
 }
 
 /**
--- a/JS8_Audio/NotificationAudio.h
+++ b/JS8_Audio/NotificationAudio.h
@@ -1,6 +1,8 @@
 #ifndef NOTIFICATIONAUDIO_H
 #define NOTIFICATIONAUDIO_H
 
+#include "NotificationSoundOutput.h"
+
 #include <QAudioDevice>
 #include <QBuffer>
 #include <QByteArray>
@@ -32,7 +34,7 @@ class NotificationAudio : public QObject {
     static QByteArray pcm24le_to_int32le(const QByteArray &in);
     static bool upmixMonoToStereoInPlace(QAudioFormat &fmt, QByteArray &data);
 
-    QScopedPointer<SoundOutput> m_stream;
+    QScopedPointer<NotificationSoundOutput> m_stream;
     Cache m_cache;
     QAudioDevice m_device;
     QBuffer m_buffer;
--- a/JS8_Audio/NotificationSoundOutput.cpp
+++ b/JS8_Audio/NotificationSoundOutput.cpp
@@ -1,0 +1,176 @@
+/**
+ * @file NotificationSoundOutput.cpp
+ * @brief Self-contained audio output for notification sounds.
+ *
+ * Designed for fire-and-forget playback of buffered audio data.
+ * Manages its own buffer lifecycle and releases the audio device
+ * automatically when playback completes.
+ */
+
+#include "NotificationSoundOutput.h"
+
+#include <QAudioSink>
+#include <QBuffer>
+
+NotificationSoundOutput::NotificationSoundOutput(QObject *parent)
+    : QObject(parent)
+{
+}
+
+NotificationSoundOutput::~NotificationSoundOutput()
+{
+    if (m_sink) {
+        disconnect(m_sink.get(), nullptr, this, nullptr);
+        m_sink->stop();
+    }
+}
+
+/**
+ * @brief Sets the audio device and buffer size.
+ * @param device The QAudioDevice to use.
+ * @param msBuffer The buffer size in milliseconds.
+ */
+void NotificationSoundOutput::setDevice(QAudioDevice const &device,
+                                        unsigned const msBuffer)
+{
+    m_device = device;
+    m_msBuffer = msBuffer;
+}
+
+/**
+ * @brief Sets the attenuation in dB.
+ * @param a The attenuation value (0 = full volume).
+ */
+void NotificationSoundOutput::setAttenuation(qreal const a)
+{
+    Q_ASSERT(0.0 <= a && a <= 999.0);
+    m_volume = qPow(10.0, -a / 20.0);
+
+    if (m_sink)
+        m_sink->setVolume(m_volume);
+}
+
+/**
+ * @brief Plays audio from the provided data and format.
+ * @param data The raw PCM audio data.
+ * @param format The QAudioFormat describing the data.
+ *
+ * Any currently playing notification is stopped and replaced.
+ */
+void NotificationSoundOutput::play(QByteArray const &data,
+                                   QAudioFormat const &format)
+{
+    if (m_sink) {
+        disconnect(m_sink.get(), nullptr, this, nullptr);
+        m_sink->stop();
+    }
+
+    if (m_buffer) {
+        m_buffer->close();
+        m_buffer.reset();
+    }
+
+    if (m_device.isNull()) {
+        Q_EMIT error(tr("No audio output device configured."));
+        return;
+    }
+
+    if (!format.isValid()) {
+        Q_EMIT error(tr("Requested audio format is not valid."));
+        return;
+    }
+
+    if (!m_device.isFormatSupported(format)) {
+        Q_EMIT error(tr("Requested audio format is not supported on device."));
+        return;
+    }
+
+    if (data.isEmpty()) {
+        Q_EMIT error(tr("Audio data is empty."));
+        return;
+    }
+
+    m_buffer.reset(new QBuffer);
+    m_buffer->setData(data);
+
+    if (!m_buffer->open(QIODevice::ReadOnly)) {
+        m_buffer.reset();
+        Q_EMIT error(tr("Unable to open audio buffer."));
+        return;
+    }
+
+    // Only create a new sink if we don't have one, or if the
+    // format has changed since the last playback.
+    if (!m_sink || m_currentFormat != format) {
+        m_sink.reset(new QAudioSink(m_device, format));
+        m_sink->setVolume(m_volume);
+        m_currentFormat = format;
+
+        if (m_msBuffer > 0) {
+            m_sink->setBufferSize(
+                m_sink->format().bytesForDuration(m_msBuffer));
+        }
+    }
+
+    connect(m_sink.get(), &QAudioSink::stateChanged,
+            this, &NotificationSoundOutput::handleStateChanged);
+
+    m_sink->start(m_buffer.get());
+
+    if (m_sink->error() != QAudio::NoError) {
+        Q_EMIT error(tr("Failed to start audio output."));
+        m_sink->stop();
+        if (m_buffer) {
+            m_buffer->close();
+            m_buffer.reset();
+        }
+    }
+}
+
+/**
+ * @brief Stops playback and releases all resources.
+ */
+void NotificationSoundOutput::stop()
+{
+    release();
+}
+
+// Private
+void NotificationSoundOutput::release()
+{
+    if (m_sink) {
+        disconnect(m_sink.get(), nullptr, this, nullptr);
+        m_sink->stop();
+    }
+
+    if (m_buffer) {
+        m_buffer->close();
+        m_buffer.reset();
+    }
+}
+
+/**
+ * @brief Handles state changes from the QAudioSink.
+ * @param newState The new audio state.
+ */
+void NotificationSoundOutput::handleStateChanged(QAudio::State const newState)
+{
+    switch (newState) {
+    case QAudio::ActiveState:
+        Q_EMIT status(tr("Active"));
+        break;
+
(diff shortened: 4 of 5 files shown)


Joe-K0OG (changes requested):
@rruchte This source has an adverse effect on Windows audio.  There is a strange "buzzing" sound in the audio output to the radio.  

This is not observed in the audio output to the radio on a Linux build with this source, nor with the JS8Call-improved official source on either platform.

I don't use audio notifications, so that is all disabled on my setups, hence I've never noticed the reported problem with connections left open.

This is definitely a no-go since it breaks Windows audio.

73,
-Joe-
K0OG

rruchte:
@Joe-K0OG Rats.

Chris-AC9KH (approved):

Chris-AC9KH:
@rruchte I don't know what to think of this. I believe this was/is a linux-only issue. The Mac and Windows audio backends are generally more robust and tolerant of redundant state transitions. Linux audio is problematic and has been since forever. It tends to spit out additional state transitions that MacOS and Windows do not and it doesn't look like any of this got "fixed" with PipeWire based on the graph that was posted in the original issue.

It LOOKS like this SHOULD work since releaseStreamOnly() now does disconnect before stop() and reset(), which breaks the signal chain before any state transitions? While I don't personally use notification audio and I haven't tested this on MacOS, I fail to see where it will break anything. I added a Linux label to this because I don't believe it ever was a problem on Mac or Windows.

rruchte:
@Joe-K0OG Can you test the latest changes when you get a chance? Unfortunately I don't have a Window system I can test on.

@Chris-AC9KH Yeah, Linux audio is a horror show.

Chris-AC9KH:
@rruchte if you're comfortable with this, and it passes Joe's testing, go ahead and merge it. I don't see any issues with it at all on MacOS.

Chris-AC9KH:
> @Chris-AC9KH Yeah, Linux audio is a horror show.

@rruchte I think the problem on linux is that every call to restart() creates a new QAudioSink and connects stateChanged to handleStateChanged. Linux audio fails to tear this down and saves it forever. So you'll have a dangling connection from the previous sink's lifetime firing into handleStateChanged, spitting out random "idle" status. Why linux does this, I do not know. Such a thing would be considered broken in MacOS CoreAudio just from playing simultaneous streams. How do they get advanced 256 channel audio mixers to work on linux that are similar to Logic Pro on Mac? Or maybe they don't.

rruchte:
@Chris-AC9KH They just have to take more responsibility for managing the connections, I guess. I don't know, I have to keep an old Mac around to do pro audio because Ableton doesn't run on Linux, and neither does my MOTU interface. Seriously considering switching back, TBH.

Chris-AC9KH:
Well, we've been whittling away at these linux audio issues, and it's way better than it was. I built this this morning and ran it on MacOS, and as I suspected it works perfectly fine. I had JS8Call playing AC/DC and Steppenwolf at the same time. That was a weird duet but it doesn't do anything bad.

I could test it on Windows 11 as well. But my Windows setup consists of running in the UTM virtual machine on Mac (Apple silicon so it's arm64 Windows). The VM supplies "drivers" to Windows so it can use the Mac native hardware, which may not be as good of a test as what Joe can do on regular PC hardware. The VM's "drivers" defer to Mac CoreAudio to make it work so it might actually make something work on Windows that won't work on a real PC. So we'll have to rely on @Joe-K0OG testing to make sure this is actually ok on Windows 11.

Joe-K0OG:
@rruchte Well, Rob, it's still a disaster on Windows.  Emits a "buzzing" sound and almost sounds as if there are multiple phase-shifted audio streams playing at once.  I have notifications disabled, I'm just referring to the primary JS8 signal.

73,
-Joe-
K0OG

P.S.  I also tried with a fresh settings - same problem - so it's not a settings issue, but a fundamental audio code issue.

rruchte:
@Joe-K0OG OK, well, I'm out of my depth here then. If I had a way to test on Windows I might try to fix it, but at this point someone who knows what they're doing with Windows audio needs to take this over or take a different approach.

Chris-AC9KH:
Well, darn. This code looked really good to me. The logic of it makes sense. Why would this upset the Windows WASAPI audio back end? Doesn't make sense.

@rruchte give me a day or so to experiment with this on Windows 11 in my VM. Unfortunately, working on this project I've learned more about Windows than I ever wanted to know. Let me see if I can identify the issue. It has to be something fairly simple.

rruchte:
@Chris-AC9KH Sounds good. Typically I would keep at it, but this is a critical piece of the app, and if I can't be confident in the changes, I want someone with more experience to take it over.

Chris-AC9KH:
@rruchte I can't say I'm an expert on audio either. But there is some Windows audio trouble-shooting tools that might identify what's going on. Like I said, the code looks good to me, just doesn't make sense.

Joe-K0OG:
@rruchte Rob, send me your Windows build installer with #260 included so I can see if your build performs any differently than my build on JTSDK64-Tools.  Maybe some library like ffmpeg is misbehaving?

73,
-Joe-
K0OG

rruchte:
@Joe-K0OG I would just use the one from the CI build:

https://github.com/JS8Call-improved/JS8Call-improved/actions/runs/24943875752/artifacts/6643545401

Joe-K0OG:
> @Joe-K0OG I would just use the one from the CI build:
> 
> https://github.com/JS8Call-improved/JS8Call-improved/actions/runs/24943875752/artifacts/6643545401

@rruchte Rob, I poked around earlier and couldn't find it!  I downloaded the CI build, installed it, and indeed it exhibits the same "buzzing" and hiccuping audio.

When you test, go to JS860 and send several separate frames.  It usually seems good on the first couple of frames, then starts breaking up and "buzzing" on subsequent transmissions.  Reinstalling from a build with unmodified code gives smooth audio output under the same conditions.

This doesn't seem to cause any problem on Linux, but it sure messes up Windows!

73,
-Joe-
K0OG

rruchte:
@Joe-K0OG Yeah, it's pretty hard to find, buried in the output of the action:

Screenshot From 2026-04-27 12-37-18

I tried it on the only Windows instance I have access to, a VM with shaky USB and audio integration with the host. I can't get sound to play at all from either this build or the release, which is why I'm hesitant to pursue it further. Maybe I'll put some effort into getting a working instance this evening.

Joe-K0OG:
@rruchte Rob, sometimes I do Win11 testing on a Win2Go installed on a fast USB3 memory stick.  Using Rufus, you can install Win2Go from a Win11 ISO file.  Works pretty well.

There are numerous instructions on how to do it on line.  Here is one that is pretty close to what I do.  The only difference is that I tick the box to have it create a local account, so that's all set up during the installation and saves you the trouble of doing it after initial boot from the USB stick.

https://coastipc.com/blog/post/creating-a-portable-windows-to-go-operating-system-on-a-usb-drive

73,
-Joe-
K0OG

P.S.  SUPER-IMPORTANT Warning:  Be sure to DISABLE ACCESS TO LOCAL DRIVES in the setup checkboxes.  This may corrupt any existing installation.  Don't ask me how I know....  I was given opportunity to rebuild a PC last year...  Ugh!

Chris-AC9KH:
@Joe-K0OG where are you picking up the buzzing or phase-shifted audio? Are you looking at the audio waveform on a scope, or playing back thru the speakers?

Joe-K0OG:
> @Joe-K0OG where are you picking up the buzzing or phase-shifted audio? Are you looking at the audio waveform on a scope, or playing back thru the speakers?

Through speakers.  Easiest way to test.

73,
-Joe-
K0OG

Joe-K0OG:
@rruchte @Chris-AC9KH I just did a test on another completely different PC (Windows 11) with built-in audio chipset, and get the same result.  The unmodified JS8Call emits nice clean audio, the #260 modified version emits a "buzzing" sound after a few frames, and even hiccups occasionally.  So, it is repeatable on multiple hardware platforms.

73,
-Joe-
K0OG

Chris-AC9KH:
@rruchte ok, so verified on Windows 11 (arm64) running on Mac. Looking at the new code and the old code I think the problem is in SoundOutput.cpp 

`void SoundOutput::restart(QIODevice *source)` on line 102. releaseStreamOnly() calls m_stream->stop() before replacing the sink. This is the main audio path for the JS8 signal, NOT the notification audio path. Recorded the audio output with Logic Pro on Mac and analyzed the waveform to see where it's coming from. Windows is not fully releasing the audio on stop(). A new QAudioSink is immediately constructed for the same audio device, Windows opens a second handle while the first is still closing, resulting in both being active briefly on the same output. It is two slightly-offset copies of the same signal summed together. There actually IS a reason Macs absolutely rule in the world of professional audio and studio recording :-)

The original code didn't have this problem on the main audio path because it called m_stream.reset(new QAudioSink() directly, which destroys the old sink implicitly via the QScopedPointer reset before constructing the new one.

So on line 235 don't call stop() expressly. Let QScopedPointer::reset() destroy the sink. Get rid of that line. Even if a stale signal somehow slips through on Linux between the disconnect and the reset, it'll be dropped safely in the queued handler because m_stream.data() != finishedSink will be true by then.

I think just removing that line of code on 235 will fix it.

rruchte:
@Chris-AC9KH Ah, right, that makes sense.

@Joe-K0OG Can I impose upon you yet again to try the Windows build? Third verse, same as the first, a little bit louder and a little bit worse. https://github.com/JS8Call-improved/JS8Call-improved/actions/runs/25015249521/artifacts/6669368723

Chris-AC9KH:
@rruchte turns out that didn't work either. Built it with that line of code removed and still got the problem on Windows 11 and the waveform did not change. It's still a slightly offset copy of the same signal summed.

restart(QIODevice *source) overload is used by the modulator. The notification path and main audio path are different. So the main path passes in an external QIODevice *source. What is the state of that source when restart is called?

I'm thinking the problem is in the modulator in the interaction between the modulator and the audio stream lifecycle. If you look in the modulator on line 127 `m_stream->restart(this);` is passing itself as the QIODevice source so we now have a double-open window. In Modulator::start() it is calling stop() but with the changes in SoundOutput it needs to ensure that the stream is fully released before it does anything.

And now in Modulator::close() we have a redundant m_stream->stop();

This one is a head scratcher. Let me cipher on this a bit........

rruchte:
@Chris-AC9KH OK, given the differences in behavior between the notifications and modulator, maybe the answer is to stop trying to use the same SoundOutput class for both, revert SoundOutput and just create a new output class for notifications.

Chris-AC9KH:
@rruchte I do actually think that would be a more pragmatic approach. The modulator needs a long-lived, externally-driven QIODevice* stream where lifecycle is managed by the caller, with rapid restart behavior.

NotificationAudio needs fire-and-forget playback of a self-contained buffer, auto-release on completion, and a  clean linux teardown that was causing the original problem.

These are different enough that forcing them through the same class is going to create tension and gnashing of teeth. The fixes for one are breaking the other. A dedicated NotificationSoundOutput class would be clean and efficient.

rruchte:
@Chris-AC9KH Yeah, agreed, I'm just going to do that.

Joe-K0OG:
> @Chris-AC9KH Ah, right, that makes sense.
> 
> @Joe-K0OG Can I impose upon you yet again to try the Windows build? Third verse, same as the first, a little bit louder and a little bit worse. https://github.com/JS8Call-improved/JS8Call-improved/actions/runs/25015249521/artifacts/6669368723

Yeah, I also verified that it still fails with a bunch of buzzing bees on this one...  I'll wait until you guys break it out into separate audio streams, then I'll test.  Let me know when ready, and if we aren't dodging severe thunderstorms or tornadoes, I'll test it.

73,
-Joe-
K0OG

Chris-AC9KH:
@rruchte since I had it all in my head about how this audio conglomeration works I wrote up a NotificationSoundOutput class and wired it up. It took me like 17 tries to fix all the compiler errors. But it builds now and it seems to work.

https://github.com/Chris-AC9KH/JS8Call-improved/tree/pr/audio_fix

You can use it if you want, maybe there's something useful in it. I don't know if it fixes the problem on linux or not. I did it fundamentally different - instead of destroying the sink and creating new ones, just re-use it if the format hasn't changed since last playback.

Some different "features"
- it will only play one notification at a time - no more multiple simultaneous playbacks
- it has a built-in buffer
- I put an attenuator in it in case you want to wire up a slider volume control in the UI to set notification volume. This is for the outside chance that you have a Mac mini with a 200w bluetooth sound system on it, you left it cranked with AC/DC:You Shook Me All Night Long in the Notifications queue, you get an incoming MSG in the middle of the nite, suddenly the house starts to shake and it sounds like Angus Young and Brian Johnson broke in and are holding a concert.

Joe-K0OG:
@Chris-AC9KH @rruchte Chris, I tested your pr/audio_fix on Windows and Linux, and as far as I can tell it works properly for both JS8 modulation audio and notification audio.  I don't have two audio outputs, so just tested on the built-in chipset, but there doesn't seem to be any conflict nor distortion of any kind for either audio, even when generated simultaneously.  The output is clear and undistorted.

Good fix?  Seems like it to me.  Someone who routinely uses audio notifications through different audio outputs ought to test it though.

73,
-Joe-
K0OG

Chris-AC9KH:
I don't know if it will be suitable for what Rob wants. It's just a draft that I got to work with some different ideas in it. In theory, since there's only one m_sink instance for the lifetime of NotificationSoundOutput, the worst case should be one lingering device handle for linux rather than the accumulating pile that the previous destroy-and-recreate approach was causing. Since we got a new class we can do this independent of the main audio path.

But I have no clue if it works or not. It would have to be tested on linux for the original problem that this PR is addressing. If it works, maybe Rob can use some of the ideas in it. Or use it as-is and expand on it. After seeing that no matter what you do to notification audio when shoving everything thru one class, I got the idea why not try something different for notification audio? I don't even know if it's the right approach.

rruchte:
@Chris-AC9KH Looks good to me, I think that's the right approach. Let's go with what you have and close this. It would be great if we could get the user who documented the open connection problem to test it, I was never able to recreate that, I just got segfaults.

Chris-AC9KH:
@rruchte since you already got a PR going here on it, if you can, just grab it and update the PR. You have to revert the code back to master, there's just some minor changes in the existing files to wire it up, then drop the new class files in. That would be simpler than creating a new PR for it. Then we preserve all the history of what we tried by attempting to shove it all thru one class.

Chris-AC9KH:
@rruchte I think it should fix the segfault issues you were having too. I'm pretty sure that was manifested by attempting to play simultaneous notification streams without shutting down the first one before another one starts. Every call to restart() creates a new QAudioSink and connects stateChanged to handleStateChanged and linux was emitting additional state transitions during device teardown. That should now be limited to just one and not allow linux to do that anymore. That's my theory anyway.

rruchte:
@Chris-AC9KH  Yeah, that's what I saw happening. I rebased on master, reverted my changes, merged yours, and squashed everything together. I think it's good to go.

Chris-AC9KH:
It would be good now if @Joe-K0OG could thrash it, and maybe @rruchte could see if the method you were using to crash JS8Call before will still crash it. I'm pretty sure the segfaults you were experiencing were related the hanging handles that the original issue poster noted. If you can still crash it, then it's not fixed.

rruchte:
I tried to induce the default but was not able to.

Joe-K0OG:
> It would be good now if @Joe-K0OG could thrash it

I cloned and checked out @rruchte issue_260... this morning on both Linux and Windows, and find that it behaves nicely, producing clean audio output from both the modulator and notification audio, even simultaneously.  I think this is good to go.  I don't know how to duplicate the original problem, but I certainly don't find any problem with this one.  Even if I bring up the notifications settings window while a transmission is happening and repeatedly/quickly press the "Test" button, nothing breaks.  Surely it would under that condition if there were going to be a problem.

73,
-Joe-
K0OG

P.S.  Maybe this is relevant to the original problem:  When I repeatedly/quickly press the "Test" button, the currently playing notification audio is abruptly stopped and the new audio initiates.  I think the point is that when a notification audio is playing, and another comes along, rather than queueing the audio, it stops the previous audio and commences playing the new one, thus not leaving any junk behind in a queue.  While all this is happening with notification audio, the modulator output is completely unaffected, and continues to emit smoothly.

rruchte:
@Joe-K0OG Excellent, thanks for testing. I am able to consistently induce segfaults in builds from master by rapidly pressing the "Test" button, this solves that problem, which I believe is the same problem that causes the hanging connections.

Chris-AC9KH:
I'll merge this then, and we'll run with it. I figured it should fix the segfault issue because I found what was causing that. I don't know about the hanging connections issue, but with the change to only allow one notification stream at a time it will limit the open handles to one, and instead of killing it and replacing it all the time it will reuse it. So I THINK it will keep linux from accumulating a bunch of open handles, time will tell on that.

PR #273, opened 2026-04-26
Fix security vulnerability

Chris-AC9KH:
These were fixed 22-element stack arrays. They get populated by two separate loops over regex matches derived from network input. There was no check that i * 2 or i * 2 + 1 stays within [0, 21]. An attacker can supply an arbitrarily long string and crash JS8Call.

This was reported here:
https://github.com/JS8Call-improved/JS8Call-improved/security/advisories/GHSA-98hp-pjp7-w62x

### DIFF
--- a/JS8_Main/APRSISClient.cpp
+++ b/JS8_Main/APRSISClient.cpp
@@ -176,10 +176,9 @@ QPair<float, float> APRSISClient::grid2deg(QString locator) {
     foreach (QStringList matched, lats) {
         char x = matched.at(1).at(0).toLatin1();
         char y = matched.at(2).at(0).toLatin1();
-
+        if (i * 2 + 1 >= 22) break;
         valx[i * 2] = x - A;
         valy[i * 2] = y - A;
-
         i++;
         tot++;
     }
@@ -188,9 +187,9 @@ QPair<float, float> APRSISClient::grid2deg(QString locator) {
     foreach (QStringList matched, lons) {
         int x = matched.at(1).toInt();
         int y = matched.at(2).toInt();
+        if (i * 2 + 1 >= 22) break;
         valx[i * 2 + 1] = x;
         valy[i * 2 + 1] = y;
-
         i++;
         tot++;
     }


Chris-AC9KH:
I had to fix the initial fix. I made a mistake where the guard exits the loop after x and y are already declared inside the loop. So I had to move the guard after those declarations.

Chris-AC9KH:
My thoughts on this are, since JS8Call 3.0 is in sight on the horizon we're not going to patch 2.5.2 for this. If JS8Call 3.0 was not so close, then I'd push out a 2.5.3 patch release. This has existed since back in the early days and to my knowledge it has never been exploited.

Chris-AC9KH:
I think the original author of this code intended the `tot` variable to do the job. However, it limits reads but does nothing to protect the write loops above it. The stack overflow happens before tot applies.

punk-kaos:
Looking at the patch, is index 21 actually invalid? It looks like its not... 

If I'm reading it right  with i = 10, the actual target index is i * 2 + 1 = 21, which is valid but the check i * 2 + 2 >= 22 stops it. Shouldn't that check be if (i * 2 + 1 >= 22)?

Maybe I'm reading it wrong....

Chris-AC9KH:
@punk-kaos oops. Little finger hit the wrong key there. I fixed that. I was going to inline comment the guards but it should be pretty obvious.

I suppose a more robust approach would be to pre-validate that the grid string fits the 12 character limit before calling grid2deg and stop it at the network layer. But there was still no index guard on the array and that will also stop it.

Chris-AC9KH:
Like I said, I don't think it's a serious issue because I've never heard of it being exploited. I thinks it's good to patch the hole in master. But I don't see where it's worth it to patch previous releases because they'd have to be patched clear back to at least 2.2.0. So IMO just put the index guard on the array for 3.0 and move on. I'm not sure how others see that.

punk-kaos (approved):

wmiler (approved):

Chris-AC9KH:
There was a private fork that was generated for this vulnerability report. There was no actual "fix" for it in that fork. The whole idea is that you're supposed to collaborate on a fix and merge it from that private fork. That seemed a little excessive to me for something so simple so I just bypassed all that and deleted the fork when I merged this PR. The original author of the private fork and the security report could've submitted a PR for it, but that wasn't done.

wmiler:
He did it the proper way, as it was a security issue, and will now be enforced that way as I crank up the security.

Joe-K0OG:
I'm late to the game here, but built & tested on Linux and Win 11.  Both build and run fine.  On my Linux machine, I have a GPS dongle, and use a Ruby script to feed 12-digit GPS to Js8Call through the API, and that also works fine.  So, this didn't introduce any problems that I can see.

73,
-Joe-
K0OG

Chris-AC9KH:
@wmiler I published the security advisory now rather than waiting for the 3.0 releases. As noted in the advisory comments it will be reference in the release notes, along with the PR that fixed it. But I do NOT like the idea of working from a private fork that later has to be merged into master. A fix should be submitted publicly as a PR so all available eyes can be on it. And not just limited to administration or a select few. This is an open-source project and we have nothing to hide, nor anything that should be worked on privately without the whole team getting a chance to take a look at it.

PR #280, opened 2026-05-03
Fix to show status of auto_reply when program is first started.

Chris-AC9KH:
This is a small fix for the Auto Reply label. When the program is first started and automatically enable auto reply on startup is not enabled in the settings, it would show a blank label until the setting was toggled. This fixes that.


wmiler (approved):

Joe-K0OG (approved):
Yes @Chris-AC9KH - Good fix!  Works fine here.

73,
-Joe-
K0OG

PR #281, opened 2026-05-03
Replace UI lockup code that prevented a MSG from being composed with

### LINKED ISSUE #276: More than 5 of the Same Character TX Window Hangs JS8Call
If any single character is typed more than five times without spaces in the TX text window, then Sent, it hangs JS8Call.

**To Reproduce**
Open JS8Call
Type five of any character, number, symbol, (but all the SAME character) without any spaces, and click Send (or press Enter) to transmit.  It will transmit without problem.  However, type six of any character, number, symbol without any spaces and Send, and JS8Call will fail to send, and "hang" so no further input is necessary, requiring that JS8Call be closed and re-open, possibly even leaving a hung JS8 process.  For example, type "AAAAA" (five "A") and it will send, but "AAAAAA" (six "A") and it will not send.  However "AYAYAYAYAY" (mixed characters of any length) will send without problem.

**Expected behavior**
Typing ANY string of ANY length without spaces should still send normally over several frames.

Verified on Linux Mint 22.3 and Win11 24H2.

This is an oddball problem because it would be rare for six of the same character to be typed as a valid message, but I think this should be fixed anyway because it might inadvertently lead to an occasional hang.

73,
-Joe-
K0OG

Chris-AC9KH:
with a string that contains a word with more than 5 repeated characters with a warning popup message

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/276

### DIFF
--- a/JS8_UI/mainwindow.cpp
+++ b/JS8_UI/mainwindow.cpp
@@ -3421,19 +3421,11 @@ bool UI_Constructor::ensureCreateMessageReady(const QString &text) {
 
     if (!ensureKeyNotStuck(text)) {
         on_stopTxButton_clicked();
-
-        ui->monitorButton->setChecked(false);
-        ui->monitorTxButton->setChecked(false);
-        on_monitorButton_clicked(false);
-        on_monitorTxButton_toggled(false);
-
-        foreach (auto obj, this->children()) {
-            if (obj->isWidgetType()) {
-                auto wid = qobject_cast<QWidget *>(obj);
-                wid->setEnabled(false);
-            }
-        }
-
+        JS8MessageBox::warning_message(
+            this,
+            tr("Message Not Sent"),
+            tr("It appears you may have a stuck key (more than 5 repeated characters). "
+               "Please check your keyboard and try again."));
         return false;
     }
 


Joe-K0OG (commented):
@Chris-AC9KH This seems to work acceptably to me.  However, it will only catch the following form:

xxxxxx

With no following characters.  For instance, these strings will not trigger the pop-up:

xxxxxx yyy
NOW IS THE TTTTTTTTTTIME FOR ALL GOOD MEN

In my opinion, what you have implemented here is sufficient, but if you think it should catch those other strings, then maybe it needs a little more work.  I'm good either way.

73,
-Joe-
K0OG

Chris-AC9KH:
I did not change the logic of how it works. All I did was prevent it from shutting down the UI and replace that with the warning. The logic only looks at the first characters in the string so that's not changed.

Chris-AC9KH:
@Joe-K0OG basically, this is just intended to stop what you thought was a "hang" if the ensureKeyNotStuck() function ever fires. The previous implementation shut down the UI but left the clock and stuff running and might have looked like a "hang" but it was not, and it gave no recovery path without shutting down and restarting the program.

This is all that is intended to fix, not to change the logic of what fires that function(). There may be legitimate reasons why the logic was designed as it is in the first place, such as maybe transmitting something with six zeros in it: `000000` in the message body. It appears it was designed to fire that function() only if somebody tries to spam the airwaves with a `XXXXXXXXXXXXXXXX` string or something, and nothing else.

So I hesitate the change the logic unless you can come up with a really, really good reason as to why it's an issue. I've never heard of it being fired before and anybody complaining about the UI "hanging". But if I do change the logic to fire it anywhere in a MSG string, and we have these various groups that might be using some secret "code" that has repeated 6 or more character words, then we'd be introducing an actual bug that will elicit user complaints.

Joe-K0OG:
@Chris-AC9KH Like I said, I'm fine with it the way it is, so let's push it.

Thanks Chris.
-Joe-
K0OG

PR #283, opened 2026-05-03
Auto HB ACK limiting

Chris-AC9KH:
This code is not experimental. I've been running it for at least 9 months now. I cherry-pick it into whatever build I'm running at present. It's been hardwired in my personal code, but I just added a UI setting to enable/disable it, that is disabled by default. I cleared my blacklist to test the new setting and it does indeed work and now my station has been responding the frequent HB'ers and not blacklisting them.

Screenshot 2026-05-03 at 15 10 06

It is something I'd like to see in JSCall 3.0 so decided to PR it.

What it does:
- It creates a new database, hb_blocking.db3
- When a callsign sends a HB that causes your station to ACK it, it records that callsign with the timestamp in the database
- The next time that same station sends out a HB it queries the database table and compares the timestamp
- If the timestamp is < 55 minutes it adds that station to the blacklist in the UI settings where it says `Never acknowledge heartbeats from these callsigns (comma separated)` then purges that callsign from the database
- It updates the last heard HB timestamp for any callsign that is stored in the database if that station is heard again
- It is VERY effective at shutting the habitual offenders that think it's necessary to send HB's at less than 1 hour intervals
- You can remove an offender any time from the blacklist in your settings and their status will be updated to being "allowed" again

I figured it is a new feature that users can enable if they want. So it won't affect anybody and how they use HB's unless they want to use it. The reason I set it at 55 minutes is because even if you set your timer to one hour, sometimes it will send again in 57-58 minutes. So I set it at 55 minutes so it doesn't blacklist stations that actually attempt to only send once per hour with their timers.

Been going thru my list of stuff that I wanted to accomplish for 3.0. Got the UI done, squashed a few other bugs, this was a last-minute one that I decided to PR because I think some people will use it and like it.

### DIFF
--- a/JS8_Mainwindow/processHeartbeatRateLimit.cpp
+++ b/JS8_Mainwindow/processHeartbeatRateLimit.cpp
@@ -1,4 +1,4 @@
-/** \file
+/** @file
  * @brief member function of the UI_Constructor class
  *  process HB ACK rate limiting
  */
@@ -6,6 +6,11 @@
 #include "JS8_UI/mainwindow.h"
 
 void UI_Constructor::processHeartbeatRateLimit(const QString &callsign) {
+    // Feature disabled in settings
+    if (!m_config.hb_rate_limit()) {
+        return;
+    }
+
     constexpr int BLOCK_WINDOW_SECS = 55 * 60;
 
     HBBlockingDB db(hbBlockingPath());
--- a/JS8_Mainwindow/processHeartbeatRateLimit.cpp
+++ b/JS8_Mainwindow/processHeartbeatRateLimit.cpp
@@ -1,0 +1,33 @@
+/** \file
+ * @brief member function of the UI_Constructor class
+ *  process HB ACK rate limiting
+ */
+
+#include "JS8_UI/mainwindow.h"
+
+void UI_Constructor::processHeartbeatRateLimit(const QString &callsign) {
+    constexpr int BLOCK_WINDOW_SECS = 55 * 60;
+
+    HBBlockingDB db(hbBlockingPath());
+    if (!db.open()) {
+        qCDebug(mainwindow_js8)
+            << "HBBlockingDB failed to open:" << db.error();
+        return;
+    }
+
+    auto now  = DriftingDateTime::currentDateTimeUtc();
+    auto last = db.getTimestamp(callsign);
+
+    if (last.isValid() && last.secsTo(now) < BLOCK_WINDOW_SECS) {
+        qCDebug(mainwindow_js8)
+            << "HB rate-limit triggered for" << callsign
+            << "- only" << last.secsTo(now) << "s since last HB";
+
+        m_config.addToHbBlacklist(callsign);
+        db.deleteTimestamp(callsign);
+        return;
+    }
+
+    // First contact or window expired — record/refresh the timestamp
+    db.upsertTimestamp(callsign, now);
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,7 @@ target_sources(
   JS8_Main/FrequencyLineEdit.cpp
   JS8_Main/FrequencyList.cpp
   JS8_Main/Geodesic.cpp
+  JS8_Main/HBBlockingDB.cpp
   JS8_Main/HelpTextWindow.cpp
   JS8_Main/IARURegions.cpp
   JS8_Main/Inbox.cpp
@@ -312,6 +313,7 @@ target_sources(
   JS8_Mainwindow/processBufferedActivity.cpp
   JS8_Mainwindow/processCommandActivity.cpp
   JS8_Mainwindow/processDecodeEvent.cpp
+  JS8_Mainwindow/processHeartbeatRateLimit.cpp
   JS8_Mainwindow/processRxActivity.cpp
   JS8_Mainwindow/UI_Constructor.cpp
   JS8_Mode/DecodedText.cpp
--- a/JS8_Main/HBBlockingDB.cpp
+++ b/JS8_Main/HBBlockingDB.cpp
@@ -31,6 +31,7 @@ bool HBBlockingDB::open() {
 
     rc = sqlite3_exec(db_, SCHEMA, nullptr, nullptr, nullptr);
     if (rc != SQLITE_OK) {
+        close();
         return false;
     }
 
--- a/JS8_Main/HBBlockingDB.cpp
+++ b/JS8_Main/HBBlockingDB.cpp
@@ -1,0 +1,118 @@
+/**
+ * @file HBBlockingDB.cpp
+ * @brief HB ACK rate-limiting database implementation.
+ */
+
+#include "HBBlockingDB.h"
+#include "vendor/sqlite3/sqlite3.h"
+#include "DriftingDateTime.h"
+
+#include <QTimeZone>
+
+namespace {
+constexpr char SCHEMA[] =
+    "CREATE TABLE IF NOT EXISTS hb_seen_v1 ("
+    "  callsign VARCHAR(255) PRIMARY KEY, "
+    "  last_seen TEXT"
+    ");";
+} // namespace
+
+HBBlockingDB::HBBlockingDB(const QString &path)
+    : path_{path}, db_{nullptr} {}
+
+HBBlockingDB::~HBBlockingDB() { close(); }
+
+bool HBBlockingDB::open() {
+    int rc = sqlite3_open(path_.toLocal8Bit().data(), &db_);
+    if (rc != SQLITE_OK) {
+        close();
+        return false;
+    }
+
+    rc = sqlite3_exec(db_, SCHEMA, nullptr, nullptr, nullptr);
+    if (rc != SQLITE_OK) {
+        return false;
+    }
+
+    return true;
+}
+
+void HBBlockingDB::close() {
+    if (db_) {
+        sqlite3_close(db_);
+        db_ = nullptr;
+    }
+}
+
+bool HBBlockingDB::isOpen() const { return db_ != nullptr; }
+
+QString HBBlockingDB::error() const {
+    if (db_) {
+        return QString::fromLocal8Bit(sqlite3_errmsg(db_));
+    }
+    return {};
+}
+
+bool HBBlockingDB::upsertTimestamp(const QString &callsign,
+                                   const QDateTime &ts) {
+    if (!isOpen()) return false;
+
+    const char *sql =
+        "INSERT INTO hb_seen_v1 (callsign, last_seen) VALUES (?, ?) "
+        "ON CONFLICT(callsign) DO UPDATE SET last_seen = excluded.last_seen;";
+
+    sqlite3_stmt *stmt;
+    if (sqlite3_prepare_v2(db_, sql, -1, &stmt, nullptr) != SQLITE_OK)
+        return false;
+
+    auto c8 = callsign.toLocal8Bit();
+    auto t8 = ts.toUTC().toString("yyyy-MM-dd HH:mm:ss").toLocal8Bit();
+    sqlite3_bind_text(stmt, 1, c8.data(), -1, nullptr);
+    sqlite3_bind_text(stmt, 2, t8.data(), -1, nullptr);
+    sqlite3_step(stmt);
+
+    return sqlite3_finalize(stmt) == SQLITE_OK;
+}
+
+QDateTime HBBlockingDB::getTimestamp(const QString &callsign) {
+    if (!isOpen()) return {};
+
+    const char *sql =
+        "SELECT last_seen FROM hb_seen_v1 WHERE callsign = ? LIMIT 1;";
+
+    sqlite3_stmt *stmt;
+    if (sqlite3_prepare_v2(db_, sql, -1, &stmt, nullptr) != SQLITE_OK)
+        return {};
+
+    auto c8 = callsign.toLocal8Bit();
+    sqlite3_bind_text(stmt, 1, c8.data(), -1, nullptr);
+
+    QDateTime result;
+    if (sqlite3_step(stmt) == SQLITE_ROW) {
+        auto raw = QByteArray(
+            (const char *)sqlite3_column_text(stmt, 0),
+            sqlite3_column_bytes(stmt, 0));
+        result = QDateTime::fromString(
+            QString::fromUtf8(raw), "yyyy-MM-dd HH:mm:ss");
+        result.setTimeZone(QTimeZone::utc());
+    }
+
+    sqlite3_finalize(stmt);
+    return result;
+}
+
+bool HBBlockingDB::deleteTimestamp(const QString &callsign) {
+    if (!isOpen()) return false;
+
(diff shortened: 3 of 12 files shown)


Chris-AC9KH:
This is a workaround to this discussion https://github.com/orgs/JS8Call-improved/discussions/163

JS8_Mainwindow/processHeartbeatRateLimit.cpp
```diff
@@ -0,0 +1,38 @@
+/** \file
```
wmiler: @file processHeartbeatRateLimit.cpp

wmiler (approved):
/remind me API hook-in on Wednesday
Edit: And apparently that slash command doesn't work 🤦‍♂️

Chris-AC9KH:
I fixed the doxy block. I just pulled that member function out of mainwindow.cpp this morning and stuck it in JS8_Mainwindow since I'm always preaching about "don't add anything to mainwindow.cpp" :-)

punk-kaos:
I'd like to see the time limit configurable. What if I wanted say a 30 minute limit instead?

Chris-AC9KH:
@punk-kaos the reason it's not adjustable is by design. Shortening the period doesn't do anything to achieve the intended purpose.

Consensus on operating procedure by experienced operators is that when you first come on the band you send out one HB and that's the last one you need to send. As new people come on the band they send their HB and your mesh network is built. But then timers on the HB system started getting abused with stations sending HB's on timers with nobody home. And that's all they use JS8Call for. The biggest offenders run their timers and they themselves don't have ACK enabled. All they're after is SNR reports to see how many "contacts" they got and the band is totally polluted with ACK's - some of them scattered around on the passband because they turn off limiting them to the lower 500Hz.

This debate on the HB networking system has been going on for ages. A lot of people hate it because it's a nuisance if they want to do MSG'ing. So they turn off ACK's to keep their station from firing up and transmitting useless band noise, which then defeats the purpose of the HB networking system.

So there's been various solutions fielded on how to deal with the abuse. Remove the timer. Force HB's to only the lower 500Hz and remove that setting. Make the timer so it can't be set less than 60 minutes. Remove the HB system altogether because it's not actually needed anyway. Jordan said he be would onboard with removing the ACK replies. Trying to get a handle on the chaos it has caused where on some bands like 40m all it is a giant HB system with automated timers going and the vast majority of the stations not even attended.

So this feature is designed to assign a reasonable compromise to what's allowable for ACK's when HB timers are being used. 60 minutes is considered acceptable, less than that doesn't accomplish anything. It is designed to allow people to participate in the HB networking system but eliminate replies to the habitual offenders who abuse the system. With the setting I added to turn it on or off it is optional so you don't have to blacklist anybody if you don't want. Over about 9 months of running it in my personal code I arrived at 55 minutes being a reasonable number because for whatever reason the people running timers sometimes end up on a 57 or 58 minute interval instead of 60 and we don't want to blacklist them. They are "following the rules" on what's considered acceptable if timers are going to be used.

So the design is also a deterrent. When people realize they may get blacklisted if they send HB's on less than 60 minute intervals maybe they'll correct their behavior and not do it anymore. Making it adjustable totally defeats the purpose. My original code never had the setting to turn it off. That's the way I would like to see it done. But that would send the "professional HB'ers" into a coma so I made it opt-in so they can't complain if they get blacklisted for abusing the system. I just added the on/off setting yesterday when I refactored it from the old code structure prior to 2.5.1.

wmiler:
If anybody wants to write a better blurb on this issue for the docs, I'll take a separate PR for it. Still have quite a bit of documentation to do before 3.0.

Chris-AC9KH:
It was discussed in the HB hoopla that documentation should be added on recommended operating procedure regarding HB's, querying `@ALLCALL` for stored MSG's etc. where your first thing you do on the band is send your HB, then immediately query for stored MSG's in case somebody has HB ACK's turned off.

But as far as I know it never got done. People are all over the board on the HB thing; some think it's a propagation reporter like WSPR and if they don't show up in PSK Reporter they crank the watts. Some hate it and wish it was gone from the program. The HB thing has turned into a nightmare over time because of timers where people set it and walk away. But taking away the timers don't work because people will stay on the old software that has timers. We currently suppress HB ACK's if there's MSG activity going on to prevent breaking that. But for 3.0 I'd really like to see a mechanism that gives users the OPTION to break the automated timer loops at less than 60 min intervals which has been arrived at as being "reasonable".

At the end of the day many ham operators are no different from CB, just different equipment and frequencies. So some restrictions need to be put in the software in an attempt to get a handle on it. This PR has proven to be VERY effective at breaking those automated timer loops without having to do it manually. If you have a habitual offender your station will only ACK them once. The second time at less than hour they get no ACK and are permanently locked out unless you remove them from your do not reply list.

If you're one of the operators that think frequent automated HB's and ACK's are the greatest thing since sliced bread and are philosophically opposed to blacklisting anybody, then don't check the box. But if you do want to use it, it only takes a couple days and it builds a list of habitual offenders.

Joe-K0OG:
> I'd like to see the time limit configurable. What if I wanted say a 30 minute limit instead?

I agree with @Chris-AC9KH -- this is most useful if it's 55 minutes for people who really want to use it.

73,
-Joe-
K0OG

Chris-AC9KH:
> I agree with @Chris-AC9KH -- this is most useful if it's 55 minutes for people who really want to use it.

Yes @Joe-K0OG you've been involved with those HB discussions and kind of understand what the problem is. People are just turning off HB ACK's because they're annoyed with the whole thing. Which defeats the purpose of it. The hope is that this will give them a mechanism to still participate in the HB network, but eliminate the problem stations that are causing their transmitter to constantly light up. I personally didn't use HB networking for ages because of this problem. Then I dragged this old code out and cherry-picked it in and it is very effective at solving that problem. I never thought it would fly in the production code but I thought, hey, make it opt-in and then it doesn't force it on anybody.

Joe-K0OG:
> I never thought it would fly in the production code but I thought, hey, make it opt-in and then it doesn't force it on anybody.

Yes, this is very hard to argue with, even for those who might be opposed to the idea.  Good solution @Chris-AC9KH !

One minor thing I noticed in my testing (works great on both Linux and Win) is that I must have HB ACK enabled for it to start collecting stations for the database.  Is that intentional, or can it be set up so that even with HB ACK disabled it will still collect stations on the blacklist?  I can see that a station might not want to have HB ACK enabled, yet still collect those blacklist stations.

73,
-Joe-
K0OG

punk-kaos:
> It was discussed in the HB hoopla that documentation should be added on recommended operating procedure regarding HB's, querying `@ALLCALL` for stored MSG's etc. where your first thing you do on the band is send your HB, then immediately query for stored MSG's in case somebody has HB ACK's turned off.
> 
> But as far as I know it never got done. People are all over the board on the HB thing; some think it's a propagation reporter like WSPR and if they don't show up in PSK Reporter they crank the watts. Some hate it and wish it was gone from the program. The HB thing has turned into a nightmare over time because of timers where people set it and walk away. But taking away the timers don't work because people will stay on the old software that has timers. We currently suppress HB ACK's if there's MSG activity going on to prevent breaking that. But for 3.0 I'd really like to see a mechanism that gives users the OPTION to break the automated timer loops at less than 60 min intervals which has been arrived at as being "reasonable".
> 
> At the end of the day many ham operators are no different from CB, just different equipment and frequencies. So some restrictions need to be put in the software in an attempt to get a handle on it. This PR has proven to be VERY effective at breaking those automated timer loops without having to do it manually. If you have a habitual offender your station will only ACK them once. The second time at less than hour they get no ACK and are permanently locked out unless you remove them from your do not reply list.
> 
> If you're one of the operators that think frequent automated HB's and ACK's are the greatest thing since sliced bread and are philosophically opposed to blacklisting anybody, then don't check the box. But if you do want to use it, it only takes a couple days and it builds a list of habitual offenders.

I can follow that logic, seems like a good compromise.

punk-kaos:
Looks like in this change we're opening and closing the SQLite file handle every HB. That seems heavy, as we could potentially be opening and closing the file handle very rapidly. Does it really make sense to do it that way vs a persistent handle?

Chris-AC9KH:
@Joe-K0OG that's by design, it only fires if HB ACK's are on and your station processes the ACK response. That function checks the blacklist at the same time to see if it should fire an ACK and I don't think it's a good idea to accumulate a blacklist without actually participating in it. The reason is that it does allow one "courtesy" ACK the first time an offending station appears. But if they don't "obey the rules" it's the last one they'll get. If you blacklist them beforehand, they never get that one "courtesy" ACK. So then what happens is they get that first ACK, but they don't get anymore. Pretty soon they send you an `SNR?` to see if you're still there because they see you ACK'ing other people. Then you get a MSG that says, "why isn't your station sending my signal reports?" You send one back, "you got blacklisted for sending too many HB's too fast." That usually works.

@punk-kaos I figured since it's such an infrequent connection to the database and SQLite is a simple library that uses a file handle rather than a network socket for it's connection, it doesn't really make any difference. And it's probably not worth it to to leave the database connection open and increased risk of corrupting the database file if you have a power outage or some other outside factor that crashes the program. Even with MySQL it's recommended good coding practice to close the connection to the database unless you are doing constant queries and writes to the tables that requires high thruput. This is not that type of an application.

Joe-K0OG:
@Chris-AC9KH All right then, it seems to be good to push as far as I can see.  This is a very good idea Chris!

73,
-Joe-
K0OG

Chris-AC9KH:
@punk-kaos I'm more familiar with MySQL than I am with SQLite3. But from what I understand SQLite's default journal guarantees atomic commits. So a power loss mid-write leaves the database in the last committed state, not a corrupt one. But if you're connected all the time with a persistent handle it might not close it cleanly on app exit, leaving the file locked and then the system doesn't work. I'm not totally sure on that, but from what I gathered from the SQLite folks, unless it's a high-throughput application with many queries per second don't leave the file handle open.

punk-kaos:
Sounds reasonable I think. Looks good from my end.

punk-kaos (approved):

Joe-K0OG (approved):
Excellent feature for those who wish to use it.  Tested and verified to work well on Linux and Windows.

73,
-Joe-
K0OG

Chris-AC9KH:
@punk-kaos I vetted my code before merging because this was written awhile back. I added a close(); statement in the open() function for the failure branch.  If sqlite3_exec fails, the database handle was successfully opened, so closing it before returning avoids leaking the open connection for the lifetime of the object. Somehow I had skipped over that one.

withoutgettingwet:
Hi, I’m new to JS8Call. It sounds like this has been a problem, so I understand this change. 

However, I am not sure how someone who is new to JS8 is supposed to know that HBs more often that every 60 minutes is a faux pas:

- It is not in the user manual
- The default suggested values (in 2.5.2, at least) are 10/15/30/60 minutes. Assuming that’s not been changed in 3.0, that makes three out of those four options noob traps.
- This change provides no feedback to offender that they are not following best practice
- The blacklist is not a “timeout”, it requires the operator to remove a station manually. Someone could make a mistake once and be blacklisted until they reach out to the operator of the station that blacklisted them. And they can’t know for certain if they’ve been blacklisted.

And as a new JS8 user, the only reason I’m aware of this is because I was checking recent pull requests in this repository.

I’m not saying that it isn’t a necessary change, just that it needs to be communicated somehow. Either in the user manual, in the program as some kind of pop up warning, or something. 

IMO I think a good addition to this would be having JS8call log the time of the last HB tx, and warn the user if they attempt to send a HB <60 minutes later (manually or otherwise)

I would also add that since “custom” is an option I think it would make more sense to change the default values to be higher, e.g. 60/120/180/360. That way, even someone who is brand new will have an implicit understanding that if they go under 60 they are going off the beaten path. 

I have not yet had the opportunity to test the 3.0 version yet, so if some of these changes have been made already, great!

Thank you all for working on this awesome program. 73!

wmiler:
@withoutgettingwet See issue #285

Chris-AC9KH:
The HB on timer thing is all over the board on opinions. While I don't think timers should've ever been put on HB's, others think it's the greatest thing since sliced bread. I wrote the code for the auto-blacklisting like late last summer sometime and I've been running it on and off in various development builds for a long time. So I finally decided to make it optional (not enabled by default), and PR it in master and the next release.

It doesn't have to change how people use the HB networking system. It's there for the people who get annoyed by the continuous ACK'ing because of the frequent HB'ers. The experience range of the frequent HB'ers varies from newbie to been using JS8Call for years. There is no "rules" that say how you have to use JS8Call and even some experienced users think it's necessary to get those all-important SNR reports more often than once an hour.

So as a newbie to JS8Call you see the checkbox in the HB Networking section. If, after using JS8Call for awhile, you get annoyed by the people who send more often than once an hour, you can check the box. They'll get one courtesy "ACK" after that, but the second time will blacklist them and limit your ACK's to those who only send at greater than 55 minute intervals. After a day or so it gathers a list of the frequent HB'ers in the box above it. You can clear that list at any time and give them one more ACK. But over time you'll see it's the same stations that always appear in your blacklist.

When people learn that this setting is in the new software, maybe it will be a deterrent to people who abuse the HB system, realizing they might get blacklisted if they don't stop sending HB's out so often. Putting any sort of explanation for it in documentation or popups is about as useless as trying to explain the box right above it where it says "Never ACK these stations". It's a pretty much self-explanatory totally optional feature.

jsherer:
I think it is a good add for folks who are bothered by HBs.

One option that may be worthwhile to explore in the future is a "soft"
blacklist (graylist?), which throttles responses to offenders to once per
hour or longer, rather than immediately blacklisting them outright. Think
token-bucket throttling, but for responses.

The key here is that we should 1) point folks to notes in the documentation
about it and 2) be clear that it's disabled by default and a trade-off an
operator is making in how _their_ station performs.

Chris-AC9KH:
> One option that may be worthwhile to explore in the future is a "soft" blacklist (graylist?), which throttles responses to offenders to once per hour or longer

Yep, that might be a good idea too. When I originally wrote the code it was hard-wired with no on/off switch. I added the switch when I PR'd it, purposely setting it to disabled by default, figuring we'll see what kind of feedback we get on it.

I think there's people that WANT to participate in the HB system, but turn it off because of the frequent HB'ers. This was designed with those people in mind, without otherwise changing anything with how the HB system works. Some people would be unhappy if anything with the HB networking system was changed or restricted, others would be elated over it. It's impossible to make everybody happy, but this is an attempt. And possibly a deterrent to the frequent HB'ing once people learn that it's in the new software and they might get blocked if they don't be a little more reasonable with their HB intervals.

Joe-K0OG:
> One option that may be worthwhile to explore in the future is a "soft" blacklist (graylist?), which throttles responses to offenders to once per hour or longer, rather than immediately blacklisting them outright. Think token-bucket throttling, but for responses.

I really like this idea.  You could have two checkboxes - select either (but not both):
 x Blacklist if station send more than one HB per hour
 x Throttle my ACK to once/hour to station if he sends more than one HB per hour

In either case, the station's callsign will be placed in the list.  The operator can select either option (or none) to be active on the list of stations already in the list.

73,
-Joe-
K0OG

PR #284, opened 2026-05-04
Update windows workflow

wmiler:


wmiler:
Hmm, has a bug in it when run here... marking WIP till I figure out why it does that.

wmiler:
Ok, I'll admit that some times I'm abit slow to remember things....

PR #289, opened 2026-05-09
Add callsign blocking feature, settable in the UI

Chris-AC9KH:
This is a feature that was requested to be added for the next release to stop certain callsigns from spamming people's MSG pane.

### DIFF
--- a/JS8_UI/Configuration.cpp
+++ b/JS8_UI/Configuration.cpp
@@ -641,6 +641,7 @@ class Configuration::impl final : public QDialog {
     QStringList auto_blacklist_;
     QStringList hb_blacklist_;
     QStringList spot_blacklist_;
+    QStringList rx_callsign_blocklist_;
     QStringList primary_highlight_words_;
     QStringList secondary_highlight_words_;
     QString eot_;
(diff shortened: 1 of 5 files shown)


Chris-AC9KH:
Description of the PR more in detail, and why it exists is in the Discussions: https://github.com/orgs/JS8Call-improved/discussions/174

Apparently the Subspace guy is spamming people on 40m so they requested this.

wmiler:
I can't test this until later Saturday when I get off the mountain, but looks ok. And, yes, the idiot is spamming.

I really should POTA from up here...

Chris-AC9KH:
I added isNewLine to the conditional just for the event that the EOT character is never received from the spammer. The whole thing is pretty simple - I just copied what spot_blacklist is doing, and created a simple conditional logic block in mainwindow to implement it. So the total overhead is zero. Blocking at the decoder level would be quite extensive.

Joe-K0OG (approved):
This is EXCELLENT - Thanks for doing this Chris.  Some ideas on how to use "Hide messages from...", in combination with "Never autoreply from these...":

Image

* With both fields empty, operation as normal
* With "Hide messages from..." in use, messages will not appear in the message pane, but will still ACK with a ">" or "MSG"
* With both fields in use, JS8Call will not display in the message pane, AND will not ACK with a ">" or "MSG" and will NOT store a MSG (thus completely blocking that callsign)

I built & tested this on both Linux Mint 22.3 and Win11.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG it did seem like a sensible request for the feature and very easy to implement as long as we're NOT trying to do it at the decode level. So we let them in the door, so to speak, so everybody knows they're there. But we don't let them say anything :-)

I does make sense to me because it is theoretically possible for somebody to use JS8Call to spam people with ALLCALL. The only way we provided to prevent that is to remove yourself from the ALLCALL group. So IMO a quite sensible reason to have it, and worth implementing in the code.

Chris-AC9KH:
I don't think there's any objections to merging this. It does not really change how anything works except for a simple logic block that keeps a particular callsign out of the message pane. So I'm going to drop it in, then cherry-pick it to the 3.0 release candidate. We'd have a bunch of disappointed users if this feature didn't make it into the next release.

PR #296, opened 2026-05-17
fix CQ notification in call activity pane

Chris-AC9KH:
This fixes an issue that I thought worked, but was brought up on the forum here:
https://js8call.groups.io/g/main/topic/119349038?p=Created,,,20,1,0,0

When a CQ comes in it's supposed to display the little telephone and make the table row turn green. It does not do that, but it does now. This was broken in 2.2.0 and never caught until now.

Screenshot 2026-05-16 at 21 07 31

### DIFF
--- a/JS8_Mainwindow/processCommandActivity.cpp
+++ b/JS8_Mainwindow/processCommandActivity.cpp
@@ -737,7 +737,22 @@ void UI_Constructor::processCommandActivity() {
 
         // PROCESS CQ
         else if (d.cmd == " CQ") {
-            qCDebug(mainwindow_js8) << "skipping incoming cq" << d.text;
+            CallDetail cd = {};
+            cd.bits = d.bits;
+            cd.call = d.from;
+            cd.dial = d.dial;
+            cd.offset = d.offset;
+            cd.grid = d.grid;
+            cd.snr = d.snr;
+            cd.tdrift = d.tdrift;
+            cd.submode = d.submode;
+            cd.utcTimestamp = d.utcTimestamp;
+            cd.cqTimestamp = d.utcTimestamp.isValid()
+                                 ? d.utcTimestamp
+                                 : DriftingDateTime::currentDateTimeUtc();
+
+            logCallActivity(cd, true);
+            logHeardGraph(d.from, d.to);
             continue;
         }
 


Chris-AC9KH:
What's truly amazing with this one is that all the code was already there to make it happen. But it was never hooked up. Due to compiler differences and how Qt works on Mac vs Windows/Linux, it would work on Mac as long as you are using the AppleSystemUIFont that contain the glyph for the unicode character. But if you used any of the other fonts you can select in the UI configuration tab it is broken.

It is absolutely amazing nobody ever caught this until now until Joe Burnham mentioned it and I'm like, what? We already got that feature. I'm reading thru the code, and yep, it's all there. Then I went to where it originates from in the processCommandActivity() member function and holy crap - it was skipped over and never hooked up to the rest of the code! I went clear back to 2.2.0 and looked and it must've been an oversight then.

Jordan put a debug line in to see if it's firing, then a continue; statement obviously assuming the rest of the code would execute. Back at that time both functions() were in the same mainwindow.cpp file. And we had mainwindow, and we had a MainWindow class, too many things named "Main Window". So it's easy to understand how it could be skipped over if you're tearing your hair out trying to keep it straight in your head how that monster executed. But since they're broken down now, it was pretty easy to see that it was never hooked up to the displayCallActivity() member function of the UI_Constructor class.

rruchte (approved):
LGTM

wmiler (approved):

Chris-AC9KH:
I only tested this on Windows and Mac. I'll wait for @Joe-K0OG to test it on linux to make sure it works there.

It was actually kinda funny - Joe Burnham said it states right in the User Guide that it's supposed display that little telephone when a CQ comes in. But in all the years he's been using it since 2.2.0 he has never seen it work.

Joe-K0OG (approved):
>I only tested this on Windows and Mac. I'll wait for @Joe-K0OG to test it on linux to make sure it works there.

@Chris-AC9KH What is funny about this is that I HAVE on occasion seen the telephone icon - over the years, but never paid much attention to it. I can confirm that it works just as it should on both Linux Mint 22.3 and Win11, and works well in combination with the callsign filtering (blocking) feature. The filtering/blocking "x" supersedes the icon, but the line highlights indicating a CQ. It looks good to me!

73,
-Joe-
K0OG

PR #309, opened 2026-05-27
Rx pane scroll bar fix

### LINKED ISSUE #302: js8call.ini annoyance
On Sat, May 16, 2026 at 03:36 PM, Richard KO4JNH wrote:

    When I click File -> Settings -> OK, the center yellow area (directed message area) jumps "back in time" to some earlier point, and I have to slide the scroll bar back down to the bottom to get back to the "present."

This has been going on forever. When you click OK in the settings dialog, openSettings() calls displayActivity(true) at the end of the function, which triggers a repopulation of the RX pane from the saved state. The saved state in the JS8Call.ini was written by the last writeSettings() call, which may not be current.

Chris-AC9KH:
Force Rx pane scroll bar to the bottom on program start and save the scroll bar position on settings reload.

This fixes an issue noted on the forum where when the Settings dialog is opened and click OK, the scrollbar in the Rx pane will jump back to a certain position in time. I thought this was coming from reloading the settings in JS8Call.ini but I was wrong. I never did find the actual call site that is triggering this. And it does not happen all the time but I also did not figure out the exact set of parameters that sets it off. It appears to happen when there is a variety of ACK's, maybe a directed or two, possibly mixed in with some MSG's you have gotten? Who knows? I only know that it does happen and it started doing that this afternoon. So this is a good opportunity to see if I can fix it.

Since this is not a bug, it is perfectly normal from Qt's standpoint and maybe they even consider it a "feature", and I could not identify the exact mechanism that causes it, I took the easy way out:
- In UI_Constructor::openSettings() save the scroll bar position before we open any settings
- Let that function run its mysterious witchcraft
- Reload the scroll bar position before the function exits

It works, even though it was not quite the type of "fix" I had envisioned.

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/302

### DIFF
--- a/JS8_UI/mainwindow.cpp
+++ b/JS8_UI/mainwindow.cpp
@@ -506,6 +506,10 @@ void UI_Constructor::readSettings() {
         m_config.reset_activity()
             ? ""
             : m_settings->value("RXActivity", "").toString());
+    QTimer::singleShot(0, this, [this](){
+        ui->textEditRX->verticalScrollBar()->setValue(
+            ui->textEditRX->verticalScrollBar()->maximum());
+    });
     ui->actionShow_Band_Heartbeats_and_ACKs->setChecked(
         m_settings->value("BandHBActivityVisible", true).toBool());
     m_settings->endGroup();
@@ -1012,6 +1016,11 @@ void UI_Constructor::on_actionSettings_triggered() { openSettings(); }
 
 void UI_Constructor::openSettings(int tab) {
     m_config.select_tab(tab);
+    
+    // Save scroll position before opening settings
+    QScrollBar *bar = ui->textEditRX->verticalScrollBar();
+    const bool wasAtBottom = (bar->value() == bar->maximum());
+    const int savedPos = bar->value();
 
     // things that might change that we need know about
     auto callsign = m_config.my_callsign();
@@ -1072,6 +1081,16 @@ void UI_Constructor::openSettings(int tab) {
 
         m_opCall = m_config.opCall();
     }
+    
+    // Restore scroll position after settings dialog
+    QTimer::singleShot(0, this, [this, wasAtBottom, savedPos](){
+        QScrollBar *bar = ui->textEditRX->verticalScrollBar();
+        if (wasAtBottom) {
+            bar->setValue(bar->maximum());
+        } else {
+            bar->setValue(savedPos);
+        }
+    });
 }
 
 void UI_Constructor::prepareApi() {


punk-kaos (approved):
Looks clean and simple to me.

rruchte (approved):
Looks good to me.

Chris-AC9KH:
While this works, I wish I could figure out the mechanism that causes the behavior. Accepting settings triggers a lot of UI work: displayActivity(true), setupJS8(), enable_DXCC_entity, etc..... The history in textEditRX is loaded from a HTML string in JS8Call.ini so it actually repopulates that pane when the settings are accepted with OK. But why it jumps back to a certain place in time when it is reloaded by accepting settings remains a mystery. In theory, when the program is started it should do the same thing, but doesn't. It sets the scroll bar to the top at the very beginning of the log instead of at that point in time that accepting settings moves it to.

I could find nowhere in the code in either path that explicitly tell that text pane to scroll anywhere except in some conditions like writeNoticeTextToUI() and those don't run on a settings save. So it has to be some quirk in Qt doing it, which makes it not a bug in our code. And like @Joe-K0OG noted where he can't duplicate it, if I clear that incoming pane, let it load up with "normal" stuff it won't do it. But if I let it collect several days worth of stuff, then it starts happening. Like right now, in v3.0.2, starting with scroll bar at the bottom, open Preference, click OK and it jumps back to this exact point in time, every time, which was about 3 or 4 days ago.

Screenshot 2026-05-27 at 07 50 55

So obviously, to jump back to that exact point every time, means something here is being cached SOMEWHERE. But where? I'm not using ALL or DIRECTED.txt so it's not coming from there. The js8call_wisdom.dat appears to store some fftw stuff and I don't think it's coming from there. It could be coming from JS8Call.ini but I can't identify what key it would come from. So as far as I can tell, what causes it is good for an episode of Unsolved Mysteries on TV.

Joe-K0OG (approved):
@Chris-AC9KH I tested your patch on both Linux and Windows, and it does not cause any problem that I can find.  Of course, I didn't observe the scroll-back problem to begin with, and I still don't, so it seems your fix is at worst harmless, and at best fixes a real problem I just can't observe.  You mentioned that you see it after a few days' operation.  I NEVER leave my station on-air when I'm not present (at least in the house and pass by it occasionally), so don't have days-worth of RX Text to scroll through, only a few hours at most.

So, it seems good to me.

Chris-AC9KH:
I'll merge this. I been running it for a couple days with no issues at all.

PR #311, opened 2026-05-30
Waterfall cursor fix for split mode.

Chris-AC9KH:
Explanation of issue:
When I added the frequency offset slider I had basically designed it so the frequency offset slider controls the waterfall cursor. At the same time, the waterfall cursor can also control the frequency offset slider. I had hooked it up for HB's, but failed to hook it up for split mode. This causes the frequency slider to display the correct offset, but the waterfall cursor did not move when transmitting in split mode.

### DIFF
--- a/JS8_UI/mainwindow.cpp
+++ b/JS8_UI/mainwindow.cpp
@@ -1358,6 +1358,7 @@ void UI_Constructor::displayDialFrequency() {
     }
 
     freqOffsetWidget->setValue(audio_frequency);
+    m_wideGraph->setFreq(audio_frequency);
 
     auto const onAir = dial_frequency + audio_frequency;
         frequency_label.setText(QString("Freq: %1").arg(Radio::pretty_frequency_MHz_string(onAir)));


rruchte (approved):
LGTM

Joe-K0OG:
This change may lead to a little confusion.  When I look at the waterfall, I see where my cursor is in relation to other stations on the waterfall.  If running split (either rig or fake-it), when I transmit, my location on the waterfall as seen by other stations is where I placed it before transmission.  With this change, looking at the waterfall while I'm transmitting in split mode (if the cursor is outside the 1500-2000Hz range), my waterfall cursor jumps to where it is NOT really located relative to other stations on the waterfall.  This is misleading.  The cursor on the waterfall is used to display my location **in relation to other stations**, which location is produced **on-air** by the combination of dial frequency and offset as displayed in the upper-left corner of JS8Call.

I therefore wish to argue for reversion of this change so that the cursor does not jump on the waterfall while transmitting in split mode.  @Chris-AC9KH Please consider reversion.  I think it makes more sense the way it was.

73,
-Joe-
K0OG

Chris-AC9KH:
It's easily reverted. Just one line of code. What does the old software do before I added the offset slider?

Chris-AC9KH:
Ok, I see the old code did NOT move the waterfall cursor during split operation. For some reason, there was some folks on the forum that thought this is a "bug". The offset slider moved and indicated the actual offset, but the waterfall cursor didn't move so they thought something was wrong.

Joe-K0OG:
Yeah, as I remember, it has ALWAYS stayed put on the waterfall, from pre-1.0 days.  This keeps the cursor in proper relation to other signals on the waterfall, so makes more sense to revert to the old behavior.

Chris-AC9KH:
Ok, I'll just pull that line of code out and direct-push it to master. No PR needed for that

Joe-K0OG:
Thanks Chris!

Good evening & 73,
-Joe-
K0OG

Issue #297, opened 2026-05-17
Question: Windows compiler switch


wmiler:
We build with MSYS2 for our Windows executables, Currently via the MINGW64 compiler set. MSYS2 is phasing this compiler set out as below:

```
2026-03-15 - Deprecating the MINGW64 Environment

As support for Windows 8.1 has been dropped, there is no longer a need for non-UCRT environments such as 
MINGW64. Consequently, we are beginning to phase out the MINGW64 environment. To start, no new 
packages will be added to this environment, and existing leaf packages may be removed if issues arise. If you 
are currently relying on the MINGW64 environment, please consider switching to UCRT64 or CLANG64 
instead.
```

```
UCRT (Universal C Runtime) is a newer version which is also used by Microsoft Visual Studio by default. It 
should work and behave as if the code was compiled with MSVC.
```

CLANG64 is a bit of CLANG and a bit of GNU C++ mixed together into a working system.

So, which route do we want to go?

Chris-AC9KH:
I'd say it depends on if you want to use gcc or clang for the compiler. I switched to clang (llvm-mingw) long ago in Qt Creator and it works GREAT.

So there's a couple different ways I see to approach this. You can switch from MinGW64 to Clang64. Or abandon MSYS2 altogether and switch the workflow to a Qt Creator workflow.  This is supported with Qt's jurplel/install-qt-action that supports llvm-mingw directly, so it's the same as a desktop build.

You would fetch and use the same Windows library we host in js8lib that is used for a Qt Creator build. And make sure Qt is built with the clang compiler instead of gcc because the two aren't compatible. After I got used to Qt Creator I made a test workflow that I was going to try on my own repo, but never got around to it yet.

wmiler:
Huh, wasn't aware of the cli workflow for QT Creator, something to consider. I'm up for what ever is considered the most maintainable route.

Chris-AC9KH:
@wmiler just for the heck of it I copied that test workflow to a different branch and pushed it here:
https://github.com/Chris-AC9KH/JS8Call-improved/tree/pr/workflow

I never tried it, but it should get rid of all the GNU components for a Windows build. Even the C library is changed to libc++ instead of the GNU one and that's all contained in the Windows library package on js8lib. The dll's that need to be copied to the JS8Call folder are in the lib in a dll folder.

[...]

PR #322, opened 2026-06-07
Windows compiler switch

wmiler:
Issue #297 

Sorry for the really long delay.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -48,100 +48,61 @@ jobs:
     steps:
     - run: git config --global core.autocrlf input
     - uses: actions/checkout@v5
-    - uses: msys2/setup-msys2@v2
-      with:
-        update: true
-        cache: true
-        install: >-
-          git python base-devel make unzip zip tar
-        pacboy: >-
-          toolchain:p
-          autotools:p
-          cmake:p
-          ninja:p
-          libusb:p
-          wget:p
-          boost:p
-          fftw:p
-          sqlite3:p
-          python:p
 
     - name: Setup build directory
-      shell: msys2 {0}
-      run: mkdir -p build
+      run: mkdir build
 
     - name: Install Qt
       uses: jurplel/install-qt-action@v4
       with:
         version: '6.9.3'
         target: 'desktop'
-        arch: 'win64_mingw'
+        arch: 'win64_llvm_mingw'
+        tools: 'tools_llvm_mingw1706'
         modules: 'qtmultimedia qtserialport'
         cache: true
         dir: 'D:/a/JS8Call-improved/JS8Call-improved/local/qt693'
 
-# hamlib
-    - name: Cache Check Hamlib 4.7.1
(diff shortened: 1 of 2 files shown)


wmiler:
This prolly doesn't have the right names on it for the installer packages. Just double check, I've been out of pocket for the last couple of weeks.

Chris-AC9KH:
@wmiler, looking at this, it looks like you patched instead of starting over. You're failing to install the Vulkan SDK. You're failing to extract the version from CMake. You're using Windows-2022 instead of Windows-2025. You're using the completely wrong prefix path.

Basically the template for what works is here:
https://github.com/Chris-AC9KH/JS8Call-improved/blob/pr/workflow/.github/workflows/ci-windows-clang.yml

You need to start with that template, then put your signing stuff into it and get rid of all the "-devel" stuff which we don't even use.

Chris-AC9KH:
The changes to the Inno Setup script were very minor to pull the version in with CMake. That is here, and it works:
https://github.com/Chris-AC9KH/JS8Call-improved/blob/pr/workflow/.github/workflows/scripts/ci-windows-installer.iss

Chris-AC9KH:
@wmiler to be clear, there is no such thing as a "Release" or "CI Build" anymore. They're all done as Release even if it's CI. The only difference is that Release will be done on a release branch and be versioned. CI runs on master and is non-versioned, identified by git hash. And signing would be optional on CI.

This is all baked into the branch on my repo, tested and it works. The installer comes out of the oven with the same git hash as the build if it's on master, and the same version if it's on a release branch. The naming convention for the installer, etc. is baked into that example branch. The only thing I didn't do was put the signing in it.

wmiler:
Yeah, got it. Dunno why it's failing to unzip the archive, which is why I dropped back to the Win10 image. The windows-2025 image is now Win11.

Chris-AC9KH:
@wmiler well, at preset you have a hybrid of stuff you tried to patch into the old workflow script, which is not going to work. Recommend starting with the new one and patch the other way to add the signing and stuff. What's going on with the bottom of the current script with copying all those libraries and stuff I have no clue. That's handled in one step with the new library. But the prefix path is busted, that looks like it's from the old script.

So not totally sure what you got going on there, or trying to do.

Chris-AC9KH:
@wmiler this is fixed with https://github.com/JS8Call-improved/JS8Call-improved/commit/e73f0b136c1b21d3ba3a2a19afe27114a8486c20 and https://github.com/JS8Call-improved/JS8Call-improved/commit/4bf36c985aa323f0521f4204aa22b368141899f3 but I did NOT do anything with linux. That script is considerably more complex and I'm afraid I would break something there.

Chris-AC9KH:
I tested the new windows Release script (which only runs manually for creating a signed release). It worked and everything went smoothly. I can't run that script on any release branch because it only exists on master at present, so I had to run it against master.

So I think Windows and Mac are good to go on workflows. Only linux is left to do yet.

PR #323, opened 2026-06-09
Switch to Clang 17 compiler for Windows workflow builds

### LINKED ISSUE #297: Question: Windows compiler switch

Chris-AC9KH:
Add Vulkan headers on the runner for Windows build
add CMake versioning to Windows workflow
update Windows workflow path for new optimized library package
update Windows workflow with separate CI and Release scripts

@wmiler, since you are very busy and it would be nice to have the Windows workflow working I thought I could help out with this. I got rid of all the complex CI/Release logic and broke it down into two scripts. The CI one runs as before and does not mess with code signing. It is only for PR's and push actions to make sure the code builds.

The Release one is run manually when we want to create a Release build, otherwise it has no trigger.

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/297

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -1,46 +1,28 @@
-permissions:
-  contents: read
-  attestations: write
-  id-token: write
-  
 name: CI Build for MacOS 26 Arm64 (M1 and higher)
-run-name: ${{ (github.event.inputs.environment) || (github.event.pull_request.title) || 'CI' }} Build for MacOS M1
+run-name: ${{ github.event.pull_request.title || github.ref_name }} CI Build for MacOS M1
 
-on: 
-  workflow_dispatch: 
+on:
+  push:
   pull_request:
-  workflow_call:
-    inputs:
-      build-type:
-        description: 'Build type (Debug/Release)'
-        required: false
-        type: string
-        default: 'Release'
-      release-ver:
-        description: 'Release Version (v3.0.0)'
-        required: false
-        type: string      
-        default: 'v3.0.0'
 
 env:
-  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
 
 jobs:
-  build_macos-latest:
-    runs-on: macOS-26
-    permissions: write-all
+  build_macos-26:
+    runs-on: macos-26
+    permissions:
+      contents: read
+      attestations: write
+      id-token: write
 
     steps:
-    - uses: actions/checkout@v6
+    - name: Configure Git line endings
+      run: git config --global core.autocrlf input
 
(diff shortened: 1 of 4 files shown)


Chris-AC9KH:
@wmiler tested the clang-built JS8Call and Windows installer that this PR spit out and it worked beautifully, complete with being versioned with the git hash like it should.

One note on the Release script - I removed the continue on fail with code signing. A Release build should not be done without signing it, so if signing fails the script will exit instead of spitting out an unsigned build.

Chris-AC9KH:
@wmiler I changed the Mac workflow to use Xcode and use CMake versioning, and drop the runner_arch. And switched the upload artifact step to v7 for both so it doesn't zip the artifacts anymore

Chris-AC9KH:
@wmiler I'm sure you're busy and the new workflows run beautifully. MacOS is the fastest one with Xcode at 2m 39sec. Windows is still the slowest, but that's inherent with the Windows platform. But we still cut the runner time significantly with this change. Both Windows and Mac now use our native CMake versioning for the artifacts.

I could attack the linux one. But I might break something there because that one is most complex. And maybe we want to create two different scripts - one for Release and one for CI like I did for Windows. The Release one COULD call and run the new script in tools to create a .deb as well as AppImage.

Anyway, I'm going to merge this because it works. It kinda killed your other PR, but I figured you were busy so I'll help out on this one for the stuff I'm familiar with.

wmiler:
Curiosity question, why is Vulcan now a requirement?

Also, you'll need to add the `workflow_dispatch` back in, otherwise we can't run it from the Actions page. You also need the `inputs->environment` to the `workflow_dispatch`, no other parts (release-ver or build-type) are needed.

edit: for ci-windows

Chris-AC9KH:
@wmiler I fixed the workflow dispatch so they should appear on the Actions page now.

The Vulkan headers are required on Windows for certain graphics drivers. Without them you'll get a warning from the compiler that Vulkan driver support is disabled in Qt. It is actually the same on linux.

PR #328, opened 2026-06-13
Bug Fix: crash on importing invalid waterfall palette files

### LINKED ISSUE #326: waterfall colorpalette import causes crashes
So I tried to install Gnuplot's "inferno.pal" color palette via the "View->Show Waterfall Controls" menu.
Then in the "Palette" box I selected the palette "User Defined", then clicking "Adjust...", "Import..." and selected the file from https://github.com/Gnuplotting/gnuplot-palettes/blob/master/inferno.pal
Immediately after confirming the file, JS8Call crashes.

OS: Win11 23H2 JS8CALL Version: 3.0.2 via github.com released binary
Rig: FT-991 via FLrig, Audio via DATA-Port ("9k6 packet port"), CAT via RS232

Chris-AC9KH:
- install guards on import_palette() and export_palette() functions to prevent std::terminate() → abort() if the file is invalid or contains too many colors
- fix directory structure and move waterfall to JS8_UI

We had a classic uncaught exception in import_palette() that has evidently been there forever. So I put a guard on it. I found the same thing potentially exists in export_palette. It is unlikely that one would ever trigger but I put a wrapper on it anyway to be safe.

At the same time I moved the waterfall stuff to JS8_UI, which I evidently missed when I did the source code refactor a few months ago. It has a .ui file associated with it, so it belongs in JS8_UI

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/326


wmiler (approved):

rruchte (approved):
LGTM

Joe-K0OG (approved):
Tested on Linux and Windows, and works fine on both - no more crash, and a nice message indicating an incorrectly formatted .pal file.  Thanks Chris!

73,
-Joe-
K0OG

Chris-AC9KH:
It also worked fine on MacOS, built locally here. It was an easily duplicatable issue so was able to get a local stack trace, which pinpointed the problem and made it easy to find.

PR #337, opened 2026-07-05
BUG FIX: processCommandActivity.cpp

Chris-AC9KH:
Directed replies to group queries can overwrite each other in RX pane (textEditRX)

Context:
When multiple stations reply to a group-addressed query (i.e. "@GROUP QUERY MSGS") within the same second, their replies share an identical rendered timestamp string. The avoid duplicate line logic in processCommandActivity() searched backward for that timestamp string alone and erased whatever block it found, with no check that the found block actually belonged to the same sender. With enough replies landing in one processing tick, each new reply erased the previous different station's line, leaving only the last reply visible in textEditRX (the incoming MSG pane)

Thanks to John W2KI for reporting this bug and testing the fix.

Note: extensive diagnostic logging was added to diagnose this issue. After reviewing the final fix and there was over 70 lines of logging code added to several files, I cleaned that up and left only the final fix in the code.

### DIFF
--- a/JS8_Mainwindow/processBufferedActivity.cpp
+++ b/JS8_Mainwindow/processBufferedActivity.cpp
@@ -25,17 +25,18 @@ void UI_Constructor::processBufferedActivity() {
         if (!buffer.msgs.isEmpty()) {
             dt = qMax(dt, buffer.msgs.last().utcTimestamp);
         }
+        
+        int ageSecs = dt.secsTo(DriftingDateTime::currentDateTimeUtc());
 
         // if the buffer has messages older than 1 minute, and we still haven't
         // closed it, let's mark it as the last frame
-        if (dt.secsTo(DriftingDateTime::currentDateTimeUtc()) > 60 &&
-            !buffer.msgs.isEmpty()) {
+        if (ageSecs > 60 && !buffer.msgs.isEmpty()) {
             buffer.msgs.last().bits |= Varicode::JS8CallLast;
         }
 
         // but, if the buffer is older than 1.5 minutes, and we still haven't
         // closed it, just remove it and skip
-        if (dt.secsTo(DriftingDateTime::currentDateTimeUtc()) > 90) {
+        if (ageSecs > 90) {
             m_messageBuffer.remove(freq);
             continue;
         }
--- a/JS8_Mainwindow/processCommandActivity.cpp
+++ b/JS8_Mainwindow/processCommandActivity.cpp
@@ -216,29 +216,25 @@ void UI_Constructor::processCommandActivity() {
             // ACKs and SNRs are the most likely source of items to be
             // overwritten (multiple responses at once)... so don't overwrite
             // those (i.e., print each on a new line)
-            bool shouldOverwrite =
-                (!d.cmd.contains(" ACK") &&
-                 !d.cmd.contains(" SNR")); /* && isRecentOffset(d.freq);*/
+            bool shouldOverwrite = (!d.cmd.contains(" ACK") && !d.cmd.contains(" SNR"));
 
             if (shouldOverwrite &&
-                ui->textEditRX->find(d.utcTimestamp.time().toString(),
-                                     QTextDocument::FindBackward)) {
-                // ... maybe we could delete the last line that had this message
-                // on this frequency...
-                c = ui->textEditRX->textCursor();
-                c.movePosition(QTextCursor::StartOfBlock);
-                c.movePosition(QTextCursor::EndOfBlock,
-                               QTextCursor::KeepAnchor);
-                qCDebug(mainwindow_js8) << "should display directed message, "
-                                           "erasing last rx activity line..."
-                                        << c.selectedText().toUpper();
-                c.removeSelectedText();
-                c.deletePreviousChar();
-                c.deletePreviousChar();
-                /*
-                c.deleteChar();
-                c.deleteChar();
-                */
+                ui->textEditRX->find(d.utcTimestamp.time().toString(), QTextDocument::FindBackward)) {
+                auto candidate = ui->textEditRX->textCursor();
+                candidate.select(QTextCursor::BlockUnderCursor);
+                QString selectedText = candidate.selectedText();
+
+                // Only treat this as "the same message being redrawn" if it's
+                // actually from the same sender - otherwise it's a distinct
+                // reply that happens to share a rendered timestamp.
+                if (selectedText.contains(QString("%1:").arg(d.from))) {
+                    c = ui->textEditRX->textCursor();
+                    c.movePosition(QTextCursor::StartOfBlock);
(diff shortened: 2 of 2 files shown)


Chris-AC9KH:
@Joe-K0OG after much experimentation to figure out what was going on, and exhausting all theories as to what was causing it, this seemed to fix John's issue. Is this fix suitable? Or is it the general consensus that we need to re-write the entire processing chain and change the order of code execution?

Order of code execution wouldn't have fixed anything here anyway since there was only two exceptions in the overwrite logic ever allowed in the code for multiple decoded replies to commands. But if the consensus is that we need to change order of code execution, then I will pull this PR and shelve it.

Joe-K0OG:
@Chris-AC9KH Not considering how the code is currently written, here is the logic that seems reasonable to me, considering that the priority should be to properly display in the RX Text window all messages directed to my station (except, of course, for callsigns in the filter lists):

* Prioritize proper display of RX Text
* Secondarily, grab that text that is sent to the RX Text window and present it to the API

If that is not what is happening now, even if your current fix seems to work for John's issue, we may run into other problems in the future.  Again, I'm not a programmer, and rewriting the code may not be worth the effort.  You'll need to decide whether or not it's worth it.  If John's problem is fixed, that just may be good enough.

I also have occasionally (but rarely) noticed duplicate decodes of the same message (same time, same frequency offset), and since I have never been able to trigger that phenomenon at will, and it's so rare in my experience, it would be very difficult to diagnose the problem.  Perhaps rewriting the code to change the order of things might clean up that issue.

73,
-Joe-
K0OG

Joe-K0OG:
@Chris-AC9KH It's not worth the effort to rewrite/reorganize the code since this is not a major issue, only rarely experienced, and you fixed the immediate problems.  So, finish with this and leave it.  I think we're good with this fix.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG I'm considering this one dead since there was already a consensus during diagnosis that it must have something to do with order of execution (which it didn't). So unless we get consensus that this is the actual fix, I'm leaving as-is - same as the linux audio thing. I'm not dropping a fix into the code as long as there's developers convinced there must be some other issue.

The code is there if somebody decides it's the actual fix - grab it and use it. Otherwise I feel I exercised due diligence in tracking down the real problem (analysis of detailed logged data to find where it was breaking is in the associated discussion thread). This is explained clearly in the code comment I put above the fix in the conditional that I added.

Chris-AC9KH:
So what I'll do for the time being is to leave it open, but marked for discussion.

Joe-K0OG (approved):
After some testing today, this fix seems to work fine - I've not observerd any missing lines in the RX Text window with multiple-simultaneous respondents to a query, and there does not seem to be any negative impact otherwise.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG @wmiler @rruchte I'm preparing a JS8Call 3.0.3 patch release because there's been several minor bug fixes since 3.0.2. I'm also going to migrate the new workflow build system for the 3.0.3 patch release since it is more reliable and uses our own internally built libraries.

I'm going to cherry-pick this one into the patch release since it does fix the issue noted by John W2KI. But leave it open on master for right now because it is possible other developers might have ideas about how to change the code execution order for JS8Call 4.0.

There seemed to be a consensus that the API/Band Activity has "priority" over textEditRX, which I did not find during my analysis of the issue to be the case. I even tried a forced buffer hang on the API client (JS8Spotter) and it did not make any difference. But that does not mean that the code execution order cannot have issues with modern "hybrid" Intel processors that run one thread on a e-core and another on a p-core, causing issues with out-of-order execution. Those e-cores don't even have vectorization and they got less processing power than an old monolithic Core2. So if a thread lands on one of those cores, especially the GUI thread, it can cause issues. This issue is non-existent with Apple's M-series, it is purely isolated to x86_64 "hybrid" cpu's. These processors are pure scrap right off the die, so with Intel posting $19 billion in losses, and even with the government bailout they're still $8 billion short this year, it's not even known if Intel will be around long-term:
https://www.pcmag.com/news/too-many-crashes-game-developer-accuses-intel-of-selling-defective-cpus

So in the event we want to re-write JS8Call's engine to be more friendly to the x86_64 "hybrid" processors and assign threads to specific cores, this would be pointless to merge into master at the present time.

Joe-K0OG:
Concerning 3.0.3, I thought we were close on a 4.0, so should we just wait a few days/weeks and release 4.0 when it's ready?

Even if M$ went belly-up today, we would still need to support legacy hardware for a long time to come, so shouldn't give up on it.  Whether or not the code will need a rewrite, I say let's just wait until problem reports come in.  It may be too much a wasted effort if not needed.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG I don't envision a 4.0 release anytime soon at the present. Maybe sometime next winter/early next year? We successfully made the move to Qt 6.11 and clang compiler across the board, which makes the step to the next Qt LTS release (Qt 6.12) easier. And allows starting work on the audio backend in JS8Call with a standardized compiler for all three platforms. This was rather hair-raising to accomplish, but now that it's in place, and it works, I'm much happier about it.

But the time available to work on 4.0 here is just not gonna happen until we get snowed in and there's nothing else to do. Kristin (my wife) has like 70 irons in the fire that keeps me busy every hour of daylight fixing blown hydraulic hoses, cylinders and pumps on track loaders and wood processors, hauling diesel fuel for her equipment, and generally keeping things running around the log yard for her wood business.

So yeah, things like the audio backend, or Rob's TCI project, is aways off yet for any 4.0 release because there's just no time for it right now. We pretty much just have the basics in place to start working on it. What little time I do have in summer and fall, I've just been focusing on fixing reported bugs, which is patch release stuff, not any major release version.

Chris-AC9KH:
As a side note that has nothing to do with this PR, but since the discussion is going..........

For the 3.0.3 release I also fixed the conglomeration of tags and how they were named. So you'll have to refresh your tags in your local repos to sync with the new ones.

There should've been a tag for v2.2.0 but it never existed because tags don't transfer unless you manually do it. So I found the commit that created the old 2.2.0 release and tagged it so you can check it out on your local repo to view upon and study the general disarray of the source code back then.

This is handy for when people arrive on the scene and say, "well it did blah blah blah back in 2.2.0 and now it does bleh bleh bleh in 3.0.3" So now you can checkout that tag and (assuming you can find what made it do blah blah blah in the mess that was mostly WSJT-x) you have a quick reference to that ancient source code that will no longer even build on any modern system. You can study the awesomeness of the LiveFrequencyValidator, the MessageAveraging class, and even mainwindow that had 13,905 lines of source code in it!

Chris-AC9KH:
I'm going to rebase and merge this to master since it has already been committed to release/3.0.3 and has been proven to address the original bug.

PR #342, opened 2026-07-11
displayCallActivity.cpp: fix horizontal text alignment

Chris-AC9KH:
for "Name" and "Comment" columns in call activity table.

Before:
Screenshot 2026-07-11 at 13 48 16

After:
Screenshot 2026-07-11 at 13 59 15

### DIFF
--- a/JS8_Mainwindow/displayCallActivity.cpp
+++ b/JS8_Mainwindow/displayCallActivity.cpp
@@ -377,12 +377,12 @@ void UI_Constructor::displayCallActivity() {
                 }
 
                 auto logNameItem = new QTableWidgetItem(logDetailName);
-                logNameItem->setTextAlignment(Qt::AlignCenter);
+                logNameItem->setTextAlignment(Qt::AlignLeft | Qt::AlignVCenter);
                 logNameItem->setToolTip(logDetailName);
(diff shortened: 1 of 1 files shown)


Joe-K0OG (approved):
This is odd - I was *sure* it was left-justified before!  Indeed it **should** be left-justified.  Good catch @Chris-AC9KH.

wmiler (approved):
Huh, never noticed that...

Chris-AC9KH:
It is likely something most people won't notice. I use the logging a bit differently than most. I don't use any sort of station "logs" to log contacts, either online or on any of my computers.

Any time I've seen a new call sign in JS8Call over the last 2-3 years I've looked them up on hamqth and enter them in the log. So over time I've collected a "database" of sorts of names and locations. I use the call activity pane like shown in the screenshots, while using a very stripped down version of the band activity. So when a call appears in the call activity I have an instant reference as to who it is and where they are from. I don't know how many log entries I have in JS8Call - probably north of 500 or so. But it is NOT "logged contacts" - it is merely that database of JS8 users that I've collected that I can use for quick reference when sending MSGs etc..

I just thought that the text alignment for those two columns was a little less than ideal.

PR #344, opened 2026-07-14
OmniRig Deprecation

Chris-AC9KH:
OmniRig was scheduled for deprecation last Dec in June of this year  if the program was never bought up to modern 64-bit standards and supported by newer versions of Qt. I had a conversation with the original developer of it and he claims it doesn't need updating because it runs fine on 64-bit. That's only because x86_64 processors can still run 32-bit code. But it doesn't help that OmniRig will not compile and run on anything newer than Qt 6.6.3. The developer of it does not understand that, nor even know how to bring it up-to-date.

So this is the end of the line for OmniRig in JS8Call.


Joe-K0OG (approved):
Agreed - it's time to formalize what has been obvious for some time.  There has not been significant complaint that OmniRig support has been dropped, and there are other alternatives for that functionality.

73,
-Joe-
K0OG

wmiler (approved):

rruchte (approved):

Chris-AC9KH:
I also don't see that it's worth trying to save it by recoding it internally to bring it up-to-date. When Qt6 dropped native, out-of-the-box support for older 32-bit Active X/COM interfaces it marked the end of the line for it. Its COM port splitting is considered a security vulnerability on Windows 11. Other software suites hold onto support for it by running separate Qt5-based ports compiled as 32-bit. I don't see this as a viable thing.

Plus it is Windows-only and there are other cross-platform alternatives like Flrig that are actively maintained.

PR #347, opened 2026-07-22
UI_Constructor.cpp: add lambda's and guards on audio startup

### LINKED ISSUE #338: New Install Default Layout and Warnings
For a new install of JS8Call on a machine where it has never been run before, the default layout of the UI is horrible and unusable. And the popups that notify that audio setup is not suitable are overly aggressive. Those should never appear until the basic configuration is completed and click OK in the configuration dialog to test your new configs.

Example:
I tried a new install on Ubuntu 26 for testing. Upon program launch the audio error popups appear immediately and cannot be cleared from the screen while you attempt to make the settings in the configuration dialog. Even after making the settings with those things in the way, and clicking Ok, the audio popups are still there. The main UI is greyed out, can't move or close either of the audio notifications. Clicking Quit on the Ubuntu dock icon fails to work because the program is locked by the warnings. Clicking any of the buttons in the top audio notification fails to work because the one that's buried underneath it has to be closed first before the top one will respond.

I ended up having to use `killall js8call` at the commandline to kill the program. Then restart it and it will finally run.

The second problem is the MSG Inbox being docked to the UI by default. It only shows a narrow sliver of it on the right side of the UI. The UI has to be resized to get to the "x" to close it. On MacOS this is not a problem because the window controls are on the left side on Mac. On Windows-style interfaces like linux typically uses, the window controls are not visible until the UI is resized to show them.

Things that need to be fixed:
- do NOT fire the audio popup failure notices until after the main configuration dialog is configured and closed. And if they aren't needed because a valid audio configuration was entered, don't fire them at all.
- Either make the MSG inbox fully visible so it can be closed if wanted, or don't launch it and dock it to the UI by default.
- The waterfall controls cover up half the waterfall on initial launch of a new install. Don't launch those by default. After resizing the initially launched UI, new users will discover soon enough how to display the waterfall controls by looking at the menus.
- The default color of textEditRX is some sort of Faded Caterpillar Yellow. How about making it a soft color that better blends-in with the new UI design?

Image

Chris-AC9KH:
This prevents firing the invalid audio notifications on a fresh non-configured install before the user even has a chance to set them. This does not address everything in the issue I posted, but it addresses the most critical bug.

**Context of why this is needed:**
I found that when I did a fresh install of JS8Call on Ubuntu 26, that the invalid audio interface notifications fired immediately on program startup. I could not close them to even get the configuration dialog to work because the first notification that fired was under the second one. And the second one (that's on top) will not respond until the first one is closed. Nor did the linux UI allow me to move them out of the way. The program was totally locked up, could not quit it from the dock icon, nothing responded in the JS8Call user interface. I ended up killing JS8Call at the command line, configured the Ubuntu user interface to allow tiling open applications, and on subsequent restart used the tiling feature of the UI to get the audio notifications out of the way so it could be actually configured.

This is very bad UI design and not a very good introduction to JS8Call for a new user who has never run the program before. So I fixed it. The other issues noted in my posted issue are still on my radar, however.

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/338

### DIFF
--- a/JS8_Mainwindow/UI_Constructor.cpp
+++ b/JS8_Mainwindow/UI_Constructor.cpp
@@ -217,7 +217,12 @@ UI_Constructor::UI_Constructor(QString const &program_info,
     connect(this, &UI_Constructor::initializeAudioOutputStream, m_soundOutput,
             &SoundOutput::setFormat);
     connect(m_soundOutput, &SoundOutput::error, this,
-            &UI_Constructor::showSoundOutError);
+            [this](QString const &errorMsg) {
+                if (m_config.audio_output_device().isNull() || m_config.is_active()) {
+                    return; // nothing configured yet, or user is fixing it in Settings
+                }
+                showSoundOutError(errorMsg);
+            });
     connect(m_soundOutput, &SoundOutput::error, &m_config,
             &Configuration::invalidate_audio_output_device);
     connect(this, &UI_Constructor::outAttenuationChanged, m_soundOutput,
@@ -253,7 +258,12 @@ UI_Constructor::UI_Constructor(QString const &program_info,
             &SoundInput::resume);
     connect(this, &UI_Constructor::finished, m_soundInput, &SoundInput::stop);
     connect(m_soundInput, &SoundInput::error, this,
-            &UI_Constructor::showSoundInError);
+        [this](QString const &errorMsg) {
+            if (m_config.audio_input_device().isNull() || m_config.is_active()) {
+                return;
+            }
+            showSoundInError(errorMsg);
+        });
     connect(m_soundInput, &SoundInput::error, &m_config,
             &Configuration::invalidate_audio_input_device);
     // connect(m_soundInput, &SoundInput::status, this,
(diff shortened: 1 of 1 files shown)


Chris-AC9KH:
The other thing that might have to be addressed in UI configurations/restrictions, and only loosely related to the issues posted for this PR, is that some people are using JS8 60 for APRS now.

Screenshot 2026-07-21 at 22 02 49

It was never intended for this but you have to know somebody is gonna do it. JS8 60, with the associated decoder modifications to get it to work, was intended as a super-fast QSO mode. Not for transmitting MSG's or APRS. HB sending is already restricted for both JS8 40 and JS8 60. May have to look at locking that down so it can't be used for sending MSG's or APRS too.

rruchte (approved):

Chris-AC9KH:
Another comment on what this should do for the testers that test it:

On a fresh config if the audio input setting is not made and OK is clicked in the settings dialog, then the input audio notification should fire and switch the settings UI to the audio tab. This is the only one that's actually required for the program to do something useful, such as if it's being used for monitoring only with a Rx SDR dongle. Once an input is configured, and OK clicked in the settings dialog, it should fire the output audio notification, and the notification audio warnings in sequence, but not at the same time.

The program should then run as a monitoring-only application without bothering the user about non-configured audio interfaces on subsequent starts of JS8Call. For people that use JS8Call for monitoring and decoding only with a little SDR dongle, I think this would be expected behavior.

Any time settings are made after that it should fire the output and notification audio warnings to remind the user that they are not configured. But it shouldn't stop the program from functioning as a Rx-only app.

I think this overly-aggressive audio interface configuration warning system came from WSJT-x originally, and was never fine-tuned to be more user-friendly in JS8Call.

Joe-K0OG:
> The other thing that might have to be addressed in UI configurations/restrictions, and only loosely related to the issues posted for this PR, is that some people are using JS8 60 for APRS now.

In my opinion, we should not prevent that use of JS860.  If it works on close-in networks with strong signals, that is fine.  Yes, I think there are better tools for APRS, but I don't think we need to control what people use JS8Call for.

73,
-Joe-
K0OG

jsherer:
The principle that I've always followed with JS8 is that there should be a
strong default for the behavior we want with an option for the operator to
modify if they choose to. I believe that operators should remain in control
of their stations and their desired transmission, including how the program
behaves at their station.

Joe-K0OG (approved):
Good fix @Chris-AC9KH - this has been an irritant for too long ;)

Tested on both Windows and Linux with fresh installations, and indeed it behaves much better, not locking out the user in initial setup.

73,
-Joe-
K0OG

Joe-K0OG:
This is significant enough of an irritant that is fixed, should we consider patching it into a 3.0.4 release?

Chris-AC9KH:
I think that would be a good idea Joe. After merge to master I'll grab the commit hash and cherry pick it into a future 3.0.4 release branch. While it doesn't affect most normal users it is a fairly significant glitch for someone new to JS8Call. For some reason, on Mac it allows me to move those notifications out of the way and shoot them down so the program can be configured. On Ubuntu it turned into a significant roadblock.

Chris-AC9KH:
I merged this and cherry-picked it into a release/3.0.4 branch. No plans to release 3.0.4 any time soon, The branch is simply there as a placeholder for now, created from release/3.0.3 but not re-versioned,  standing by for anything we might want to cherry-pick into it yet.

PR #349, opened 2026-07-31
Update to Qt 6.12.0

Chris-AC9KH:
This moves master to Qt 6.12.0. Qt 6.11.1 has a regression bug that makes testing code I recently wrote for improving the MSG inbox difficult. This has been fixed in Qt 6.12, with the only other change required for linux since 6.12 now requires bundling the OpenGL library with the AppImage instead of using system or vender-installed OpenGL.

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -56,9 +56,9 @@ jobs:
     - name: Fetch Qt
       shell: pwsh
       run: |
-        curl -L -o qt611.zip "https://github.com/JS8Call-improved/js8lib/releases/download/lib%2F4.0/qt-6.11.1-windows-x64.zip"
-        Expand-Archive qt611.zip -DestinationPath qt611
-        echo "$env:GITHUB_WORKSPACE\qt611\bin" >> $env:GITHUB_PATH
+        curl -L -o qt612.zip "https://github.com/JS8Call-improved/js8lib/releases/download/lib%2F4.0/qt-6.12.0-windows-x64.zip"
+        Expand-Archive qt612.zip -DestinationPath qt612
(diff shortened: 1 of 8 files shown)


Joe-K0OG (changes requested):
Looks good, but can you add libopengl-dev to the Debian-build-deb.sh script?

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG as far as I know that one should not need it. You can run it against the code in this PR and see if it works or not, but it should.

Chris-AC9KH:
@Joe-K0OG confirmed with a build on arm64 Ubuntu 26. The .deb build script does not run linuxdeploy with the Qt plugin. It uses dpkg-deb to build a Debian .deb which only knows how to link system libraries. I don't know how the Debian people are gonna deal with this when (and if) they ever move to Qt 6.12, as they're going to have to include "non-free" software in their repo to get the right graphics compositor. But with AppImage this is not a problem - we bundle the required libraries, which are available on Ubuntu but not on Debian, and it "just works" on any distribution. It is one of the beauties of the AppImage package format.

Chris-AC9KH:
@Joe-K0OG so I'm requesting a new review. Confirmed that dpkg-deb cannot use or link with libopengl-dev, it can only use libopenglvnd which is the vender-neutral distribution of OpenGL. If you use AppImage or the user build script, then it can link with it because one uses linuxdeploy with the Qt plugin, the other links it as a system library in the QtMultimedia/QtCore object code at compile time.

The version or library that dpkg-deb uses is named libglvnd0

Joe-K0OG (approved):
All appears to work well - even the Debian .deb builder, building with Qt6.12.

Thanks @Chris-AC9KH !

73,
-Joe-
K0OG

PR #353, opened 2026-08-02
Message Inbox Improvements

### LINKED ISSUE #319: Message Inbox does not refresh after stored message is delivered/retrieved
## Version

JS8Call 3.0.2

## Summary

When the Message Inbox is open in the unfiltered “All” view, it can show stored messages that are currently held by my station for other stations. If another station retrieves one of those stored messages, the message is apparently removed/marked delivered in the backend, but the open Inbox view still shows the old entry until I hide and show the Inbox again.

## Steps to reproduce

1. Open the Message Inbox in the unfiltered / all-messages view.
2. Have at least one stored `MSG TO:` message in my local station’s inbox/storage for another callsign.
3. Let that other station retrieve the message, e.g. by querying and fetching the stored message.
4. Keep the Message Inbox window/panel open while this happens.

## Actual behavior

The delivered/retrieved message remains visible in the open Inbox table. It looks as if the message were still stored locally and available for delivery.

Only after hiding and showing the Inbox again, or otherwise forcing a refresh, the message disappears from the list.

## Expected behavior

When a stored message is successfully delivered/retrieved and removed or marked delivered in the local storage, the open Message Inbox view should update automatically.

At minimum, the UI should not continue to show the message as if it were still pending.

## Why this is confusing

The Inbox view appears to show the current state of local message storage. If a message remains visible after successful delivery, the user may assume that it is still pending, has not been retrieved, or may be delivered again.

## Possible fix

Refresh the Message Inbox table after a stored message is marked delivered, or emit a signal from the inbox/storage update path that causes the MessagePanel to reload its current filtered view.

### LINKED ISSUE #320: Message Inbox call filter is hidden and cannot be reset from within the Inbox UI
## Version

JS8Call 3.0.2

## Summary

The Message Inbox can be shown either as an unfiltered “All” view or as a callsign-filtered view when opened from a callsign context menu. However, once the Inbox is filtered by a callsign, the active filter is not clearly visible inside the Inbox UI itself, and there is no obvious control inside the Inbox to reset the filter back to “All”.

## Steps to reproduce

1. Open the Message Inbox normally so it shows all messages.
2. Select a station/callsign in the call list.
3. Use the callsign context menu and choose the option to show that station’s Inbox/messages.
4. Look at the Message Inbox panel.

## Actual behavior

The Inbox contents are filtered to that callsign, but the filter is not shown as an obvious filter control inside the Inbox panel.

There is also no visible “clear filter”, “show all”, or callsign selector inside the Inbox panel itself. The practical way back to the unfiltered view is to hide/show the whole Inbox again or use another external UI path.

## Expected behavior

The Inbox should make the active filter visible and directly editable/resettable inside the Inbox panel.

For example, the top of the Message Inbox could contain a filter selector such as:

* All messages
* Messages for/from selected callsign
* Specific callsign: [CALL]

and a clear/reset button.

## Suggested improvement

Add an explicit filter control to the Message Inbox itself.

This would make the feature more discoverable and less surprising. It may also make the callsign-list context-menu filtering less necessary, or at least make it clear that the context-menu action simply changes the Inbox filter.

## Why this matters

Currently it feels as if the Inbox has changed into a different hidden mode. A user can enter the filtered view through the callsign context menu, but the Inbox panel itself does not clearly explain that a filter is active or how to return to the unfiltered view.

Chris-AC9KH:
Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/319
Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/320

In addition to making the Message Inbox update when a stored MSG is retrieved, and adding the resettable filter to the inbox I implemented the following:
- change column order so column 2 is From and column 3 is To. This makes the message inbox display a more common sense column order when docked in a narrow panel on either side of the main window
- displays the floating message inbox as a Qt::Window instead of a Qt::Tool. This adds the normal chrome (max/min/hide) window controls to the floating panel. There is some caveats with this, hence the extensive doxy block in ensureMessageDock.cpp to thoroughly document the issues for developers who may want to make changes in the future, outlining what's functional, what's not, what I tried that worked, and what I tried that didn't work.
- Updated and coded this for Qt 6.12. Qt 6.11.1 has a regression bug with dockable panel handling that is fixed in Qt 6.12. While this may also be fixed in Qt 6.11.2 when it is released on Aug 13, Qt 6.12 is the next LTS release and is getting all the developer attention. Qt 6.11 is a stepping stone to 6.12, so may as well get on the 6.12 code now. Qt 6.12 also allowed me to make significant improvements in this code over earlier Qt versions so the only real outlier is Wayland-based desktops on linux, which is thoroughly noted in the doxy comments.
- moved the ensureMessageDock UI_Constructor member function from mainwindow.cpp to its own file in JS8_Mainwindow

Included here is a couple screenshots of the new resettable filter in both docked and floating mode.

Screenshot 2026-08-02 at 07 10 44

Screenshot 2026-08-02 at 07 11 01

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
-set(CMAKE_OSX_DEPLOYMENT_TARGET 14.0)
+set(CMAKE_OSX_DEPLOYMENT_TARGET 14.4)
 
 #------------------------------------------------------------------------------#
 # Project configuration. The software build version is set here. If the version
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,6 +307,7 @@ target_sources(
   JS8_Mainwindow/dataSink.cpp
   JS8_Mainwindow/displayBandActivity.cpp
   JS8_Mainwindow/displayCallActivity.cpp
+  JS8_Mainwindow/ensureMessageDock.cpp
   JS8_Mainwindow/initializeDummyData.cpp
   JS8_Mainwindow/initializeGroupMessage.cpp
   JS8_Mainwindow/networkMessage.cpp
--- a/JS8_Mainwindow/UI_Constructor.cpp
+++ b/JS8_Mainwindow/UI_Constructor.cpp
@@ -1466,3 +1466,30 @@ UI_Constructor::UI_Constructor(QString const &program_info,
     if (!m_valid)
         throw std::runtime_error{"Fatal initialization exception"};
 }
+
+bool UI_Constructor::markMsgDelivered(int mid, Message msg) {
+    Inbox inbox(inboxPath());
+    if (!inbox.open()) {
+        return false;
+    }
+
+    msg.setType("DELIVERED");
+    bool ok = inbox.set(mid, msg);
+    if (ok) {
+        emit messageAdded(mid);
+    }
+    return ok;
+}
+
+bool UI_Constructor::markGroupMsgDeliveredForCallsign(int msgId, QString callsign) {
+    Inbox inbox(inboxPath());
+    if (!inbox.open()) {
+        return false;
+    }
+
+    bool ok = inbox.markGroupMsgDeliveredForCallsign(msgId, callsign);
+    if (ok) {
+        emit messageAdded(msgId);
+    }
+    return ok;
+}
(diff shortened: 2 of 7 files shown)


Joe-K0OG (approved):
All appears to work fine on Linux and Windows.

73,
-Joe-
K0OG

Chris-AC9KH:
IMO it works ok on linux and windows, and it does keep inline with what those platforms mandate for handling floating app windows. My hope is that eventually the Wayland developers will realize the error of their ways and maybe fix Wayland to be more like users expect it to be. In the meantime, selecting distributions that use Xwayland appears to be a better choice than pure Wayland distributions like Ubuntu/GNOME. Or maybe you can install Xwayland on Ubuntu and "fix it" that way, I'm not sure.

Chris-AC9KH:
@Joe-K0OG I found out you can't replace Wayland on Ubuntu 26 because they got it "baked in". But there is a way around it. Run JS8Call from the terminal with `QT_QPA_PLATFORM=xcb ./JS8Call.AppImage` and it "fixes it".

rruchte (approved):

PR #356, opened 2026-08-06
Convert replyPushButton from QPushButton to QToolButton to support a menu-driven choice between three reply targets.

### LINKED ISSUE #321: Replying to a retrieved relayed stored message addresses the relay instead of the original sender
## Version

JS8Call 3.0.2

## Summary

The reply behavior for retrieved relayed store-and-forward messages is unintuitive.

Example scenario:

* `CALL1` stores a message for `CALL2` on `RELAY`: Call1: Relay msg to: call2 something
* `CALL2` retrieves the stored message from `RELAY`. Call2: Relay query msg xyz
* `CALL2` then sees a message in the Inbox that appears to come from `RELAY`, with the original sender only included in the message text as something like `FROM CALL1`.
* If `CALL2` uses Reply, the reply is addressed to `RELAY`, not to `CALL1`.

From the user’s point of view, the message is semantically from `CALL1`, so replying to `RELAY` is surprising.

## Steps to reproduce

1. `CALL1` sends/stores a message for `CALL2` using `RELAY` as the store-and-forward station.
2. `CALL2` retrieves the message from `RELAY`.
3. `CALL2` opens the Inbox and selects the received/retrieved message.
4. `CALL2` clicks Reply.

## Actual behavior

The reply is generated toward the relay station/path, because the locally received message is technically a normal `MSG` from `RELAY`, with the original sender only included in the text as `FROM CALL1`.

This makes the generated reply target `RELAY`.

## Expected behavior

The Reply action should help the user reply to the original sender, or at least make the relay/original-sender distinction explicit.

Possible expected behaviors:

1. Reply directly to the original sender:

   * `CALL1 MSG [reply text]`

2. Or, if the relay path should be preserved, generate a store-and-forward reply through the relay:

   * `RELAY MSG TO: CALL1 [reply text]`

3. Or prompt/offer a choice:

   * “Reply to original sender CALL1”
   * “Reply via relay RELAY”

## Suggested improvement

When a retrieved stored message contains an original sender marker such as `FROM CALL1`, the Inbox could parse and store that original sender as message metadata instead of treating it only as plain text.

The Inbox could then display something like:

* From: `CALL1`
* Via: `RELAY`

and the Reply button could use the original sender as the semantic reply target.

## Compatibility concern

I understand that the retrieved message is transmitted as a normal `MSG` from the relay station, and that changing the stored Inbox metadata may not be fully compatible with the current message model. But from a user perspective, the current behavior is confusing: the user received a message from `CALL1`, but Reply answers `RELAY`.

Even if the internal protocol cannot be changed easily, the UI could improve this by detecting `FROM <CALL>` in retrieved stored messages and offering a clearer reply target.

Beiri22:
This is a prototype implementation of a flexible reply button that supports replying either to the relay or to the original sender, as discussed in #321. It is mainly intended for review and evaluation at this stage and does not have to be adopted in exactly this form. After the extensive discussion, I wanted to explore how large and invasive the required changes would actually be. The overall scope of the implementation turned out to be relatively small.

Fixes https://github.com/JS8Call-improved/JS8Call-improved/issues/321

### DIFF
--- a/JS8_UI/MessagePanel.ui
+++ b/JS8_UI/MessagePanel.ui
@@ -218,7 +218,7 @@
        </spacer>
       </item>
       <item>
-       <widget class="QPushButton" name="replyPushButton">
+       <widget class="QToolButton" name="replyPushButton">
         <property name="minimumSize">
          <size>
           <width>0</width>
@@ -227,6 +227,12 @@
         </property>
         <property name="text">
          <string>Reply</string>
+        </property>
+        <property name="toolButtonStyle">
+         <enum>Qt::ToolButtonStyle::ToolButtonTextOnly</enum>
+        </property>
+        <property name="autoRaise">
+         <bool>false</bool>
         </property>
        </widget>
       </item>
--- a/JS8_UI/MessagePanel.cpp
+++ b/JS8_UI/MessagePanel.cpp
@@ -10,7 +10,9 @@
 #include <QDateTime>
 #include <QMenu>
 #include <QRegularExpression>
+#include <QStyle>
 #include <QTimeZone>
+#include <QToolButton>
 
 #include <algorithm>
 
@@ -21,6 +23,21 @@ auto pathSegs(QString const &path) {
   auto segs = path.split('>');
   std::reverse(segs.begin(), segs.end());
   return segs;
+}
+
+void setToolButtonMenu(QToolButton *button, QMenu *menu) {
+  button->setPopupMode(menu ? QToolButton::MenuButtonPopup
+                            : QToolButton::DelayedPopup);
+  // Qt style sheets cache property-selector matches. Re-polish after changing
+  // popupMode so the split-button padding rule is applied or removed now.
+  button->style()->unpolish(button);
+  button->style()->polish(button);
+
+  // setMenu() invalidates QToolButton's cached size hint. Keep it after the
+  // style refresh so the next hint uses the newly selected padding rule.
+  button->setMenu(menu);
+  button->updateGeometry();
+  button->update();
 }
 } // namespace
 
@@ -298,8 +315,7 @@ QString MessagePanel::prepareStoreForwardReplyMessage(QString path,
 }
 
 void MessagePanel::resetReplyButton() {
-  ui->replyPushButton->setMenu(nullptr);
-  ui->replyPushButton->setPopupMode(QToolButton::DelayedPopup);
+  setToolButtonMenu(ui->replyPushButton, nullptr);
 
   replyToPathAction->setText(tr("Reply"));
   replyToPathAction->setData(QVariant());
@@ -347,8 +363,7 @@ void MessagePanel::configureReplyButton(int row, const QString &messageText) {
   replyToOriginalAction->setData(prepareReplyMessage(originalSender, placeholder));
   replyToOriginalAction->setEnabled(true);
 
-  ui->replyPushButton->setMenu(replyMenu);
-  ui->replyPushButton->setPopupMode(QToolButton::MenuButtonPopup);
+  setToolButtonMenu(ui->replyPushButton, replyMenu);
 }
 
 void MessagePanel::emitReplyAction(QAction *action) {
--- a/JS8_UI/MessagePanel.cpp
+++ b/JS8_UI/MessagePanel.cpp
@@ -1,12 +1,15 @@
 #include "MessagePanel.h"
 #include "JS8_Include/EventFilter.h"
 #include "JS8_Main/Radio.h"
+#include "JS8_Main/Varicode.h"
 #include "JS8_Widgets/DateTableWidgetItem.h"
 #include "JS8_Widgets/SemiSortableHeader.h"
 #include "ui_MessagePanel.h"
 
+#include <QAction>
 #include <QDateTime>
 #include <QMenu>
+#include <QRegularExpression>
 #include <QTimeZone>
 
 #include <algorithm>
@@ -27,6 +30,16 @@ MessagePanel::MessagePanel(QString inboxPath, QWidget *parent)
   inbox->open();
 
   ui->setupUi(this);
+
+  replyMenu = new QMenu(ui->replyPushButton);
+  replyToPathAction = replyMenu->addAction(QString());
+  replyToOriginalViaPathAction = replyMenu->addAction(QString());
+
+  ui->replyPushButton->setPopupMode(QToolButton::MenuButtonPopup);
+  ui->replyPushButton->setDefaultAction(replyToPathAction);
+  connect(replyToPathAction, &QAction::triggered, this, [this]() { emitReplyAction(replyToPathAction); });
+  connect(replyToOriginalViaPathAction, &QAction::triggered, this, [this]() { emitReplyAction(replyToOriginalViaPathAction); });
+  resetReplyButton();
 
   // connect selection model changed
   connect(ui->messageTableWidget->selectionModel(), &QItemSelectionModel::selectionChanged, this,
@@ -277,9 +290,71 @@ QString MessagePanel::prepareReplyMessage(QString path, QString text) {
   return QString("%1 MSG %2").arg(path).arg(text);
 }
 
+QString MessagePanel::prepareStoreForwardReplyMessage(QString path,
+                                                      QString recipient,
+                                                      QString text) {
+  return QString("%1 MSG TO:%2 %3").arg(path).arg(recipient).arg(text);
+}
+
+void MessagePanel::resetReplyButton() {
+  ui->replyPushButton->setMenu(nullptr);
+
+  replyToPathAction->setText(tr("Reply"));
+  replyToPathAction->setData(QVariant());
+  replyToPathAction->setEnabled(false);
+
+  replyToOriginalViaPathAction->setText(QString());
+  replyToOriginalViaPathAction->setData(QVariant());
+  replyToOriginalViaPathAction->setEnabled(false);
+}
+
+void MessagePanel::configureReplyButton(int row, const QString &messageText) {
+
+  auto *pathItem = ui->messageTableWidget->item(
+      row, ui->messageTableWidget->columnCount() - 5);
+  if (!pathItem) {
+    return;
+  }
+
+  const auto path = pathItem->data(Qt::UserRole).toString().trimmed();
+  if (path.isEmpty()) {
+    return;
+  }
+
+  const auto placeholder = QStringLiteral("[MESSAGE]");
+  replyToPathAction->setText(tr("Reply to %1").arg(path));
+  replyToPathAction->setData(prepareReplyMessage(path, placeholder));
+  replyToPathAction->setEnabled(true);
+
+  static const QRegularExpression fromSignature(R"((?:^| )FROM (?<callsign>\S+)$)");
+  const auto match = fromSignature.match(messageText);
+  if (!match.hasMatch()) {
+    return;
+  }
+
+  const auto originalSender = match.captured("callsign");
+  if (originalSender.startsWith('@') || !Varicode::isValidCallsign(originalSender, nullptr)) {
+    return;
+  }
+
+  replyToOriginalViaPathAction->setText(tr("Reply to %1 via %2").arg(originalSender, path));
+  replyToOriginalViaPathAction->setData(prepareStoreForwardReplyMessage(path, originalSender, placeholder));
+  replyToOriginalViaPathAction->setEnabled(true);
+  ui->replyPushButton->setMenu(replyMenu);
+}
+
+void MessagePanel::emitReplyAction(QAction *action) {
+  const auto message = action->data().toString();
+  if (!message.isEmpty()) {
+    emit replyMessage(message);
+  }
+}
+
 void MessagePanel::messageTableSelectionChanged(
     const QItemSelection &selected,
     const QItemSelection &/*deselected*/) {
+
+  resetReplyButton();
 
   auto items = ui->messageTableWidget->selectedItems();
   if (items.isEmpty() || items.size() > ui->messageTableWidget->columnCount()) {
@@ -300,6 +375,7 @@ void MessagePanel::messageTableSelectionChanged(
 
   auto text = item->data(Qt::UserRole).toString();
   ui->messageTextEdit->setPlainText(text);
+  configureReplyButton(row, text);
 
   // Mark deselected as read in table
   auto selectedItemIndexes = selected.indexes();
@@ -327,21 +403,4 @@ void MessagePanel::messageTableSelectionChanged(
   }
 
   markMessageRead(mid);
-}
-
-void MessagePanel::on_replyPushButton_clicked() {
-  auto row = ui->messageTableWidget->currentRow();
-
-  // from column
-  auto item = ui->messageTableWidget->item(
-      row, ui->messageTableWidget->columnCount() - 5);
-  if (!item) {
-    return;
-  }
-
-  auto path = item->data(Qt::UserRole).toString();
-  auto text = "[MESSAGE]";
-  auto message = prepareReplyMessage(path, text);
-
-  emit replyMessage(message);
 }
--- a/JS8_UI/MessagePanel.cpp
+++ b/JS8_UI/MessagePanel.cpp
@@ -35,7 +35,6 @@ MessagePanel::MessagePanel(QString inboxPath, QWidget *parent)
   replyToPathAction = replyMenu->addAction(QString());
   replyToOriginalViaPathAction = replyMenu->addAction(QString());
 
-  ui->replyPushButton->setPopupMode(QToolButton::MenuButtonPopup);
   ui->replyPushButton->setDefaultAction(replyToPathAction);
   connect(replyToPathAction, &QAction::triggered, this, [this]() { emitReplyAction(replyToPathAction); });
   connect(replyToOriginalViaPathAction, &QAction::triggered, this, [this]() { emitReplyAction(replyToOriginalViaPathAction); });
@@ -298,6 +297,7 @@ QString MessagePanel::prepareStoreForwardReplyMessage(QString path,
 
 void MessagePanel::resetReplyButton() {
   ui->replyPushButton->setMenu(nullptr);
+  ui->replyPushButton->setPopupMode(QToolButton::DelayedPopup);
 
   replyToPathAction->setText(tr("Reply"));
   replyToPathAction->setData(QVariant());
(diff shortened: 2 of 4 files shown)


Chris-AC9KH:
@Beiri22 although I haven't tested it, this code looks pretty solid to me. I'm going to run the workflows on it and if it builds OK, and works as designed, I'd have no problems merging this to master.

Chris-AC9KH:
@Joe-K0OG there is linux and Windows builds now available in the workflows for this, if you want to try it out.

Joe-K0OG:
@Chris-AC9KH I built @Beiri22 source reply branch on both Windows and Linux and tested it between my two stations (changing callsigns to simulate a third party).  It builds fine on both platforms, and as far as I can tell the option to either directly reply or reply via works perfectly.  Good solution!

73,
-Joe-
K0OG

JS8_UI/MessagePanel.ui
```diff
@@ -218,7 +218,7 @@
        </spacer>
       </item>
       <item>
-       <widget class="QPushButton" name="replyPushButton">
+       <widget class="QToolButton" name="replyPushButton">
```
Chris-AC9KH, line 221: Built and tested on Mac. We replaced the QPushButton here with QToolButton so on Mac we get the ugly GTK/Linux-style button:

Image

It should look like this:

Image

The reason is because QToolButton was not defined in styles.h which defines the native look and feel for every platform. You need to add this to styles.h like so, and recommended would be up to update the doxy block above it like shown here:

```
/**
 * @brief Returns a platform-native QPushButton/QToolButton stylesheet.
 *
 * Generates a stylesheet that approximates the native button conventions of the
 * specific platform, using system-appropriate fonts, geometry, and accent
 * colors. Hover, pressed, and disabled pseudo-states are defined for all active
 * variants. QToolButton's native menu-indicator glyph is suppressed since
 * platform styles (notably QMacStyle) draw it outside the stylesheet's box
 * model, producing an inconsistent look next to the styled button body;
 * callers relying on a popup menu should indicate this in the button's own
 * text (e.g. appending "▾") if a visual affordance is needed
 * 
 * @return A QSS QString suitable for use with @c QWidget::setStyleSheet(),
 *         or an empty default Qt style on unsupported platforms.
 */
inline QString buttonStyle() {
#if defined(Q_OS_WIN)
    return R"(
        QPushButton, QToolButton {
            background-color: #6699ff;
            color: black;
            border: none;
            border-radius: 4px;
            padding: 3px 9px;
            min-height: 15px;
            max-height: 15px;
            font-family: "Segoe UI";
        }
        QPushButton:hover, QToolButton:hover {
            background-color: #4d7fff;
            color: white;
        }
        QPushButton:pressed, QToolButton:pressed {
            background-color: #003EAA;
        }
        QPushButton:disabled, QToolButton:disabled {
            background-color: #ececec;
            color: #888888;
        }
        QToolButton::menu-indicator { image: none; }
    )";

#elif defined(Q_OS_MACOS)
    return R"(
        QPushButton, QToolButton {
            background-color: #6699ff;
            color: black;
            border: none;
            border-radius: 6px;
            padding: 3px 9px;
            min-height: 15px;
            max-height: 15px;
            font-family: "-apple-system";
        }
        QPushButton:hover, QToolButton:hover {
            background-color: #4d7fff;
            color: white;
        }
        QPushButton:pressed, QToolButton:pressed {
            background-color: #003EAA;
        }
        QPushButton:disabled, QToolButton:disabled {
            background-color: #ececec;
            color: #888888;
        }
        QToolButton::menu-indicator { image: none; }
    )";

#elif defined(Q_OS_LINUX)
    return R"(
       QPushButton, QToolButton {
           background-color: #6699ff;
           color: black;
           border: none;
           border-radius: 5px;
           padding: 3px 9px;
           font-family: "Ubuntu", "Noto Sans";
       }
       QPushButton:hover, QToolButton:hover {
           background-color: #4d7fff;
       }
       QPushButton:pressed, QToolButton:pressed {
           background-color: #003EAA;
       }
       QPushButton:disabled, QToolButton:disabled {
           background-color: #ececec;
           color: #888888;
       }
       QToolButton::menu-indicator { image: none; }
   )";

#else
    return QString();
#endif
}
```

Chris-AC9KH:
@Beiri22 a note with the above requested change to this PR: Mac will automatically draw the dropdown menu indicator (or not) depending on the menu items in the dropdown if it exists. Windows or linux will not do this. So in your testing on Linux and Windows if you want the native dropdown indicator on your QToolButton just remove the `QToolButton::menu-indicator { image: none; }` from those platform definitions in styles.h

Beiri22:
Thanks for the note. I may be misunderstanding something here, but I can't actually find QToolButton::menu-indicator { image: none; } in the current styles.h platform definitions. I may simply be missing something obvious here, as this part of the codebase is a little outside the area I usually work in. So I'm not quite sure what exactly I should remove on Linux/Windows to get the native dropdown indicator behavior you described.

ChatGPT had also suggested one additional change related to the button's behavior, see commit.

Chris-AC9KH:
@Beiri22 no, we do not want to define any styles inline in the code because that will mess up people's custom stylesheets. I reverted your second commit and helped you out with this by defining the style in styles.h. That file defines the platform-native styles for all platforms and all that was required to was to add QToolButton to the styles.

Please don't use AI to do your coding - 90% of the time it doesn't know what's going on, 95% of the time it writes code that I won't approve, and 99.9% of the time it writes code that needs changes because of the 90% of the time it doesn't know what's going on.

Chris-AC9KH:
@Beiri22 I'll outline a little more what I changed here for you:

My revert of your second commit removed the popup-mode toggling logic entirely. 

- setPopupMode(QToolButton::MenuButtonPopup) is called exactly once in the constructor, and never changed again
- resetReplyButton() and configureReplyButton() only touch setMenu(), not setPopupMode()
- So replyPushButton is permanently in MenuButtonPopup mode, regardless of whether it currently has a menu attached or not

That matters because in Qt, the split-button arrow (SC_ToolButtonMenu) is drawn whenever popupMode() == MenuButtonPopup. It does not check whether an actual menu is set. hasMenu() only adds a separate style-option flag; it doesn't suppress the arrow region. So the dropdown arrow area is always present on the button, even right after resetReplyButton() clears the menu.

This stems from my original review where I noted that we changed to a QToolButton, which is not defined in the platform styles, making the new button look and feel inconsistent with the rest of the UI. So I reverted your second commit and defined that for you because it is obviously outside the scope of what your AI program you are using to code this understands.

So, since you originally coded this to display that menu dropdown indicator, and presumably you are testing on Windows or linux, is this normal behavior that you intended? I specifically removed this behavior in styles.h on Mac, since such a thing is considered bad UI design on MacOS. The reason we do this in styles.h instead of inline in the code is because these platform native styles differ. So when I defined QToolButton in styles.h I applied `QToolButton::menu-indicator { image: none; }` for Mac since that is technically correct for MacOS. Is it correct for Windows and linux? That is what you have to test here and make the decision as to whether or not to add that for those two platforms in styles.h

Joe-K0OG:
@Beiri22 @Chris-AC9KH The logic is not quite right I think.  When I have a relayed message on my station, the drop-down list should indicate to reply *directly* to the originator, **or** reply to originator via the intermediate station.  As you can see from the screenshot from a test build of Beiri22's commit cc2754fa, it is showing to reply to the intermediate **or** reply to originator via the intermediate station.  The first option should be to reply to originator, in my example, "Reply to KB0FPD", ***not*** "Reply to K0OG".
image

73,
-Joe-
K0OG

Chris-AC9KH (changes requested):
@Beiri22 I did the testing for you since I have the capability to test on all three platforms and the button behavior and platform-native rendering is correct after my last force push to styles.h

I'm requesting one further change to the commit message rather than just "flexible reply button". Do an interactive rebase on this and change that commit message to:

```
MessagePanel.cpp: Convert replyPushButton from QPushButton to
QToolButton to support a menu-driven choice between two reply targets:
- Reply to the relay path directly (existing behavior)
- Reply to the original sender via the relay path, when the message
  text contains a recognizable "FROM <callsign>" signature
```

The commit message "flexible reply button" is not very descriptive and doesn't mean much in the commit history, or for the future release notes.

Joe-K0OG:
Otherwise, besides the reply to logic I mentioned above, commit cc2754fa button behavior and appearance on both Linux and Windows is good.

73,
-Joe-
K0OG

Joe-K0OG:
> MessagePanel.cpp: Convert replyPushButton from QPushButton to
> QToolButton to support a menu-driven choice between two reply targets:
> - Reply to the relay path directly (existing behavior)
> - Reply to the original sender via the relay path, when the message
>   text contains a recognizable "FROM <callsign>" signature

Per my previous comment, I thought the idea was to give the option to either reply *directly* to the originating station, **or** reply via the relay path to the originating station.  That would seem more useful to me, considering the original conversation.

73,
-Joe-
K0OG

Chris-AC9KH:
@Joe-K0OG replying directly to the originating station is never an option here because with a relay at no time has your station ever been in communication with the originator of the MSG. The message Inbox can only form a reply to a station you have directly communicated with so you have two choices:
- reply to the relay station where you got the MSG from
- reply to the originator thru the relay

Replying directly to the originator will send it into the bit bucket in outer space.

Joe-K0OG:
@Chris-AC9KH There is little or no reason to reply to the intermediary in the case of a relayed message being delivered by him to me.  Later, when I pick up the message, the originating station may be available to me, so I can reply directly to him.  This can be determined either by my direct communication with the originating station, or by my SNR query to him.  If you do not want to put "Reply to originating station" in the pull-down list, the only logical option to present is "Reply to originating station via relay station".

Joe-K0OG:
Maybe there is some confusion:  One could either store a reply at the intermediary for the originator (intermediary MSG TO: originator [reply message]), or one could relay via (intermediary>originator [reply message]).  Maybe all those options should be present?  The code I tested did not do the ">" via but instead stored a reply at the intermediary ("MSG TO:" format).

Beiri22:
@Chris-AC9KH Thanks, I think I understand now what you meant with the changes in `styles.h`. What confused me was that I did not see that change in Git at the time when I made my second commit. Looking at the PR history now, I think part of my confusion came from the force-pushes rewriting the branch history. I may be missing something about the workflow here, but I think a normal revert/follow-up commit would have made the sequence much easier to understand afterwards, because the original commit, the revert, and the replacement would all still be visible in order.

Regarding my second commit, though, I think that may actually be a separate issue from the styling change. As I understand your own explanation, `MenuButtonPopup` causes Qt to treat the `QToolButton` as a split/menu button independently of whether `setMenu(nullptr)` is currently used. In other words, the popup mode itself affects the button's behaviour/layout, while `styles.h` controls how that button is drawn on the different platforms. That was the reason for my second commit: when there is no alternative reply action, I switch the button back to a normal/non-split mode, and only switch it to `MenuButtonPopup` when the additional reply menu is actually attached. That was the intention.

So I think both changes may be needed for different reasons:
* the `styles.h` change makes the `QToolButton` visually consistent with the existing buttons;
* changing `popupMode` dynamically makes it behave as an ordinary button when there is no menu, and as a split button only when the menu is available.

Is there something I am overlooking here? From your explanation of how `MenuButtonPopup`, `SC_ToolButtonMenu`, and `hasMenu()` interact, it actually sounds to me like the behaviour change from my second commit is still desirable independently of the stylesheet changes. If so, I would suggest keeping the `styles.h` changes and restoring the small `setPopupMode()` changes from my second commit. Could you please give this another thought and let me know what you think?

Beiri22:
@Joe-K0OG As I understand the situation, the message was originally written by KB0FPD and then stored on the intermediate/store-and-forward station K0OG, from which we retrieved it. Since we never actually heard KB0FPD directly, my understanding from the earlier discussion in #321 was that replying via K0OG - i.e. storing the reply there for KB0FPD to retrieve - would be the most logical way to reply to the original sender in this situation. Chris also pointed out that replying directly to the intermediate station can be useful, for example simply to thank K0OG for providing the store-and-forward service. That is why the current implementation has: Reply to K0OG, Reply to KB0FPD via K0OG -- with the first one as the default action. At least that is how I understood the conclusion of the first part of the discussion in #321.

I left out a third option, Reply to KB0FPD directly, for that reason - not because it would be difficult to implement. Adding that option with the current architecture should be quite straightforward. So if the consensus is that we should offer all three possibilities, I am completely fine with adding it. I think we just need to agree on which options we actually want to expose, and which one should be the default.

Chris-AC9KH:
@Beiri22 yes, you are correct that the force pushes rewrote the branch history. I ended up having to do that due to the fact that it took several commits, builds and testing to make sure it worked properly and I needed your second commit out of the way to make sure of this from styles.h. I ended up not needing the `QToolButton::menu-indicator { image: none; }` exclusion for any of the three platforms, which due to the differences in how Qt renders these items on Mac vs Windows vs Linux can only be determined thru actual testing on all three platforms.

At this point, after doing the testing and force-pushing those changes,  I agree that your second commit can be added back in but I think it is dead code. The simple addition of QToolButton to styles.h accomplishes the same thing, plus defines the button style, which your second commit did not do. If you build the current code and test it you will see that it appears as a regular button when applicable, and adds the dropdown indicator if there is a menu selection to be made. This is handled dynamically by the QToolButton class without having to define it.

So on your end, add your second commit back in and test it. I think you will find that behavior does not change. All that was missing was to define it in the global stylesheet instead of inline in the code. Since you seemed to be confused about that, and I wrote that original stylesheet code, I just figured it was simpler for me to contribute that to your PR rather than making you figure it out.

Chris-AC9KH:
@Beiri22 @Joe-K0OG you can add in that direct reply as a menu option, as it is harmless. It is technically incorrect based on the fact that JS8Call's messaging system is based on realtime in the call activity pane, and hence such a menu item would fail. But viewing and replying to a MSG that is several days old, and that call no longer exists in the call activity panel, also fails. So from that standpoint, adding it in is harmless.

What this stems from is a tendency in Windows and linux (which is mostly a Windows copy) UI design to add in complex context menus with some items that don't even apply, just because they MIGHT apply someday. This is severely frowned upon in Mac UI design which stresses clean, uncluttered, user-friendly design. So such a thing in a Mac UI design would check the call activity panel to see if such a menu item is even applicable before plastering it into a context menu wholesale.

So if Joe thinks it has to be there, then add it in. At some point in the future when I'm looking at UI cleanup issues I will fix it for Mac, using preprocessor guards if necessary, to keep the UI on Mac clean and uncluttered with unnecessary menu items that will fail unless the proper checks are run to make sure it's a valid reply path before even displaying it. The same thing applies to the current Reply button - it makes no sense to even have that ungreyed out if the path does not exist to deliver a reply.

Beiri22:
@Chris-AC9KH I have checked the changes of my second commit; just under linux.

Previous  / your code:
Bildschirmfoto vom 2026-08-09 21-54-45
Bildschirmfoto vom 2026-08-09 21-54-49

With "commit 2":
Bildschirmfoto vom 2026-08-09 21-56-25
Bildschirmfoto vom 2026-08-09 21-56-29

I dont know if this is different in win / mac, but the behaviour with "commit 2" is what I intended.

Chris-AC9KH:
@Beiri22 ok, add in commit #2 as I THINK it will do the same thing on Mac. I can test that. As far as I can tell, this code is working code that can be merged. If you're happy with it, remove the WIP label, update the main commit message to this  so it's a little more descriptive:

```
MessagePanel.cpp: Convert replyPushButton from QPushButton to
QToolButton to support a menu-driven choice between two reply targets:
- Reply to the relay path directly (existing behavior)
- Reply to the original sender via the relay path, when the message
  text contains a recognizable "FROM <callsign>" signature
```

Add in Joe's menu item if you want - I'll clean that up later for Mac (at the minimum) to get rid of dead context menu items that don't apply, because I got way bigger changes to this pending than what this PR involves.

Beiri22:
@Chris-AC9KH Do you see a simple way to keep some horizontal padding between the end of the button text and the start of the dropdown subcontrol on the right?

Beiri22:
@Chris-AC9KH Isn't it possible to override the final commit message when using "Squash and merge" anyway? If so, I'd prefer to leave the existing commit messages untouched, just to avoid creating another round of force-push confusion in the PR history. See https://docs.github.com/en/pull-requests/reference/pull-request-merges "When you squash and merge, GitHub generates a default commit message that you can edit.". I would add the third option to the message:
```
MessagePanel.cpp: Convert replyPushButton from QPushButton to
QToolButton to support a menu-driven choice between three reply targets:
- Reply to the relay path directly (existing behavior)
- Reply to the original sender via the relay path, when the message
  text contains a recognizable "FROM <callsign>" signature
- Reply to the original sender directly, when the message
  text contains a recognizable "FROM <callsign>" signature
```

Beiri22:
@Chris-AC9KH  I have used AI to fix the right spacing. Please have a look at the last commit; if you don't like it, be free to undo this commit or tell me to do so.

Chris-AC9KH:
@Beiri22 the button styling code is fine.

But one other small bug you need to address. You are using:

```
static const QRegularExpression fromSignature(R"((?:^| )FROM (?<callsign>\S+)$)");
  const auto match = fromSignature.match(messageText);
  if (!match.hasMatch()) {
    return;
```
as the trigger for your dropdown menu. This will fail. The software can send NEXT MSG ID xxx appended to it and the $ anchor sets the end of the string so it doesn't match and you'll get no dropdown menu.

You need to loosen that a bit, something like:

`static const QRegularExpression fromSignature(R"((?:^| )FROM (?<callsign>\S+)(?: |$))");`

Or maybe something like this to make it non-capturing group:

`static const QRegularExpression fromSignature(R"((?:^| )FROM (?<callsign>\S+)(?:\s+NEXT MSG ID \d+)?\s*$)");`

Beiri22:
I see. I used ```  static const QRegularExpression fromSignature(R"((?:^| )FROM (?<callsign>\S+)(?: NEXT MSG ID \d+(?: \+\d+)?)?$)"); ``` to account for NEXT MSG [ID] and NEXT MSG [ID] +[CNT]

Chris-AC9KH (approved):
I think it looks good. We'll run the workflows and if that passes this is good to merge

Chris-AC9KH:
Merged - thanks!

Chris-AC9KH:
@Beiri22 just a note on your question on squashing and rebasing. It is best if the author of the code does this on their local working branch. Otherwise whoever does it takes ownership of the commit. So I elected to leave this one as-is for merging.

In future, study interactive rebasing. Say you have five commits on a branch and you want to change the commit message on the oldest commit. You can run `git rebase -i HEAD~5` and this will display the last five commits in your history. If you want to re-word one, change the "pick" at the beginning of the line to "r" and save it. This will bring up a new dialog in your commandline text editor where you can now reword that commit and save it with a new commit message.

Squashing works the same way. You can replace the "pick" with "s" and that will attempt to squash that commit into the one above it in the list. That is how you can submit a PR with a clean commit history (take a look at my most recent PR). It consists of only three commits that each do specific thing. But in developing that code I probably had 50-60 commits. I cleaned all those up before posting the PR with interactive rebasing.

PR #358, opened 2026-08-11
Push Notification Handler

Chris-AC9KH:
This consists of three commits.

- The first is an int to 64-bit bug fix that the second one depends on
- The second migrates the inbox.db3 v1 database to a new v2 relational database and the final commit in this PR depends on this one
- The third commit adds a new push notification handler

**Context of this PR:**
Delivering stored MSG's is extremely unreliable. It seems most stations are running unattended with HB timers going. You can send a HB ACK with their SNR report and MSG ID and they never get it. The message just continues to sit on your station.

QUERY MSGS is more reliable, but very very few actually use it. The message just continues to sit on your station

So I invented a new push notification system that notifies the recipient if you have a MSG stored for them. If the recipient call is seen in the call activity within the last 15 minutes the handler fires off a notification in the form `ACK MSG ID xxx` On the receiving end this fires a popup notification:

Screenshot 2026-08-10 at 20 57 24

The popup is NOT a SelfDestructMessageBox. It is a QMessageBox::Information style that persists and stays there until the recipient of the notification shoots it down. In testing of the push notification handler, even the push notifications were being ignored. The message just continues to sit on your station. Again, it's mainly because the station is running attended, not being watched, this seems to be the common theme with JS8Call. Addition of the popup notifier met with 100% instant success in getting the recipient to actually query for that MSG and get it among the three testers I have running this code.

**How does the push notification handler work?**
It does NOT run on the UI on-the-minute tick. It has its own scheduler so the code only runs once every 15 minutes from the time the software starts. It scans the call activity, queries the v2 database for matches and MSG's that are marked STORE. If it finds a match it fires the notification.

When it fires the notification it stores the MSG ID, callsign and timestamp that the notification was sent in a new database that keeps track of this. Even if you have 50 MSG's stored for other stations it will only send ONE for the 15 minute interval, and ONLY if it finds a match for a seen callsign. It will gradually work thru all 50 of them (beauty of using a database will real indexing) and everybody eventually gets notified if they don't get the stored MSG before they get a notification. It will only send ONE notification per callsign, per MSG ID every 8 hours (again, beauty of using a SQL database that keeps the timing persistent even if JS8Call is shut down and restarted).

At present it does not handle group messages by design. Those seem to be more reliable since group operators are more proactive with getting them.

Almost invariably the linux users who think you need 2,843 buttons and switches, and 1,322 context menus for anything to work, are going to claim you need a setting in the UI to turn it on or off. It's tied in with the ui->actionModeAutoreply->isChecked() so if that's enabled, the push handler is enabled, no need to add in another switch to the settings.

**WARNING:**
if you test this code it WILL migrate your old v1 database to the new v2 schema. This is not the end of the world because it retains the old v1 table for the time being so if you go back to older software it will still be there. But the migration is fire-once - done, deal. The two database tables cannot talk to each other and one or the other will be stale if you switch software versions. So recommended if you want to run this code is to back up your current inbox.db3, try the new code out. When you want to move back to the older code delete the inbox.db3 and restore it with your backup. That way, when you eventually go to the new code it will run the fire-once migration and populate your new v2 database with the current MSG's in your inbox. If you know how to work with SQL databases you can dump the old one and load the dump into the new one manually, but most people don't know how to do that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,17 @@ target_precompile_headers(
 )
 
 #------------------------------------------------------------------------------#
+# Suppress noisy narrowing-conversion warnings from vendored sqlite3.c.
+# It's third-party code we don't maintain; no need to see 78 warnings
+# on every build.
+#------------------------------------------------------------------------------#
+set_source_files_properties(
+  vendor/sqlite3/sqlite3.c
+  PROPERTIES
+    COMPILE_OPTIONS "-Wno-shorten-64-to-32"
+)
+
+#------------------------------------------------------------------------------#
 # Source files common to all platforms.
 #------------------------------------------------------------------------------#
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,6 +289,7 @@ target_sources(
   JS8_Main/fileutils.cpp
   JS8_Main/main.cpp
   JS8_Main/Modes.cpp
+  JS8_Main/MsgNotifyDB.cpp
   JS8_Main/PillRenderer.cpp
   JS8_Main/ProcessThread.cpp
   JS8_Main/qt_helpers.cpp
@@ -315,6 +316,7 @@ target_sources(
   JS8_Mainwindow/processCommandActivity.cpp
   JS8_Mainwindow/processDecodeEvent.cpp
   JS8_Mainwindow/processHeartbeatRateLimit.cpp
+  JS8_Mainwindow/pushNotificationHandler.cpp
   JS8_Mainwindow/processRxActivity.cpp
   JS8_Mainwindow/UI_Constructor.cpp
   JS8_Mainwindow/confirmThenEnqueueMessage.cpp
--- a/JS8_Main/Inbox.cpp
+++ b/JS8_Main/Inbox.cpp
@@ -4,6 +4,7 @@
  *        inbox_v2 schema. On first open, if a legacy inbox_v1 (JSON-blob)
  *        table is present, its rows are migrated into inbox_v2 once;
  *        inbox_v1 itself is left untouched and is never written to again.
+ * @see migrateV1ToV2() with associated notes on deprecation of v1
  * @author Chris-AC9KH (C) 2026 - All Rights Reserved
  */
 
@@ -197,11 +198,19 @@ bool Inbox::ensureSchemaAndMigrate() {
     return true;
 }
 
+/**
+ * @deprecated The inbox_v1 (JSON-blob) schema was deprecated as of JS8Call
+ *             3.0.3 in favor of inbox_v2's normalized columns. This method
+ *             exists solely as a one-time upgrade path for databases created
+ *             by pre-3.0.3 versions of the software: it is invoked once from
+ *             ensureSchemaAndMigrate() when inbox_v2 does not yet exist and a
+ *             legacy inbox_v1 table is found. inbox_v1 is read here but never
+ *             written to, and is otherwise left in place untouched (removal
+ *             is left for a future release). New code should not read from
+ *             or write to inbox_v1; use inbox_v2 via the query helpers below
+ *             instead.
+ */
 bool Inbox::migrateV1ToV2() {
-    // Read every row out of the legacy JSON-blob table and re-insert it into
-    // inbox_v2 as normalized columns, preserving the original row id exactly
-    // (it's referenced by value elsewhere: inbox_group_recip_v2.msg_id, and
-    // the separate msg_notify_v1.msg_id table).
     const char* selectSql = "SELECT id, blob FROM inbox_v1 ORDER BY id ASC;";
     sqlite3_stmt* sel = nullptr;
     int rc = sqlite3_prepare_v2(db_, selectSql, -1, &sel, nullptr);
--- a/JS8_Main/Inbox.cpp
+++ b/JS8_Main/Inbox.cpp
@@ -1,6 +1,10 @@
 /**
  * @file Inbox.cpp
- * @brief Implementation of the message inbox
+ * @brief Implementation of the message inbox, backed by the normalized
+ *        inbox_v2 schema. On first open, if a legacy inbox_v1 (JSON-blob)
+ *        table is present, its rows are migrated into inbox_v2 once;
+ *        inbox_v1 itself is left untouched and is never written to again.
+ * @author Chris-AC9KH (C) 2026 - All Rights Reserved
  */
 
 #include "Inbox.h"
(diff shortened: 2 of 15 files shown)


wmiler (commented):
Should have a @deprecated block regarding inbox v1 is replaced by v2 for the docs. Otherwise it all looks fine.

Chris-AC9KH:
Yeah, a @deprecated block would be a good idea. Didn't think about that.

Chris-AC9KH:
I will add here, for information, the only testers I have running this code at present is on Mac. It has never been tested on Windows or Linux (yet). I see no reason there should be any problems with it on Windows or Linux since there is nothing here that should even remotely require a preprocessor guard to compile different code for those platforms. They all handle sqlite3 databases the same, there's nothing really special about the scheduler, other that it does not run on the UI on-the-minute tick, and that was by design. Being they all handle SelfDestructMessageBox the same, I see no reason that QMessageBox::Information style would be any different.

Any testing on those platforms was delayed because I had the code for this in a different branch that's been around for awhile. With this, we no longer use things like columnCount() - 5 to extract a data item from a populated Qt widget table. It comes from the database fields instead. So it did require a fairly extensive rebase on the MessagePanel code after Rico's last PR, since that one still used the old method with the v1 JSON blob database table.

@Joe-K0OG could provide more extensive testing now that the code is forged in steel, but care must be taken to preserve your old v1 database as mentioned above to revert to the older code. If there is no changes to your basic MSG storage that matter, then it doesn't make any difference. But it WILL make a difference in the future if you go to the new code because if this code detects the v2 table it won't fire the migration code. When you start the program up you won't be able to tell that it did anything along those lines. But in the background it does a lot to make that migration to the v2 database. The earlier code didn't do that, but then I found some issues with using that old v1 database table with json_extract, which wasn't really designed for this type of use.

Chris-AC9KH:
@wmiler added the full @deprecated block at the function and referenced it from the file header doxy comments.

wmiler (approved):
Thanks

wmiler:
As an aside, we might want consider prebuilding doxygen somewhere. I'm almost done with my RL transmitter project (finally), and I did want to check if moving to the current release (1.17) would cause any problems before locking that in stone, so consider that :)

Chris-AC9KH:
@wmiler during my recent crash course in yaml and doxy it seems to me that I read where you can build and install a newer version of doxy in a sandboxed install on the runner, set it up as workflow_dispatch only, then set up an override string where it can publish the docs without touching the production /docs directory in GitHub Pages. Then don't add any deployment step to it to make sure nothing touches GitHub Pages. It would have to be set up as a completely separate workflow from the current one.

wmiler:
@Chris-AC9KH I meant just the doxygen executable, not what it generates. This last run took around 10 mins to compile it.

Chris-AC9KH:
Ah, this is a different idea. The runner did cache it (I think), but that probably expires before we run it again. I wonder if this something we can do on a local machine running Ubuntu 24 Server and use js8lib to "cache" the executable and fetch it like we do with the pre-built libs?

Chris-AC9KH:
I think it might be possible to do what I propose. I looked at the last full run and it looked to me like it fetched the executable from the cache in the form a tarball. If it can be stored there in a tarball, it can be stored on js8lib as a tarball too.

wmiler:
> Ah, this is a different idea. The runner did cache it (I think), but that probably expires before we run it again. I wonder if this something we can do on a local machine running Ubuntu 24 Server and use js8lib to "cache" the executable and fetch it like we do with the pre-built libs?

I don't see why that wouldn't work.

Chris-AC9KH:
@wmiler I just wrote a yaml, semi-coped from the workflow, to build Doxygen on the runner as a one-time build. Then grab the artifact from the runner and upload that to the v4.0 releases on js8lib. I put some ldd output in 'er that will go into the build log so we can figure out what dependencies have to be installed on the runner to simply fetch that binary, unpack it and run it without having to build it every time.

I pushed that yaml script to the workflows, have a look and if it looks good to you, run it and give it whirl. At present I have it building the same version of doxy that we use now. Get that to work, and it should be a snap to bump the doxy version, build it again, and have the workflow pull the new binary.

Chris-AC9KH:
@wmiler I figured out I can run that doxy building script on my personal repo. Still debugging it, which is what I hate about working with github runners. I got a successful doxy build, but on the last run the tarball was b()rked. I think one more push and I might have it working :-)

wmiler:
It works now. I did play with it a bit (when I belatedly realized which repo I was on 🤦‍♂️). It is possible to build doxygen with clang/llvm and get better parsing of the code, but the default runner will need a few more dependencies for that to work.

Chris-AC9KH:
I created a yaml that fetches the pre-built doxy and does the run, all except for actually uploading to the /docs on GitHub Pages. It works! It ran on my personal repo in only 46 seconds! I'd say that's a BIG improvement in runner time. I'll push it to master now and you can see how it did it, incorporate that into the regular doxy run, and test it with a real deployment.

Screenshot 2026-08-11 at 20 10 03

Chris-AC9KH:
@wmiler even though the doxy build script runs, and should probably be on js8lib instead of master here, there was a bug in that script as to how the sha256 is handled. I'm fixing that now and rebuilding the doxy binary that is hosted on js8lib.

Chris-AC9KH:
@wmiler Wyatt: got it! I rebuilt the pre-built doxy and uploaded that to js8lib. There is a new workflow on master now that uses the pre-built doxy. It's called "Test Doxygen Fetch (Dry Run). It will make the run using our own hosted doxy binary, and create a downloadable artifact of the HTML output, without deploying it to Pages.

You can run that and test it out. It works perfectly on my personal repo.

BUT, I think we should put that doxy-building workflow in js8lib in the Linux_4.0 branch, instead of on JS8Call-improved master. It has to be built on the same glibc and library set as the runners use, and it has to run on Ubuntu 24.04 or it won't work. So that seems like the logical place to put it since I already have a workflow on js8lib that builds the linux libraries for both architectures, and that's where the output is hosted anyway.

wmiler:
> @wmiler Wyatt: got it! I rebuilt the pre-built doxy and uploaded that to js8lib. There is a new workflow on master now that uses the pre-built doxy. It's called "Test Doxygen Fetch (Dry Run). It will make the run using our own hosted doxy binary, and create a downloadable artifact of the HTML output, without deploying it to Pages.
> 

Good deal, thanks.

> You can run that and test it out. It works perfectly on my personal repo.
> 
> BUT, I think we should put that doxy-building workflow in js8lib in the Linux_4.0 branch, instead of on JS8Call-improved master. It has to be built on the same glibc and library set as the runners use, and it has to run on Ubuntu 24.04 or it won't work. So that seems like the logical place to put it since I already have a workflow on js8lib that builds the linux libraries for both architectures, and that's where the output is hosted anyway.

I concur.

Chris-AC9KH:
Wyatt, the doxy building workflow is now on js8lib and it works fine. I just ran it. After it runs all you have to download the two artifacts and upload them to the current version 4.0 library releases and the workflow on master here can pull it from there.

You can set whatever version of doxy you want to build with that script. Then test the new doxy build with the new Dry Run tester, and if it all looks good, make the changes to the regular workflow.

It wasn't very hard to use our own hosted doxy build. As small as the doxy binary is, it's amazing it takes that much time to build it on the runner.

So everything is in place for it except for changing the actual workflow to use our own pre-built doxy. It think it was a BIG improvement, even though we got 11 commits on master now debugging the stupid thing :-)

Chris-AC9KH:
@wmiler Wyatt, I didn't want to touch the build-docs.yml for fear of breaking something ......

I took a look at it this morning and I THINK the only change needed in build-docs.yml is swapping out the old Cache Check doxygen / Install doxygen / Save cache doxygen steps with the Fetch prebuilt doxygen + Sanity-check doxygen steps in the dry run script. It should "just work". I took the liberty of running the dry run script this morning and it made the run in 41 seconds, so I'm pretty sure it's really fast now!

To design that I looked at the last full run and looked at how the runner fetched the cached doxy build. I pretty much duplicated that by fetching our own "cached" build instead. I thought about doing it with a shell script so it can be run independent of the runner like the AppImage one. But being it can be run in ~45 seconds and the dry run produces a nice artifact you can download to examine the HTML output, leaving it in yaml is the best way I think.

Beiri22:
With multiple messages waiting for one call, will it send ACK MSG ID xxx +cnt like in other places?

Is it only the message box or may there be an indicator icon in the call list?

Are you planning something like auto retrieve msg on the other side later on? Or is this too much automation?

Chris-AC9KH:
Tested on Windows and Ubuntu 26, this works great. Merging.....

Chris-AC9KH:
It notifies of additional stored MSG's a little differently than standard ACK or QUERY MSGS. Since it is a push notification I kept it short with just the MSG ID and no +cnt. Once the recipient gets that notification then they go thru the regular procedure to get it, plus any additional.

There is no "flag" in the call activity, the band activity row will turn red when it comes in. But the problem with unattended stations that have autoreply turned on, especially, is that all that expires from those panels before they actually check it, which is what is causing these notifications to be lost now. The popup will persist and force acknowledgment by having to click OK for it to go away. It was the single most effective way out of everything I tried to not miss the notification among the three testers I have running the code.

Automated QUERY MSGS is not going to happen because of too much danger of loosing it. It usually requires the recipient to sent a SNR? first to make sure the path is good enough to retrieve it. If the first fetch fails, it requires sending AGN? to have it repeated.. This requires operator interaction and any sort of automation will most certainly fail.

The goal of this PR was to make sure that the recipient does not miss the fact that they have a stored MSG, which the current system of relying on HB response fails to work because people either have HB and HB ACK turned off, or they aren't actually there when it comes in so it gets lost. I tried firing the notification popup on a regular HB ACK response, but that only covers a small percentage of use cases where that is actually turned on. This covers them all, and in the event the push notification is not received (doesn't decode), it will keep trying indefinitely every time the station is seen again. It is deliberately throttled to prevent spamming the waterfall. One of the testers for which I had a MSG stored for two years is running this code on his MacBook Pro. He had been notified of it several times over the last two years thru the HB system and missed every single notification. This code worked and he finally got it.

Why does this happen? He doesn't use the HB system, except to send out a single HB the first time he comes on the air. He gets responses but he doesn't actually read them. Then he leaves and his station runs all day answering everybody else's HB's, when he gets home at night he looks at it and if there's a lot of activity then the band is "open". But once again he never actually reads any of it. Then after supper he might sit down and "operate" for a couple hours and have a couple QSO's with somebody, then turn it off and go to bed - repeat the next day. The automation in JS8Call has bred this complacency. My station is also on the air 24/7, but I don't answer HB's unless I'm at the station. In both use cases the push notification handler works, while the HB notification system fails.

Chris-AC9KH:
@Beiri22 I'll mention that this PR was based on experimentation and implementing what I found that works. That doesn't mean it can't be improved upon, however. The code is semi-backwards compatible with the older software in that the notification will be received and decoded, but no popup in the older software. So if you want to test it out just grab the artifact from the latest build run on master.

Just pay attention to the notes on the database migration as it won't actually "break" anything, but it could put your inbox out of sync if you revert to the older code. The database migration code is designed for end-users that upgrade to the next version. Not for developers who may run several different versions in testing. And that is the only reason I left the old v1 table in place for now - it won't "break" anything that way, but it also can't sync between the two database versions. Along with that I also fixed the compiler spam that sqlite3 spits out when it compiles the C code with the clang compiler. That is a common problem with sqlite3 and we don't need to study 78 warnings on integer conversions in vendor code on every build :-)

Beiri22:
ACK

PR #361, opened 2026-08-31
activity.db3

### LINKED ISSUE #267: Call activity and RX history jumps to wrong band or disappears
The RX history and call activity data sometimes disappear completely after switching between bands a few times in the frequency dropdown menu and sometimes the information "jumps" to the wrong band.

I have found a repeatable way to make the information jump to the wrong band:

1. Select 40m in the dropdown menu, the radio jumps to 7078 kHz, if not already there.
2. Operate to fill some data into the RX history and call activity windows.
3. Close down JS8Call.
4. Manually tune the radio to 80m, for example 3567 kHz.
5. Restart JS8Call. It polls the radio, shows 3567 and the old RX and call activity data _from 40m_!
6. Switch to 40m in the dropdown menu, the RX and call activity data is gone - It was moved to 80m!

The above process seems to be reversible. Close JS8Call, move the radio manually back to 7078 kHz and restart JS8Call. Voila! The RX and call activity data is back om 40m.

**Expected behavior**
The activity data is expected to _not_ move between bands or to disappear.

**Desktop**
 - OS: Linux Mint 22.3
 - Version: v2.5.2 official Appimage
 - Radio: IC-7300 Mk2
 - Direct to radio: Yes
 - via FLrig or others: No

**Additional context**
The problem seems to be inherited from old code as I observed the same behavior in v2.2.0 on Win10, Raspberry Pi and Linux Mint.

By the way, v2.5.2 always seem to start up with the band activity window cleared. Not a big problem, it may even be intentional, but "band activity" is still mentioned in this checkbox:

    General/Behavior/"[] Reset the **Band Activity**, Call Activity and RX History at startup".

All the frequently changing activity data seems to reside in the file JS8Call.ini together with some more or less static configuration data. I think it would be easier to manage the activity data if it is moved to a separate file, or a number of band-specific files.

Aqueum:
Fixes #267

Call Activity and the RX text panel were stored in RAM and only saved to the .ini when JS8Call closed normally, with all bands mixed together, so:
1. a crash loses everything since the last clean close,
2. on restart the mixed history all lands on whichever band the rig is on.

What it does:
- Saves call activity and RX text to a new database, activity.db3, keyed by configuration, band and callsign.
- Saves as it goes, within a few seconds of hearing something, so a crash or power cut only loses the last few seconds. The database survives being interrupted mid-write.
- Imports your existing .ini history once, each record filed under the band it was actually heard on. The .ini is left alone so older builds still work.
- Each configuration keeps its own history: a clone starts with a copy of its parent's and the two diverge from then on, renames carry it over, and a settings reset starts clean.
- Ticking 'Reset ... at startup' now asks first, and a QSY from outside the band plan no longer carries the previous band's activity across.
- Controller logging moved to its own category — use activitystoragecontroller.js8, not mainwindow.js8.
- Also fixes stored message dial frequencies above 2.1 GHz being read back through a 32-bit conversion.

Developed with [vibe.andeye](https://github.com/andeyePro/vibe) with Fable 5 leading Opus 5 adversarial teams, reviewed on my shack Pi (Debian 13 aarch64, Qt 6.8.2) and IC-725:  
- Tested against master on the same profile: after a kill -9 mid-session everything comes back on restart whereas master loses whatever came after its last clean close. 
- Bands each keep their own history across restarts including after a power cut.
- Making the database unreadable shows a warning and JS8Call carries on, then picks the database back up once it is readable again. 

Not tested: 
- issue 2. above, as my station doesn't give CAT frequency, 
- Windows/macOS builds (but compiles).

**Known issues:**
- ~~Control -> Clear Activity... -> Clear All Activity, or checking 'Reset the Band Activity, Call Activity, and RX history at startup', etc. on a clone would now also delete the clone's parent's history, or vice versa.~~
- ~~ActivityStorageController logs under `mainwindow_js8`, a leftover from when it lived in mainwindow.cpp; changing to `activitystoragecontroller_js8`.~~

### DIFF
--- a/JS8_Main/ActivityStorageController.cpp
+++ b/JS8_Main/ActivityStorageController.cpp
@@ -1,0 +1,1241 @@
+/**
+ * @file ActivityStorageController.cpp
+ * @brief Orchestration above ActivityDB - see the class documentation.
+ */
+
+#include "ActivityStorageController.h"
+
+#include "JS8_Main/ActivitySettingsKeys.h"
+#include "JS8_Main/Bands.h"
+#include "JS8_Main/DriftingDateTime.h"
+#include "JS8_Main/Varicode.h"
+#include "JS8_UI/Configuration.h"
+
+#include <QDir>
+#include <QLoggingCategory>
+#include <QScrollBar>
+#include <QSettings>
+#include <QTextBlock>
+#include <QTextCursor>
+#include <QTextDocument>
+#include <QTextDocumentFragment>
+#include <QTextEdit>
+#include <QUuid>
+#include <QVariant>
+
+Q_DECLARE_LOGGING_CATEGORY(activitystoragecontroller_js8)
+
+namespace {
+/**
+ * @brief Order two timestamps, treating an invalid one as the older.
+ * @param lhs The candidate timestamp.
+ * @param rhs The timestamp it must beat.
+ * @return true if lhs is valid and strictly newer than rhs.
+ *
+ * Explicit rather than QDateTime's operators: since Qt 6.8 a comparison
+ * involving an invalid QDateTime is unordered, so both < and <= return
+ * false and a row carrying no timestamp - every manually added station,
+ * and every legacy-imported row - would win. Shared by the seed and by
+ * persistCallActivity()'s cache merge, so both resolve a stored row
+ * against a live one identically.
+ */
+bool newerThan(QDateTime const &lhs, QDateTime const &rhs) {
+    if (!lhs.isValid()) return false;
+    if (!rhs.isValid()) return true;
+    return rhs < lhs;
+}
+
+/**
+ * @brief Carry a displaced row's enrichment onto the row that won.
+ * @param winner The newer row, enriched in place.
+ * @param loser The row it replaces.
+ *
+ * Mirrors the UPSERT's own CASE guards: a bare re-hearing carries no
+ * grid and neither timestamp, so taking them from the row it displaces
+ * keeps what the database correctly kept.
+ */
+void carryEnrichment(CallDetail &winner, CallDetail const &loser) {
+    if (winner.grid.isEmpty()) winner.grid = loser.grid;
+    if (!winner.ackTimestamp.isValid())
+        winner.ackTimestamp = loser.ackTimestamp;
+    if (!winner.cqTimestamp.isValid())
+        winner.cqTimestamp = loser.cqTimestamp;
+}
+} // namespace
+
+ActivityStorageController::ActivityStorageController(Context context,
+                                                     QObject *parent)
+    : QObject(parent), m_ctx(std::move(context)) {
+    m_rxTextSaveTimer.setSingleShot(true);
+    m_rxTextSaveTimer.setInterval(5000);
+    m_rxTextSaveMaxTimer.setSingleShot(true);
+    m_rxTextSaveMaxTimer.setInterval(30000);
+}
+
+ActivityStorageController::~ActivityStorageController() = default;
+
+/**
+ * @brief Wire up the debounced write-on-change persistence of the RX pane.
+ *
+ * Any change re-arms the short debounce, so the stored copy trails the
+ * pane by at most a few seconds instead of being written only at
+ * shutdown. The second timer is armed once and deliberately not
+ * restarted by further changes: sustained sub-interval traffic (fast
+ * submodes, busy nets) would otherwise re-arm the debounce faster than
+ * it can fire, deferring the write - and growing the crash-loss window -
+ * indefinitely.
+ */
+void ActivityStorageController::setupRxTextAutosave() {
+    auto const flushRxText = [this]() {
+        m_rxTextSaveTimer.stop();
+        m_rxTextSaveMaxTimer.stop();
+        saveRxTextForBand(m_activityBand);
+    };
+    connect(&m_rxTextSaveTimer, &QTimer::timeout, this, flushRxText);
+    connect(&m_rxTextSaveMaxTimer, &QTimer::timeout, this, flushRxText);
+    connect(m_ctx.rxTextEdit->document(), &QTextDocument::contentsChanged,
+            this, [this]() {
+                m_rxTextSaveTimer.start();
+                if (!m_rxTextSaveMaxTimer.isActive()) {
+                    m_rxTextSaveMaxTimer.start();
+                }
+            });
+}
+
+/**
+ * @brief Start the grace period for an as-yet unconfirmed bucket.
+ *
+ * While it runs, RX text for the startup guess is held in the pane rather
+ * than filed under a band the rig has not reported; see
+ * saveRxTextForBand().
+ */
+void ActivityStorageController::beginStartupGrace() {
+    m_activityStartupTimer.start();
+}
+
+/**
+ * @brief Show the legacy ini RX text on a degraded or disabled start.
+ *
+ * Purely for display, and only on a degraded start or a disabled
+ * session: nothing here can reach the store either way. The window's
+ * startup already restored that copy from the ini and never rewrites it,
+ * so showing it beats showing an empty pane.
+ *
+ * The copy's extent is recorded in m_rxTextLegacyBand and
+ * m_rxTextLegacyBlocks, because the recovery seed and the close-time
+ * sweep treat only the text below it as this session's own.
+ */
+void ActivityStorageController::showLegacyRxTextIfDegraded() {
+    if ((activityDB()->isOpen() && !m_activityStoreDisabled) ||
+        m_ctx.config->reset_activity()) {
+        return;
+    }
+    m_ctx.settings->beginGroup("UI_Constructor");
+    auto const legacy = m_ctx.settings->value("RXActivity", "").toString();
+    m_ctx.settings->endGroup();
+    if (legacy.isEmpty()) {
+        return;
+    }
+    m_ctx.rxTextEdit->setHtml(legacy);
+    m_ctx.clearRxFrameBlockNumbers();
+    auto const *doc = m_ctx.rxTextEdit->document();
+    m_rxTextLegacyShown = true;
+    m_rxTextLegacyBand = m_activityBand;
+    m_rxTextLegacyBlocks = doc->blockCount();
+    if (doc->lastBlock().length() <= 1) {
+        // an append may splice into this trailing empty block
+        --m_rxTextLegacyBlocks;
+    }
+}
+
+/**
+ * @brief The rig reported the band the window already believes it is on.
+ * @param band The band reported.
+ *
+ * Confirms the bucket, moving the panes first if they are still showing
+ * the startup guess. The confirmed flag is set only after
+ * switchActivityBucket(), so the flush inside that switch still sees the
+ * outgoing bucket as unconfirmed - otherwise the startup guess's pane,
+ * which by then holds this band's decodes too, is written under the
+ * guessed band.
+ */
+void ActivityStorageController::bandUnchanged(QString const &band) {
+    if (m_activityBandLoaded && m_activityBand != band) {
+        switchActivityBucket(band);
+    }
+    m_activityBandConfirmed = true;
+    m_activityBandConfirmedBands.insert(band);
+}
+
+/**
+ * @brief The rig reported a different band from the one on screen.
+ * @param band The band reported.
+ *
+ * The confirmed flag is set after switchActivityBucket(), so the flush
+ * inside it still treats the outgoing bucket as the startup guess it is.
+ * The band is also recorded as confirmed, so the close-time sweep does
+ * not skip its RX text.
+ */
+void ActivityStorageController::bandChanged(QString const &band) {
+    switchActivityBucket(band);
+    m_activityBandConfirmed = true;
+    m_activityBandConfirmedBands.insert(band);
+}
+
+/**
+ * @brief Move the activity panes from one storage bucket to another.
+ * @param band The bucket to show: a band name, or "" for out-of-plan.
+ *
+ * Every caller - a band change, and the rig's first report correcting the
+ * startup guess - goes through here, so no path can leave one bucket's
+ * activity on screen under another's name, which is issue #267 itself.
+ *
+ * The outgoing bucket, when it is still the startup guess, has its cached
+ * panes dropped rather than parked: its pane may hold text and stations
+ * heard on the band the rig has just reported, and restoring that cache
+ * on a later visit would file them under the guessed band. The bucket
+ * reloads from the store at its next visit. Where the rig has not
+ * reported at all only the two per-bucket panes move; clearActivity()
+ * would also empty the compose box and the decode, command and spot
+ * queues the session has accumulated.
+ */
+void ActivityStorageController::switchActivityBucket(QString const &band) {
+    if (m_activityBandLoaded && m_activityBand == band) {
+        return;
+    }
+    if (m_activityBandLoaded) {
+        if (m_activityBandConfirmed) {
+            saveRxTextForBand(m_activityBand);
+            m_ctx.cacheActivity(m_activityBand);
+        } else {
+            m_ctx.dropBandCache(m_activityBand);
+            m_activitySeeded.remove(m_activityBand);
+            m_rxTextDirtyBands.remove(m_activityBand);
+            if (m_rxTextLegacyShown && m_activityBand == m_rxTextLegacyBand) {
+                m_rxTextLegacyShown = false;
+                m_rxTextLegacyBand.clear();
+                m_rxTextLegacyBlocks = 0;
+            }
+        }
+    }
+
+    if (m_ctx.lastBand->isEmpty()) {
+        m_ctx.rxTextEdit->clear();
+        m_ctx.clearRxFrameBlockNumbers();
+        m_ctx.clearCallActivityPane();
+    } else {
+        m_ctx.clearActivityPanes();
+    }
+    m_ctx.restoreActivity(band);
+}
+
+/**
+ * @brief Take up the bucket the window has just restored on screen.
+ * @param band The bucket now displayed.
+ * @param paneReloaded True when the panes were actually reloaded from the
+ *        RAM band caches; false when re-entering the bucket already on
+ *        screen, where the live panes are newer than any cache entry.
+ *
+ * A reload disarms the saver and resyncs its change tracking, since what
+ * was restored is what the cache holds. A band whose last flush failed is
+ * the exception: its tracking is left at the -1 sentinel so it keeps
+ * reporting unsaved, and its debounce is restarted here, because only
+ * document activity would otherwise arm it and a quiet band would never
+ * retry.
+ */
+void ActivityStorageController::bucketRestored(QString const &band,
+                                               bool paneReloaded) {
+    if (paneReloaded) {
+        m_rxTextSaveTimer.stop();
+        m_rxTextSaveMaxTimer.stop();
+        m_rxTextLastSavedRevision =
+            m_rxTextDirtyBands.contains(band)
+                ? -1
+                : m_ctx.rxTextEdit->document()->revision();
+        m_rxTextLastSavedBand = band;
+        if (m_rxTextDirtyBands.contains(band)) {
+            m_rxTextSaveTimer.start();
+        }
+    }
+
+    m_activityBand = band;
+    m_activityBandLoaded = true;
+
+    if (!m_activitySeeded.contains(band)) {
+        seedActivityForBand(band);
+    }
+}
+
+/**
+ * @brief Path of the per-band activity store, beside the message inbox.
+ * @return The absolute native path of activity.db3.
+ */
+QString ActivityStorageController::activityPath() const {
+    return QDir::toNativeSeparators(
+        m_ctx.config->writeable_data_dir().absoluteFilePath("activity.db3"));
+}
+
+/**
+ * @brief The activity store, opened on first use.
+ *
+ * A store that fails to open - or that closed itself after repeated
+ * failures - is retried on a throttle rather than per call. Callers may
+ * always use the returned object: every method on it is a no-op while it
+ * is closed. The throttle is armed however the handle ended up closed, so
+ * a store that breaks mid-session is retried too, and discovering that
+ * self-close here is its one operator-visible signal, the per-write
+ * warnings being suppressed once the handle is closed.
+ *
+ * The status message is emitted once per failure episode rather than once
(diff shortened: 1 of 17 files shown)


Chris-AC9KH:
@Aqueum it will take me awhile to evaluate this. We've added a lot of code to mainwindow here, which usually is not desirable as that file becomes unmaintainable. So I need to evaluate why this code wasn't broken out into separate member functions or methods in JS8_Mainwindow.

Aqueum:
Don't spend any time evaluating now in that case @Chris-AC9KH.  It shouldn't take long to pull the activity-DB code out of mainwindow into a dedicated class before your review.

Chris-AC9KH:
@Aqueum that would be perfect. When ever we touch mainwindow.cpp we try to reduce the amount of code in that file.

Aqueum:
OK, that's hopefully fixed now @Chris-AC9KH.  mainwindow.cpp is much lighter, the orchestration is now ActivityStorageController in JS8_Main/, and the controller doesn't include mainwindow.h - the window state it needs comes through a small Context struct.

Built clean on the same Pi (but still not tested on macOS or Windows) with no new compiler warnings, and the persistence checks re-run - history restored per band, band round-trips, kill -9 recovery, configuration switch, and degraded start with mid-session recovery all behave as before.

I must confess that I didn't look in docs for the contributor guide, hence the three prior commits.  The second fixed an issue found in testing, the third was an improvement proposed by an independent Gemini 3.1 Pro review.  I trust you'd rather have this fourth commit separate, so we can squash after your review.  

I also couldn't add a bug label - is that due to my permissions?

I wonder if a 'minimise code in the following files: ' section in the contributor guide would help?  Also I understand that if we were to rename the guide docs/CONTRIBUTING.md or (easier to find prior to starting the PR) add a CONTRIBUTING.md in the root that links to it, GitHub would automatically surface it in the new-PR flow.

Chris-AC9KH:
@Aqueum great, I'll take a look at this tonight. Yes, we probably should mention something for developers where mainwindow.cpp is concerned. That file was over 14,000 lines of code last winter before I started breaking it down and pulled a lot of the big member functions out of (what was) the MainWindow class. I refactored that to the UI_Constructor class so the files contained in JS8_Mainwindow are member functions of that class, called from mainwindow.cpp.

So any time we touch that file (which basically powers the entire UI), we should strive to make it smaller, not bigger.

Chris-AC9KH:
@Aqueum Martin, I checked out the PR with gh and built it with Qt 6.12 on Mac. So I'll actually run the code for 2-3 days before I dig into the code that actually does it, and see how it works.

Chris-AC9KH:
The first thing I did was look at the database structure after it built it. I like it! We got two tables, one for the call activity, and another for textEditRX. And I see you're using Write-Ahead-Logging, which is pretty much crash-proof with SQLite3. Nice!

Screenshot 2026-08-31 at 13 31 02

@rruchte @wmiler you guys should take a look at this. I could see creating another table for the settings and get rid of the JS8Call.ini altogether. That .ini is extremely fragile compared to this. I'll put it through its paces for a few days and see if I can break it.

Chris-AC9KH:
@punk-kaos Kelly, I added you to the review list too. I find this one quite interesting and see some extended possibilities here.

punk-kaos:
> @punk-kaos Kelly, I added you to the review list too. I find this one quite interesting and see some extended possibilities here.

Ohh hey, that is pretty neat. Could probably think of some interesting ways to take advantage of the stored data there.

punk-kaos (commented):
cloned configurations seem to share the same activity database identity, would that mean that cloned configs are not unique and share the parent's DB entries?

wmiler:
> I also couldn't add a bug label - is that due to my permissions?
> 
Most likely.

> I wonder if a 'minimise code in the following files: ' section in the contributor guide would help? Also I understand that if we were to rename the guide docs/CONTRIBUTING.md or (easier to find prior to starting the PR) add a CONTRIBUTING.md in the root that links to it, GitHub would automatically surface it in the new-PR flow.

GitHub recommends putting CONTRIBUTING.md in the root of the repo, I just haven't gotten around to doing it. If you want to do a separate PR for that, I'll accept it. Be advised you will need to adjust the doxygen config for the new location as well.

Aqueum:
> cloned configurations seem to share the same activity database identity, would that mean that cloned configs are not unique and share the parent's DB entries?

Yes exactly, and my apologies, I actually came across this in testing this afternoon.  The assumption was that the database is what the antenna heard rather than a factor of the config.  

But Control -> Clear Activity... -> Clear All Activity or checking 'Reset the Band Activity, Call Activity, and RX history at startup', etc. on a clone would now also delete the clone's parent's history, or vice versa.  I can change clone to mint a fresh id and copy the rows across, so clones start as a snapshot and then diverge - the same semantics as Master's ini.

Unfortunately I'm rather tight on time until the weekend.  It's not a particularly complex fix, so I could push an untested commit tonight, but I'd rather test it myself on Sunday if people can wait..?

JS8_Main/ActivityStorageController.cpp
```diff
@@ -0,0 +1,1224 @@
...
+#include <QSettings>
+#include <QTextBlock>
+#include <QTextCursor>
+#include <QTextDocument>
+#include <QTextDocumentFragment>
+#include <QTextEdit>
+#include <QUuid>
+#include <QVariant>
+
+Q_DECLARE_LOGGING_CATEGORY(mainwindow_js8)
```
wmiler: Do you really want this, or would a different logging category be more appropriate?

Aqueum:
> GitHub recommends putting CONTRIBUTING.md in the root of the repo, I just haven't gotten around to doing it. If you want to do a separate PR for that, I'll accept it. Be advised you will need to adjust the doxygen config for the new location as well.

Yes, happy to do that on the weekend too.  Might we want to continue the conversation in AI created code in JS8Call #222 so we can put some extra stuff in on what is wanted from PRs like this, to minimise the risk of maintainers being asked to review slop?

JS8_Main/ActivityStorageController.cpp
Aqueum: Sorry no, that's a hangover from when this lived in mainwindow.cpp.  I'll change to activitystoragecontroller_js8 / "activitystoragecontroller.js8".

wmiler (commented):
I really like this, and I think it is solidly moving in the right direction. 

Many thanks for all the doxy comments!

The ini file should *only* contain the bare minimum to get the program up and running. I haven't looked lately to see if there is actually any bare minimum requirements in the ini file, but if there isn't anything, then we can flag it for deprecation in a follow-up PR.

Aqueum:
> The ini file should _only_ contain the bare minimum to get the program up and running. I haven't looked lately to see if there is actually any bare minimum requirements in the ini file, but if there isn't anything, then we can flag it for deprecation in a follow-up PR.

Would you agree that writeable_data_dir() is QStandardPaths::AppLocalDataLocation (Configuration.cpp:1334), and never read from the ini, so nothing in the ini is needed to find the store? If the only genuine bootstrap item is which MultiSettings configuration is current, that could equally be a row in the DB. So the bare minimum is plausibly zero.

If everyone agrees, I can put in a separate PR for that after this is merged and settled.

Chris-AC9KH:
I would prefer to not scope-creep this PR. Keep its scope on the call activity and Rx panel, fix the cloned settings bug, then we'll trash it and see if we can break it. I'm interested in seeing how the database recovers from pulling the plug from the wall, for instance.

Save moving settings to the database for a later PR. I'm not sure what's in that .ini that is absolutely required for bootstrap before database connections open. Probably nothing. But yeah, let's look at that after this has settled out.

One comment I will make for Martin. The doxy is excellent. IMO the inline comments in the code are a bit excessive. If the code is well-written it already says what it does without any comments. At most, usually a one-liner is sufficient if it needs to be commented for something that may not be clear to maintainers or developers.

Chris-AC9KH:
So, initial run thoughts here:

We've now got snr, grid, dial, submode, and cq_ts/ack_ts/utc_ts per (config, band, callsign) data, so "last heard, on what band, what mode, what SNR trend" is just a SELECT away instead of something you'd have had to reconstruct from log files or not at all. The .ini's RAM-only, unbanded model genuinely couldn't have supported this. There was no durable per-callsign history to query, just whatever happened to be in the display cache at last clean shutdown.

So I'm seeing the possibility of creating a report with band-activity summary, for a given callsign, or mode used. Something that was never possible before. The database keeps a record of every callsign ever heard, even after it expires from the call activity table.

As I run the code and see what it does, then go back to the code and look at that, all the inline comments makes it hard to read the code. Too much noise. Those definitely have to go. So I'm going to request that all that be removed, except for what might be absolutely necessary to clarify a variable in a function or something that might not be immediately apparent to a maintainer or developer.

Aqueum:
Thanks @Chris-AC9KH, @wmiler and @punk-kaos, my todo list is then:

1. Fix the cloned-settings bug: swap `ActivityDbId` for an `ActivityDbCloneFrom` marker in the `MultiSettings` clone paths, and on the clone's first use mint a fresh UUID, copy the parent's rows across to it, then clear the marker, so the two diverge from then on instead of sharing.
2. Change `ActivityStorageController`'s logging category from `mainwindow_js8` to `activitystoragecontroller_js8`.
3. Prune the inline comments, leaving only what's assumed to be genuinely unobvious to a maintainer or future developer.
4. Test as before, plus Clear All Activity and pull-the-power.

Do shout if you spot anything else.  I won't be able to resume work on it until the weekend, and hope I can test on Sunday.

Aqueum:
Quick question on what to include:

The clone and logging category fixes are coded and the comments pruned. The code review process turned up a few bugs in the new code, fixed in separate commits which I'll squash (RX text dropped at exit after a QSY, a debounce stuck at 60 s, Clear All asking twice, RX text doubling on returning to a band, a re-import after per-band clears).

Two things go beyond what we agreed, I could drop them, save for a separate PR, or keep as their own commits here:

1. "Reset activity at startup" now wipes every band's stored history, so I added a confirmation prompt when you tick that option. (17 lines, in 2 additional files: Configuration.cpp and Configuration.ui)
2. A QSY from outside the band plan carried the old Band Activity across; master does this too. (4 lines)

Branch is now +2,988 lines, or ~2,965 without the above; I'll squash to whatever you want.  I still need to conduct the GUI-only checks and pull-the-plug.

Chris-AC9KH:
Let's leave the two things in, Item 1 especially seems sensible.

Chris-AC9KH:
@Aqueum please remove the update the User Guide in this PR. I opened a new PR that makes quite extensive changes to the User Guide, and included your change to that PR https://github.com/JS8Call-improved/JS8Call-improved/pull/362/changes/a81be642cff1d48fdddd2b95835bf88365fd5df3

This will help prevent a conflict on merge.

Aqueum:
> @Aqueum please remove the update the User Guide in this PR. I opened a new PR that makes quite extensive changes to the User Guide, and included your change to that PR [a81be64](https://github.com/JS8Call-improved/JS8Call-improved/commit/a81be642cff1d48fdddd2b95835bf88365fd5df3)

Sorry missed this in the rush to push after completing my tests (forgot today was last day before eldest heads back to Uni).  Done now.  The squashes probably killed your inline comments.  I hope all is well otherwise.

One for #362: a81be64/4f8611a has the paragraph as it stood a couple of commits ago, so it's missing the end of it:
"Turning on the reset-at-startup option asks first too. Cloning a configuration gives the clone its own copy of the stored activity, after which the two keep separate histories."

Aqueum:
@Chris-AC9KH, @wmiler & @punk-kaos

Re-ran the live checks on this build - per-band restore, band round-trip, kill -9, config clone, and Clear All - plus the two new ones: while my RPi wasn't delighted with its plug being pulled, JS8Call was completely unfazed.  And deleting clones leaves their parents untouched.

Hoping this is mergeable now, but happy to do more on this (or hopefully the AGENTS.md) next weekend.  Have a great week.

Chris-AC9KH:
@punk-kaos @wmiler @Aqueum I cleaned up inline comments in the code. Excessive inline commenting merely pollutes the code and makes it hard to read. The doxy blocks are actually useful and are what is supposed to tell what's going on.

I suspect this was coded with some sort of AI. While I've never used any of these AI algorithms, from what I've seen they seem to parse the code and then plaster what they figured out as inline comments that are WAAAY too verbose. So I went thru the whole thing, trying to read the code, and when inline comments interfered with that, I got rid of it or shortened it up like a real human would do it.

Otherwise I'm happy with it. Built it and running it. I'll put it through its paces for a couple days and see if anything obvious turns up. If not, I'll approve this PR for merge.

wmiler:
> @punk-kaos @wmiler @Aqueum I cleaned up inline comments in the code. Excessive inline commenting merely pollutes the code and makes it hard to read. The doxy blocks are actually useful and are what is supposed to tell what's going on.
> 
> I suspect this was coded with some sort of AI. While I've never used any of these AI algorithms, from what I've seen they seem to parse the code and then plaster what they figured out as inline comments that are WAAAY too verbose. So I went thru the whole thing, trying to read the code, and when inline comments interfered with that, I got rid of it or shortened it up like a real human would do it.

Even just getting AI to make correct doxy blocks without breaking doxygen is a royal pain (using copilot via VS Code). Have had to reclone the repo a couple of times today already. I really was hoping it had improved, but oh well, as they say, can't always take the shortcuts on documenting!

Aqueum:
@Chris-AC9KH 
> I cleaned up inline comments in the code. 

Really sorry about that, I could/should have taken them out myself.  The new files were pruned after your initial request. Actually the initial prune pretty much just moved all the inline comments into doxy, but the commit you received was a second version that was much more ruthless.  Unfortunately neither round touched the header, mainwindow.cpp, or the JS8_Mainwindow call sites' inline comments.  If we can specify exactly what level of comments we want (like zero inline) that would probably be an easier rule for AI to enforce.  Otherwise it struggles to know what "may not be clear to maintainers or developers." and errs on the side of way too much.

> I suspect this was coded with some sort of AI. 

Yes, and I thought I was open about that from the start: "Developed with [vibe.andeye](https://github.com/andeyePro/vibe) with Fable 5 leading Opus 5 adversarial teams,"

> Otherwise I'm happy with it. Built it and running it. I'll put it through its paces for a couple days and see if anything obvious turns up. If not, I'll approve this PR for merge.

Great to hear that, and sorry again for the selective prunning - I'll try to work on an AGENTS.md next week which should help.

@wmiler 
> Even just getting AI to make correct doxy blocks without breaking doxygen is a royal pain

Chris previously said "The doxy is excellent" on this PR - so I'm hoping an AGENTS.md can make that repeatable.

Chris-AC9KH:
> Really sorry about that, I could/should have taken them out myself. The new files were pruned after your initial request. Actually the initial prune pretty much just moved all the inline comments into doxy, but the commit you received was a second version that was much more ruthless. Unfortunately neither round touched the header, mainwindow.cpp, or the JS8_Mainwindow call sites' inline comments. If we can specify exactly what level of comments we want (like zero inline) that would probably be an easier rule for AI to enforce. Otherwise it struggles to know what "may not be clear to maintainers or developers." and errs on the side of way too much.

@Aqueum not really a big deal because it forced me to read the whole PR from end to end. As I read a file I get a "picture" in my head of how it executes and what external functions it's calling. When you run into those big inline comment blocks it disrupts getting that "picture" of the execution order. Call it excessive "noise", poor SNR :-)

Inline comments aren't totally useless. They can sometimes help with figuring out the execution order of the code, or what a particular variable or function() does. I don't know that AI is "smart enough" to know how a human programmer thinks. So it requires a human programmer to review what it does.

If it can help save time, like with Wyatt's project where we want better doxy docs, then it's great. But if it keeps breaking that, then it doesn't really save much time. One of these days I'll have to try it and see how smart it thinks it is. Supposedly you can integrate it into Xcode somehow (I'm a Mac programmer) but I haven't figured that out yet because most of the time when I'm working in Xcode I have a specific idea in mind that I want to accomplish and don't have the patience to deal with something that don't think the way I do :-) In Xcode I can pull up the diff from a recent commit from the right side panel and use regex to search the code for things I want to deal with like inline comments. That saves a lot of time and I don't see how AI could do it any better.

Same with running the code sanitizer in Xcode. It instruments the code at runtime, so it will recompile the project with the necessary compiler flags (like -fsanitize=address). Then I can look at the diagnostic output to see if we have memory issues like buffer overflows, out-of-bounds array access, or use-after-free bugs. If the sanitizer detects an error during execution, the application will pause. Xcode will highlight the exact line of C++ code causing the violation and output a detailed diagnostic report in the Issue Navigator. Can AI do this better? I doubt it.

I do use VS Code on Mac for working with HTML, as Xcode is not very good at that. This CoPilot thing kept popping up in that and it won't leave me alone. I managed to shoot that down and haven't see it since, but fortunately I don't have to use that program very often.

Aqueum:
Thanks @Chris-AC9KH

I'm also a Mac user, but I have an RPi running my JS8Call near 24x7 - I access it via RealVNC on my iPhone and Mac.  I would have tested with the Mac too but my shack is in my eldest's bedroom, and they were back on holiday for the last couple of weeks.  I'm using a 1990's IC-725 hooked up to an [AIOC](https://github.com/skuep/AIOC) and it would have taken too much time in their bedroom to wire up the mac to replace the RPi. 

These days I use [vibe.andeye](https://github.com/andeyePro/vibe) in Ghostty for most things, and now just use Xcode for compiling and VS Code for diffs.  I was always very slow at coding, so I employed a full-time software developer when I needed to accelerate my online conferencing startup during COVID lockdown.   I now use AI as if it were a team of developers (as does the full-time dev I previously employed), with one expensive model ([Fable 5.1](https://www.anthropic.com/claude-fable-and-mythos-5-1)) leading teams of [Opus 5](https://www.anthropic.com/news/claude-opus-5) who do the coding, and cheaper models doing easy admin tasks.

Other than the cost saving ($2,400/year gets me that team, not full-time, but enough for my needs) what I really love is that it never complains.  Even if I've just had it pull an all-nighter then I pull a complete U-turn in the morning and have it undo everything and implement its suggestion that I previously dismissed.

It's definitely not perfect and if you enjoy coding and have the time to do it, I wouldn't suggest changing.  But it is much better than I am at coding, and it's only going to improve with time.

On the sanitiser - no, that's one place I'd say the tooling wins outright, and I haven't run ASan over this. If you do and it turns anything up in the sqlite3 binding paths I'd be more than happy to fix it.

wmiler:
> @wmiler
> 
> > Even just getting AI to make correct doxy blocks without breaking doxygen is a royal pain
> 
> Chris previously said "The doxy is excellent" on this PR - so I'm hoping an AGENTS.md can make that repeatable.

That might make things better 🤷🏻‍♂️ I've found I had to be very verbose with my prompts to even get it close to what I wanted. Currently only running it against the header files, but it still causes a boatload of warnings to be thrown out, and eats up all its tokens fixing the mistakes it made itself. 🤦‍♂️ I'll have to try it on my local models next to see if they work any better, since they are free :)

Chris-AC9KH:
@Aqueum I don't personally use linux or Windows for anything except to see if a JS8Call build will run on either one, and that a new change to the code functions properly. And the instances of those operating systems are run here in a virtual machine on the Mac as arm64.

I use the command line flags -fsanitize=address -fsanitize=thread with the clang compiler. Since this is debug code it increases memory usage and slows down execution of the code. But neither ASAN nor TSAN have turned up anything with the code in this PR.

So anyway, I don't see anything wrong with this code. It handles the database well. On multiple stack samples it shows virtually zero overhead to manage the databases, and read/writes are so fast it normally only takes <1ms. libsqlite3.dylib and com.apple.CoreData are both loaded but every sampled stack bottoms out in idle waits with virtually zero memory/cpu/disk footprint because unlike flat text files, databases don't have to be loaded in RAM. Flat text files are loaded sequentially from top to bottom into memory. SQLite organizes data into fixed-size pages, 4KB each. These pages are indexed using a highly optimized data structure called a B-Tree. When JS8Call looks up a callsign, frequency, or log entry, SQLite doesn't look at the whole database. It traverses the B-Tree index to find the exact page it needs, pulls only that 4KB chunk into RAM, and reads it. This is why it takes virtually 0-1 ms. The disk and memory I/O footprint scales flat, whether the database has 10 entries or 100,000 entries. Quite impressive, actually, to see this implemented.

It keeps a permanent record of every callsign ever heard by your station, which ALL.txt can obviously do but not with the efficiency of a SQL database. I've also been watching the handling of the database at the UI with a database client.

Screenshot 2026-09-07 at 09 38 39

Chris-AC9KH (approved):
I ran this for 9 hours on arm64 Windows 11 without any issues. I did not test on linux since Martin already did that. Approved for merge pending @wmiler or @punk-kaos review.

Chris-AC9KH:
@wmiler pending your review, we can merge this. I see nothing wrong with it and have been running it continuous for several days now. I really like the new database as I see further uses for it in the future.

wmiler:
Code looks fine, I'll get loaded up on my machine in the morning and give it a quick test.

Chris-AC9KH:
I've already stress-tested it and can't get the database to fail. SQLite3's WAL is pretty robust, although I did blow the .ini on one of those stress-tests. Good thing I had the forethought to back that up before I did dumb things like pull the plug out of the wall on the Mac mini during actual database writes :-)

Hint: good way to test it - go on a band and let it collect some activity in both the call activity and textEditRX (incoming MSG pane). Switch bands and let it collect some activity there. Do another band if you want. You will find that it repopulates textEditRX and the call activity (depending on what you have your expire time set to) spot-on every time, without fail for that band. It will retain the incoming text in textEditRX indefinitely, per band, until you select to clear it.

Pretty nice - the old .ini can no way do that and you can "confuse" the .ini and get it to carry activity from one band to another, then come back and it's all gone. This is so many times more robust it's not even a contest.

wmiler (approved):
Looks good, ran it most of the morning, bands are a bit quite, but no issues with this.
Thank you @Aqueum Martin!

Aqueum:
Thanks @Chris-AC9KH and @wmiler for all the testing.  Great to hear it's doing the trick.

I tried to pull the plug on my RPi mid-write.  Took more effort to reboot the Pi than I was expecting, but, when I relaunched JS8Call, everything seemed fine - no database or .ini corruption that I could see.

If there's anything you haven't mentioned in this thread that you'd have rather was done differently just let me know, and I'll try to generalise it for the AGENTS.md.  Be as brutal as you like, I won't take offence.

Likewise if there are any tests, etc. that you'd like to be done in any given circumstance, or any exemplar files in terms of in-line comments and doxy, style, etc. do let me know.  I guess here may not be the best place for that. If [Discussion #222](https://github.com/orgs/JS8Call-improved/discussions/222) isn't ideal either I could raise an issue.

PR #363, opened 2026-09-05
workflows: new build-and-deploy-docs script that uses prebuilt doxy

Chris-AC9KH:
@wmiler this is a new script that replaces the old one and uses pre-built doxy from js8lib.

On js8lib master there is a build-doxy.yml workflow. You can update that workflow to build whatever version of doxy you want to use. After it builds it, download the artifact and the sha artifact, upload them to the js8lib release. Then update this script (and the dry run one) to pull the new doxy build. If satisfied, delete the old doxy build from the js8lib release.

This allows building the docs in under a minute compared to the original script which built and cached doxy. The build-doxy workflow on js8lib automates building a new version of doxy that you may want to try. It was part of my effort to make this whole system more efficient and use less runner time. You'll notice there is also a workflow on js8lib for building the linux libraries we use for the AppImages, which was part of my automation effort to make this easier to update the linux libs.

The linux Qt builds cannot be done on the runner - not enough disk space. Those still have to be done on private machines.

If happy with this change, just merge it.


wmiler:
Yeah, I ran the dry run version before a few times. Seems to be working ok, I'll do a once over when I get back to the house tonight.

wmiler (approved):

Chris-AC9KH:
@wmiler did you run the build and deploy script? It should be just as fast as the dry run one. If it proves to be, it was one of those deals where I was thinking about maybe running it on a weekly cron job like the run cleanup script does. I guess I should've labeled the last merge as Documentation to trigger it, but I failed to do that so now the User Guide is out-of-date in the Code Resources on the website :-(

But I did run the dry run one, downloaded the HTML from that and made a new PDF for download on the website :-)

wmiler:
I just ran it here, so should be all up to date. I think I'd rather just set it to run on a PR merge or doc label.

Chris-AC9KH:
Ah yes, I see now that it ran in 52 seconds, which I think is pretty good? It seems to me that the old version took like 6 minutes or something just to build doxy.

PR #368, opened 2026-09-14
CONTRIBUTING.md move

Aqueum:
Moves `docs/CONTRIBUTOR_GUIDE.md` to `CONTRIBUTING.md` in the repository root, where GitHub looks
for it and links to it from the issue and pull request UI, as @wmiler suggested in #361

The Doxyfile `INPUT` line is updated to the new path in the same commit, otherwise the docs build
warns about a missing file on every run. A three-line pointer is left at the old path for anyone —
or any coding agent — used to finding the guide there; it carries no content of its own.

The guide's own text is unchanged, so the diff is a rename plus the two small additions above.

Developed with [vibe.andeye](https://github.com/andeyePro/vibe) using [Opus 5](https://www.anthropic.com/news/claude-opus-5) reviewed by [GPT-6 Astra](https://openai.com/index/gpt-6-astra/).

PR #367, opened 2026-09-13
Header doc updates Part 1

wmiler:
Updates to the documentation in the header files only. Please check for corrections.

Issue #293, opened 2026-05-13
The Great API rework Part 2


wmiler:
This is a continuation of #87 and contains items that were not completed in that issue.

- #19 Needs to be verified that this overlap issue has been fixed
- Items from [jfrancis42 PR](https://github.com/js8call/js8call/pull/92)
- Discussion on TX.START/END
- STATION.DESELECT/STATION.CLEAR_TEXT
- Add scheduling endpoints?

PR #369, opened 2026-09-14
Contributor & AI guidance

Aqueum:
**What changed and why:** 
1. Additions to CONTRIBUTING.md to avoid bloat to three large files, plus a documentation section mentioning the new DOXYGEN.md
2. addition of AGENTS.md and 3. DOXYGEN.md in response to the AI-contribution discussion in #222 and following review of #361. 

Built on master at `245f0582`, which includes #368.

**What it does:**

1. `CONTRIBUTING.md` gains a section on where new code goes: the largest hand-written files are closed to substantial new code, and there are two established places to put it instead: `JS8_Mainwindow/` for logic that needs the window's own state, a class under `JS8_Main/` for code that owns state of its own. That convention is already visible in the directory layout but wasn't documented, which is how #361 ended up adding 1,200 lines to `mainwindow.cpp` before @Chris-AC9KH pointed it out. It gains a second section on documenting what you add and keeping functions small enough to review.
2. `AGENTS.md` is new, at the repository root. It is the file a contributor points their coding agent at: how to build and test, what to ask their user to do, what the PR description has to state, and what gets an agent-written PR rejected here.
3. `docs/DOXYGEN.md` is new and covers what this project's Doxyfile expects of a Doxygen block, and how to check one locally — @wmiler noted on #361 how much work it takes to get correct blocks out of an AI without breaking the docs build, Claude suggests that most of that comes down to a handful of settings in the Doxyfile that differ from the defaults.

Some of what is in `AGENTS.md` may be useful in `CONTRIBUTING.md` too, but I decided not to include those there at this time so as to not make human-only contributions more onerous. Happy to move anything across that we would like humans to see as well as AI.

**Developed with:** 
[vibe.andeye](https://github.com/andeyePro/vibe), using [Fable 5.1](https://www.anthropic.com/claude-fable-and-mythos-5-1), [Opus 5](https://www.anthropic.com/news/claude-opus-5) and [Sonnet 5](https://www.anthropic.com/news/claude-sonnet-5), reviewed by [GPT-6 Astra](https://openai.com/index/gpt-6-astra/) with its corrections applied by [GPT-5.6 Sol](https://openai.com/index/previewing-gpt-5-6-sol/) and then reviewed again by Opus 5.

**Checks run:** 
No code changes, so the compiler, sanitiser and Debug-build checks do not apply.
Xvfb smoke start, to exercise the one check in the file that could be run: JS8Call starts clean on a virtual display with `QT_FATAL_WARNINGS=1` and emits no Qt warnings on a fresh profile (Debian 13 aarch64, Qt 6.8.2). On the documents themselves: every file and directory path the three name resolves in the tree, 22 of them; every Doxyfile setting they quote a value for matches `.github/workflows/misc/Doxyfile`, 13 of them, no mismatches; every shell snippet parses under
`bash -n`; and the diff is clean under `git diff --check`.

doxygen 1.16.1, the version CI uses, built from source and run over the whole tree on master and on this branch: 1,994 warnings on each, and the two logs are identical once each worktree's own path prefix is stripped. No new warnings. Running it did find a bug in `docs/DOXYGEN.md`'s own comparison recipe, now fixed: doxygen reports absolute paths, so diffing two worktrees' logs without normalising the prefix first returns every line as changed.

**What was tested:** 
Nothing by hand — there is no behaviour to exercise. Both new documents were read against the repository they describe, and the claims above were checked mechanically rather than by eye.

**Not tested:** 
The rendered output. The doxygen runs above were parse-only, with every generator turned off, so they prove the tree gains no new warnings but say nothing about how the two new pages look once built. The binary was doxygen 1.16.1 built from source on aarch64 rather than CI's prebuilt x86-64 one; same version, different build. @wmiler, the semantics in `docs/DOXYGEN.md` are read off the Doxyfile and doxygen's own documentation and would still benefit from your eye - I mainly needed something for AGENTS.md to point to.


Aqueum:
@wmiler, could you add a documentation label, and WIP label if appropriate?  I suspect discussion is required, and would be surprised if docs/DOXYGEN.md doesn't need at least a few edits from yourself.

docs/DOXYGEN.md
```diff
@@ -0,0 +1,175 @@
...
+
+-   **`EXTRACT_PRIVATE = NO`, `EXTRACT_PRIV_VIRTUAL = NO`, `EXTRACT_STATIC = NO`.**
+    Private members, private virtuals and file-static functions are not extracted, so they do not
+    warn and a block on one never reaches the output. Document them only where a maintainer
+    genuinely needs the explanation. `EXTRACT_LOCAL_METHODS = NO` is set as well, but it governs
+    Objective-C implementation methods and so does not apply to this tree.
+
+`WARN_AS_ERROR = NO`, so the docs build will not fail on any of this. Nobody is going to catch it
+for you; check the log yourself.
+
```
wmiler, line 39: Also, don't use `@fn` this causes problems with v1.16 for some reason that I haven't bothered to figure out yet.

docs/DOXYGEN.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+Because `INHERIT_DOCS = YES`, an undocumented override can inherit the base declaration's
+documentation. Do not restate it; document an override only where its behaviour genuinely differs.
+
+Because `DISTRIBUTE_GROUP_DOC = NO`, documentation attached to one member is not distributed to
+the other members in its member group. Give each member its own block. Separately, assign module
+membership with `@ingroup` or by placing the declaration inside an `@{` and `@}` group range;
+membership does not require physically moving the declaration.
+
```
wmiler, line 89: New groups should be defined in docs/defines.dox

docs/DOXYGEN.md
```diff
@@ -0,0 +1,175 @@
...
+An addition to the API gets `@note API x.y+` in its block, where the version is the next release,
+and `docs/API.md` is updated in the same PR.
+
+Markdown works inside blocks, and `AUTOLINK_SUPPORT` turns names that resolve into links. A name
+that does not resolve is silently left as plain text rather than warning, so wrap anything you do
+not want linked - and anything you are unsure of - in backticks.
+
```
wmiler, line 96: Caution: This doesn't exactly work right inside markdown tables (doxygen throws warnings, see a build log docs/API.md `@ALLCALL` for an example). I don't have a work around for this yet.

AGENTS.md
wmiler, line 1: I'm going to have see this in action before I can make improvement comments.

wmiler:
Related #367

docs/DOXYGEN.md
```diff
@@ -0,0 +1,175 @@
...
+    Objective-C implementation methods and so does not apply to this tree.
+
+`WARN_AS_ERROR = NO`, so the docs build will not fail on any of this. Nobody is going to catch it
+for you; check the log yourself.
+
+## What a block looks like here
+
+At the top of every new file, named to match the file exactly - a `@file` carrying the wrong
+name is a warning:
+
```
wmiler, line 44: Ensure case of file names matches.

docs/DOXYGEN.md
```diff
@@ -0,0 +1,175 @@
...
+membership does not require physically moving the declaration.
+
+An addition to the API gets `@note API x.y+` in its block, where the version is the next release,
+and `docs/API.md` is updated in the same PR.
+
+Markdown works inside blocks, and `AUTOLINK_SUPPORT` turns names that resolve into links. A name
+that does not resolve is silently left as plain text rather than warning, so wrap anything you do
+not want linked - and anything you are unsure of - in backticks.
+
+## Checking before you submit
```
wmiler, line 97: I don't know if this needs to be stated or not, but namespaces should be documented as well, excluding the default 'empty' namespace.

wmiler:
Also see [Chris's comment here](https://github.com/JS8Call-improved/JS8Call-improved/pull/367#issuecomment-5680290248)

docs/DOXYGEN.md
Aqueum: done

docs/DOXYGEN.md
Aqueum: done?

docs/DOXYGEN.md
Aqueum: done

Aqueum: Does this give the reason: https://www.doxygen.nl/manual/commands.html#cmdfn
https://www.doxygen.nl/manual/docblocks.html#structuralcommands

docs/DOXYGEN.md
Aqueum: Tested it on 1.16.1 with our Doxyfile - the table isn't special, and backticks do work
in a cell. Only a bare `@name` warns:

| in a table cell | warning | renders as |
|-----------------|---------|------------|
| `@ALLCALL`      | Found unknown command | @ALLCALL |
| backticked      | none | @ALLCALL |
| `\@ALLCALL`     | none | @ALLCALL |
| `@@ALLCALL`     | none | @ALLCALL |
| `&#64;ALLCALL`  | none | &#64;ALLCALL, unrendered |

Same results outside a table. Guide updated in d4eef7a4.

The warnings you're seeing are the ~60 bare names in the group-callsign table at
docs/user_guide/JS8Call_User_Guide.md L240-249 - I couldn't find @ALLCALL in docs/API.md.
Happy to send a PR escaping those if you want them gone, but it'll be a couple of weeks, as I'm away this weekend.

wmiler: Sorry, I had the wrong file. A similar thing happens with the API command list.

```
/home/runner/work/JS8Call-improved/JS8Call-improved/docs/API.md:38: warning: unable to resolve reference to 'rigget_freq' for \ref command
```

docs/DOXYGEN.md
wmiler: Not really, but I'll accept your fix for now until I can test a newer version of doxygen which has a related fix.

Chris-AC9KH:
What is the status of this one? I don't know enough about the nuts and bolts of doxy, I've learned enough about doxy to create what I think are sensible and concise doxy comments in my code. A recent example is here when I rebuilt styles.h for a new UI layout. I try to keep the comments useful for code documentation, and at the same time instructive for people modifying a platform-specific style:
https://github.com/Chris-AC9KH/JS8Call-improved/blob/pr/UI_overhaul/JS8_UI/styles.h

This is my adopted method, maybe not 100% "correct", but building documentation to tell other developers how to do it is a little beyond my pay grade unless I see something I consider to be way too excessive and verbose. I only documented the first section in detail, then switch to short inline comments to mark subsequent sections. I felt this tells people how the file works, the inline comments are there to help developers quickly identify what they need to find to change it based on what they learn from studying the first fully-documented section.

wmiler (approved):
Since this is mostly just documentation stuffs. I'm ok with approving this. It's sat here long enuf for anybody that wants to comment to have done so.

Aqueum:
Thanks @wmiler - I mentioned being away last weekend, and work has been manic crazy since, I was hoping to get back to address everyone's comments this weekend.  I hope I implemented all of your changes.  

The big one will be "I'm going to have see this in action before I can make improvement comments." - now you've merged it, I can try a completely vibe coded PR this weekend and you'll see whether the doxy, etc. is as desired.  If not we can tweak AGENTS, CONTRIBUTING or DOXYGEN.md as required.  

Let me know if there's a suitably not too huge Issue you'd like me to test it on.  I see The Great API rework Part 2 #293 is the only one marked 'help wanted' - I haven't looked into this so don't know how large the scope is, or if my available credits could cover it in a single weekend.  I also am no API expert - that may be good or bad, as we'll be relying on Claude to just fix it with almost no input from myself.

Alternatively, I could have a go at a small corner of the epic Re-designed User Interface #366 as my RPi version refuses to do a nice dark mode - so there I at least know what I want, and can try to generalise it so it works for any Linux user..?

